Issue #141, opened 2025-04-21
Verify Speedy.f90 default scenario


eldavenport:
Run the default case that comes with speedy.f90 (includes realistic orography and ssts) and verify the output.

duncanwp:
Closed?

PR #297, opened 2025-11-14
adding in temperature over land

eldavenport:
## Describe your changes
Already existed in our forcing file, wasn't being read in or updated by forcing.py. Required updating default_statistics.nc. Moved the lfluxland flag into ForcingData, because we had bugs arising from inconsistency in True with no ForcingData or always False. Now the flag is set to True if you pass in ForcingData and otherwise it is false. 

If we want to have lfluxland=True without supplying forcing data, then we need sensible default values for all of those fields.

#141 

- [X] I have performed a self-review of my code
- [X] My code follows the style guidelines of this project
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/jcm/forcing.py
+++ b/jcm/forcing.py
@@ -14,7 +14,7 @@ class ForcingData:
     soilw_am: jnp.ndarray # soil moisture (used to be soilwcl_ob in fortran - but one day of that was soilw_am)
     stl_am: jnp.ndarray # temperature over land
     sea_surface_temperature: jnp.ndarray # SST, should come from sea_model.py or some default value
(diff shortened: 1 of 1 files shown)


eldavenport:
ok i'm not sure how/why but the default scenario (without forcing/geometry or anything) is now broken. and that is the setup used for some of the slow tests. as soon as you add forcing/geometry it's fine. i dont know how that is affected by this... but i'm working on it

PR #298, opened 2025-11-20
updating diffusion according to comparisons to speedy.f90

eldavenport:
## Describe your changes
Separating the diffusion of different pieces of the prognostic state. To simplify things for now I have also removed the filtering of the tendencies (the function is still there, but it is not applied). That can be added back in if we want to, but for now it is not making a substantial difference. I also updated the model stats test to look at global averages for now, hopefully that is less finicky than the zonal averages. 

#141 


jcm/diffusion.py
```diff
@@ -4,19 +4,41 @@
 
 @tree_math.struct
 class DiffusionFilter:
-    tendency_diff_timescale: float # Diffusion timescale (s)
-    tendency_diff_order: int # Order of diffusion operator for tendencies
-    state_diff_timescale: float # Diffusion timescale (s)
-    state_diff_order: int  # Order of diffusion operator for state variables
+    vor_q_timescale: jnp.float32 # Diffusion timescale (s)
```
duncanwp: Is there a reason we have to specify the exact type? I think there are `jnp.float_ ` types we can use to let the JAX configuration decide? Just in case it causes problems with different devices

jcm/model.py
```diff
@@ -186,10 +186,36 @@ def conserve_global_mean_surface_pressure(u, u_next):
                 log_surface_pressure=u_next.log_surface_pressure.at[0, 0, 0].set(u.log_surface_pressure[0, 0, 0])
             )
         
+        # create diffusion filter function handles
+        diffuse_div = DiffusionFilter.make_diffusion_fn(
```
duncanwp: I'm not sure about this pattern. I think partly because it's a class method and I don't think it should be, but also that it's having to unpack specific parts of the diffusionFilter object for each call. I guess part of the problem is that we can't have real methods on pytrees? Perhaps it can be a (private) method on Model?

jcm/model_test.py
```diff
@@ -283,10 +283,10 @@ def test_speedy_model_default_statistics(self):
 
         model, predictions = run_default_speedy_model(save_interval=30.)
         pred_ds = model.predictions_to_xarray(predictions)
-        pred_ds_monthly = pred_ds.isel(time=-1).mean(dim='lon') # zonal monthly means, take the last month
+        pred_ds_monthly = pred_ds.isel(time=-1).mean(dim={'lon', 'lat'}) # global monthly mean, take the last month
```
duncanwp: Technically we should weight these averages but since we're only comparing across runs it doesn't really matter. Maybe add a todo?

duncanwp (commented):
Not quite sure about how the different filters get made. I left some comments but happy to jump on a zoom call if that would be easier

jcm/diffusion.py
```diff
@@ -4,19 +4,41 @@
...
+        from dinosaur.filtering import horizontal_diffusion_filter
+
+        def diffusion_filter(u, u_next):
+            eigenvalues = grid.laplacian_eigenvalues
+            scale = dt / (timescale * abs(eigenvalues[-1]) ** order)
+
+            filter_fn = horizontal_diffusion_filter(grid, scale, order)
+
+            u_temp = filter_fn(u_next)
+            return replace_fn(u_next, u_temp)
```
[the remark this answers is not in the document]
eldavenport: I don't think it matters that it doesn't return a new DiffusionFilter. but it doesn't operate on any class-level data which does make it less of a classmethod and more of a utility

jcm/model.py
```diff
@@ -186,10 +186,30 @@ def conserve_global_mean_surface_pressure(u, u_next):
                 log_surface_pressure=u_next.log_surface_pressure.at[0, 0, 0].set(u.log_surface_pressure[0, 0, 0])
             )
         
+        # create diffusion filter function handles
+        diffuse_div = self._make_diffusion_fn(
+            self.diffusion.div_timescale,
+            self.diffusion.div_order,
```
duncanwp: Sorry, do we need diffusion to be on `self` if we're only using it here in the constructor?

duncanwp:
No problem, yes please do that

duncanwp (approved):
Looks good, thanks!

PR #305, opened 2025-11-21
Add a bit of helper logic to streamline jvp/vjp workflows

jvmadan:
The changes here are necessary to make the `jvp` code example in the paper work.


duncanwp:
LGTM, thanks!

PR #313, opened 2025-11-25
Condense setup into pyproject.toml

jvmadan:
Turns out the setup.cfg hadn't been being read properly, so all the `D` lint checks are only now enabled. There are a couple hundred failing so I added them to the ignore list for now.


duncanwp:
Can you check the dependencies? They should really pull from the requirements.txt (if they need to be specified)

duncanwp (approved):
I'm not a big fan of having the docstrings start on the first line, but this looks good, thanks for doing it

PR #316, opened 2025-11-25
Factor out surface land fluxes

duncanwp:
## Describe your changes

Factor out the land surface flux code into a separate function and add tests specifically for it. They all pass (which is annoying because we're trying to track down an issue with precipitation over land)


duncanwp:
We can keep this or not, up to you

PR #324, opened 2025-12-10
Fix swrad icltop indexing

### LINKED ISSUE #311: Lack of convective precipitation over land
The issue is most pronounced in the tropics and may have something to do with a lack of excess humidity (qdif) over land even when convection is occurring

jvmadan:
The missing `precnv` over land (see #311) was due to the temperature over land (specifically, the temperature in the bottom two layers) being ~4-5C colder than speedy.f90 on average. Passing the same prognostic states to the physics modules of speedy.f90 and jcm and taking the difference of the computed temperature tendencies produced this plot (averaged over a year of prognostic states):

image

showing that there was substantially less heat flux from the land surfaces to the atmosphere in jcm than in speedy. (Note: the anomalies in the longwave radiation ttend were just artefacts of skipping the step of clamping specific humidity to be positive before running the physics, which turns out to be important over greenland and east antarctica.)

The missing surface heat flux turned out to be due to a lot of missing downwards shortwave radiation flux:

image

Which was due to a bug in how we were slicing into `tau2` using `icltop` (see the diff for this PR). With that issue fixed, we no longer have missing downwards shortwave radiation:

image

The annual mean temperatures in the lower atmosphere are now within 1.5C of speedy and convective precipitation is behaving realistically over land. Closes #311

### DIFF
--- a/jcm/physics/speedy/shortwave_radiation_test.py
+++ b/jcm/physics/speedy/shortwave_radiation_test.py
@@ -180,6 +180,8 @@ class TestShortWaveRadiation(unittest.TestCase):
         _, physics_data = get_clouds(state, physics_data, parameters, forcing, geometry)
         physics_data = get_zonal_average_fields(state, physics_data, forcing, geometry)
         _, physics_data = get_shortwave_rad_fluxes(state, physics_data, parameters, forcing, geometry)
+
+        # FIXME: testing against itself at the moment, need to get updated values from speedy.f90
         
         # surface downward shortwave radiation at all latitudes
         self.assertTrue(np.allclose(physics_data.shortwave_rad.rsds[0, :], [


jvmadan:
I'm opting not to add it to jax gotchas because it's not specific to jax, this is just numpy behavior - if you index into an array with another array, it's like doing an outer product with the indexing array, it doesn't just broadcast alongside the slices you've passed to the other dimensions.

jcm/physics/speedy/shortwave_radiation.py
```diff
@@ -61,8 +61,9 @@ def shortwave_rad_fluxes(operand):
     mask = icltop < kx
     clamped_icltop = jnp.clip(icltop, 0, tau2.shape[0] - 1).astype(int) # Clamp icltop - 1 to be within the valid index range for tau2
     # Start with tau2 and update the values where the mask is true
-    tau2 = tau2.at[clamped_icltop, :, :, 2].set(
-        jnp.where(mask, parameters.shortwave_radiation.albcl * cloudc, tau2[clamped_icltop, :, :, 2])
+    i_idx, j_idx = jnp.meshgrid(jnp.arange(ix), jnp.arange(il), indexing='ij')
```
duncanwp: I'm OK with not putting something in the gotchas, but can we at least add a comment here that explains the need for mesh grid?

jcm/physics/speedy/shortwave_radiation_test.py
```diff
@@ -159,7 +159,7 @@ def test_shortwave_radiation(self):
         psa = jnp.ones(xy)
         precnv = -1.0 * np.ones(xy)
         precls = 4.0 * np.ones(xy)
-        iptop = 8 * np.ones(xy, dtype=int)
+        iptop = np.ones(xy, dtype=int) * jnp.linspace(0,kx,il).astype(int)[jnp.newaxis,:]
```
duncanwp: This is one column with multiple iptops, or multiple columns with changing iptops?

duncanwp (approved):

PR #330, opened 2025-12-13
Fix a couple operations that throw slow-compilation warnings

jvmadan:
* Fixes `E external/xla/xla/service/slow_operation_alarm.cc:65] Constant folding an instruction is taking > 1s` warning that would show up (at least for me) when running on higher resolutions.
* 2 of the above warnings popped up for runs a few months long at T85, so I would estimate the amount of compilation time this fix saves as ~2s for a few months at T85 with it scaling relative to that for other durations/resolutions.
* Does not change runtime of the model run itself


duncanwp (approved):
Approve

PR #359, opened 2026-01-14
Fix various bugs

jvmadan:
Still a draft because one of the shortwave radiation tests needs updating

# convection.iptop

convection.iptop was looking totally different from the fortran values:
* the mean iptop in jcm was ~2-3 rather than the ~8-9 in speedy
* jcm's iptop seemed to be off by one in the regions where speedy's iptop was ~3-4

The former issue was due to large scale condensation setting iptop to 2 rather than leaving conv.iptop unchanged in cases where there was no large scale condensation. The latter issue was due to an off by one error (needed to be +2 rather than +1) in the areas where large scale condensation did update iptop.

Implementing the above two changes revealed an additional bug: in much of the mid/high latitudes, speedy.f90 typically exhibited large scale condensation that would update iptop, while jcm did not. This turned out to be because we were applying rhblsc to rhref at all levels instead of just the bottom one.

reference 1 month average iptop from speedy.f90:
image

jcm iptop before fixes:
image

jcm iptop after fixes:
image

I also had to fix one of the large scale condensation tests that mistakenly assumed iptop would be 0-indexed in jcm, I'm pretty sure we wanted to match the fortran values (the iptop as computed by diagnose_convection was already 1-indexed) but I'll have to double check one more time that the rest of the code is consistent with this.

# shortwave_rad.icltop

There were inconsistencies in whether it was being treated as 0-indexed or 1-indexed, I cleaned them up. There was then also a bias over deserts being caused by a subtlety in vectorizing the loop that sets icltop and cloudc using the layer of maximum relative humidity, which I was able to fix.

# tau2 (channel 0)

There was a significant discrepancy between the jcm and speedy values, I still don't know what the issue was--I just re-ported the problematic sections.

### DIFF
--- a/jcm/physics/speedy/shortwave_radiation_test.py
+++ b/jcm/physics/speedy/shortwave_radiation_test.py
@@ -142,7 +142,7 @@ class TestShortWaveRadiation(unittest.TestCase):
         forcing = ForcingData.zeros((ix, il))
         geometry = convert_to_speedy_latitudes(Geometry.from_grid_shape(nodal_shape=(ix, il), num_levels=kx))
 
-    @pytest.mark.skip(reason="needs ground truth from speedy.f90") # FIXME
+    # FIXME: currently testing against itself, needs updated values from speedy.f90
     def test_shortwave_radiation(self):
         from jcm.geometry import Geometry
         from jcm.physics.speedy.test_utils import convert_to_speedy_latitudes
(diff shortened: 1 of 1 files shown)


duncanwp:
Nice, thanks! Can you take a look at why the integration test is failing? Do we need to update the climatology?

PR #367, opened 2026-01-28
Change unit of precipitation from kg/m²/s to g/m²/s

AndrewILWilliams:
Fix unit of large-scale precipitation, pretty sure it's just a typo in the output field.


docs/source/speedy_translation.rst
```diff
@@ -1,43 +1,6 @@
 Speedy Variable Translation
 ===========================
 
-Physicsstate
```
duncanwp: How come you deleted this section?

PR #369, opened 2026-01-28
Fix sim_time overflow issue

### LINKED ISSUE #368: Weird behavior in multi-decadal run
In assessing the energy conservation of JCM, I ran a multi-decadal simulation (3000 days, outputting 30 day averages, see below). Everything looks good until year ≈70, when suddenly the seasonality disappears and there's a sudden jump in the net TOA radiation:

Similar behavior is seen in the global-mean precipitation and evaporation:

The source of the bug seems to be that the model stops experiencing seasons?? This is the `shortwave.fsol` variable:

Code to generate data:
```python
from jcm.model import Model
from jcm.geometry import Geometry
from jcm.forcing import ForcingData
from importlib import resources

trunc=31

# Load the terrain and forcing data (interpolated to the default model grid)
data_dir = resources.files('jcm.data.bc.t30.clim')
realistic_geometry = Geometry.from_file(data_dir / 'terrain.nc', target_resolution=trunc)
realistic_forcing = ForcingData.from_file(data_dir / 'forcing.nc', target_resolution=trunc)

model = Model(
    geometry=realistic_geometry,
)

predictions = model.run(
    save_interval=30,
    total_time=30000,
    output_averages=True,
    forcing=realistic_forcing,
)

pred_ds = predictions.to_xarray()
pred_ds.to_netcdf(f"/scratch/cimes/aw2143/jcm/T{trunc}_longrun.nc")
```

Could someone else run this and check if you get the same behavior?

jvmadan:
2^31 / (365 * 24 * 60 * 60) = ~68 years before an int32 number of seconds overflows and the functions that compute tyear and model year stop working. Easy fix: use the days argument when constructing the jdt.Timedelta. jdt.Timedelta already reorganizes its contents into separate int32s for days and (seconds % 86400), so this shouldn't affect the code at all.

Closes #368

### DIFF
--- a/jcm/date_test.py
+++ b/jcm/date_test.py
@@ -1,6 +1,8 @@
 import unittest
 from jcm.date import fraction_of_year_elapsed, DateData
+from jcm.model import Model
 import jax_datetime as jdt
+import jax.numpy as jnp
 
 class TestDateUnit(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


duncanwp:
Nice! Can you add a couple of tests that reproduce the issue (#368) and now pass

Issue #230, opened 2025-09-17
Refactor Geometry to be physics-agnostic


jvmadan:
our Geometry struct should be suited for any possible Physics module, at the moment it's lifted directly from Speedy. It might not take much refactoring to clean this up, since the fields in Geometry are mostly things that are generally useful for any physics module anyway, but at the very least we should change it so the sigma_layer_boundaries dictionary is provided by the physics module.

jvmadan:
* Geometry is serving two different purposes right now--it's 1) our user interface for the model grid, and it's also 2) the interface we pass to the physics tendency methods
* The main reason not to just use dinosaur coordinatesystem for purpose 1 is that we want this user interface to also handle the terrain file read-in - my intuition is that we should keep that out of the model constructor, to match the ForcingData pattern and I think just a vague sense that our model constructor used to take more arguments and I remember that feeling worse
* I think there are changes to be made on the PhysicsData side / arguments to physics tendencies side of the equation though.
* Since the Physics module is known at model construction, we could have physics modules provide their own GeometryData struct (since we had to take the physics-module-specific fields out of Geometry anyway) which would take the model coords (or whatever) and compute the relevant fields, saving the GeometryData as a Model or Physics field. Am I forgetting a reason that we aren't already doing this to avoid re-computing fields that are the same every timestep (at least the ones that don't depend on forcings, tyear, etc? We could even name it StaticData, throw static parameters like `lfluxland` in there, mark it as a static argument on the physics tendency functions, and then get to use regular if statements rather than jax.lax.cond in surface fluxes, etc

eldavenport:
Having terrain data separate also simplifies some of the invisible dependencies between lfluxland, forcingdata, and geometry (geometry has fmask, but lfluxland is in forcing data but we only want to set lfluxland if we have terrain which is read in by geometry). In this case lfluxland could just go in terrain data along with fmask and other things.

duncanwp:
Adding @shoyer for visibility in case you have any strong preferences for dealing with land masks, orography and things like layer thickness.

eldavenport:
> OK, I think this makes sense:
> 
> **Proposed Refactoring: Geometry → TerrainData + Physics-internal Coord Caching**

I like this proposed refactor, that’s pretty much what I was envisioning. I’m sure it might change a little, but I am in favor of going forward with this. As for the name, I’m ok with TerrainData or OrographyData ? No strong preferences.

PR #372, opened 2026-01-29
Feat/geometry refactor

eldavenport:
## Describe your changes
These changes separate what was Geometry into TerrainData and coords. At the user interface, coords is the CoordinateSystem type used by dinosaur. In addition, there is a SpeedyCoords object that calculates speedy-specific parameters related to the speedy coordinate system. A speedy coords instance is created from coords, and is added to physics_data at the beginning of compute tendencies. The user should never really need to use SpeedyCoords directly.

As documented in #230 we decided upon: 

Proposed Refactoring: Geometry → TerrainData + Physics-internal Coord Caching

Split Geometry into two concerns:

1. TerrainData (new struct) - boundary conditions that vary per simulation:
- contains orog, phis0, fmask, lfluxland (moved from ForcingData - logically belongs with terrain)
Factory methods: TerrainData.from_file(), TerrainData.aquaplanet()

2. Physics-internal coord caching - each physics module manages its own needs:
- SPEEDY stores a SpeedyCoords cache at init (hsg, fsg, dhs, sigl, grdsig, grdscp, wvi, sia, coa)
- not seen by the user

SIGMA_LAYER_BOUNDARIES stays in speedy/physical_constants.py (no longer imported by utils)
New Model initialization:

```
 Model(
      coords: CoordinateSystem,        # single source of truth for grid
      terrain: TerrainData = None,     # boundary conditions (defaults to aquaplanet)
      physics: Physics = None,         # builds its own coord cache from coords
      ...
  )
```

Updated Physics interface:
```
def compute_tendencies(self, state, forcing, terrain: TerrainData, date) -> ...
```

Benefits:
No circular Geometry ↔ Coords dependency
Physics-agnostic core (SIGMA_LAYER_BOUNDARIES not in utils.py)
Clear separation: terrain (boundary conditions) vs grid structure vs physics-specific caching
Each physics module handles vertical coordinates its own way
Static coord cache works because Model (including model.physics) is already static in the JIT

### DIFF
--- a/jcm/physics/speedy/humidity_test.py
+++ b/jcm/physics/speedy/humidity_test.py
@@ -31,7 +31,6 @@ class TestHumidityUnit(unittest.TestCase):
         self.qg_standard = jnp.ones((kx,ix,il))*2
 
     def test_spec_hum_to_rel_hum_isnan_ones(self):
-        from jcm.constants import grav
         xy = (ix, il)
         zxy = (kx, ix, il)
         
@@ -42,8 +41,6 @@ class TestHumidityUnit(unittest.TestCase):
         qa = 5. * jnp.ones(((kx, ix, il))) #temperature
         rh = 0.8 * jnp.ones(((kx, ix, il))) #relative humidity
         phi = 5000. * jnp.ones(((kx, ix, il))) #geopotential
-        phi0 = 500. * jnp.ones((ix, il)) #surface geopotential
-        fmask = 0.5 * jnp.ones((ix, il)) #land fraction mask
         sea_surface_temperature = 290. * jnp.ones((ix, il)) #ssts
         rsds = 400. * jnp.ones((ix, il)) #surface downward shortwave
         rlds = 400. * jnp.ones((ix, il)) #surface downward longwave


eldavenport:
I think we are close...only 2 tests are failing. One I know how to fix, the other I'm working on. There are a few things to discuss:

1) We used to have the interpolation function internally create a coords object, and then use the coords.horizontal for interpolating. The way I implemented this, you now give interpolate() the coords object. I think this is more transparent. This means that if you want TerrainData and ForcingData to do interpolation to a target_resolution, then you pass your model coords object to them (see 01_jcm_demo.ipynb). If we are ok with this paradigm, then I can go ahead and update the interpolate_tests (one is failing for missing arguments but requires a bit of work to fix so I just haven't done it yet).

2) Something is broken in the predictions_to_xarray() function that is not obvious to me. I am working on that. 

3) SpeedyPhysics.compute_tendencies() needs to know about `coords` so that it can use them to generate the cached physics data coordinate stuff at the beginning of each time step. This means that the init for SpeedyPhysics takes coords as an argument. It feels redundant that the Model() also takes coords as an argument... see below.

```
Model(
     coords=coords,
     terrain=terrain,
     physics=SpeedyPhysics(coords)
)
```

In the default case where you don't give the models a physics argument, then it internally passes the coords argument to SpeedyPhysics(coords), which seems nice. It just seems not nice that you use coords twice when you want to specify the physics... Is this ok as it is? I don't love it, but I am not 100% sure how to change it. I don't want to give SpeedyPhysics a default for the coords because that makes it possible for SpeedyPhysics to get a default CoordinateSystem that does not match the model CoordinateSystem.

docs/source/design.rst
```diff
@@ -155,7 +157,7 @@ A core design goal is full differentiability through the model. This enables:
 
    def loss(params):
        physics = SpeedyPhysics(parameters=params)
-       model = Model(physics=physics)
+       model = Model(coords=coords,physics=physics)
```
duncanwp: This is a bit annoying having to add it everywhere. To make it slightly easier for the user can we make a predefined default set of coords that can just be imported and used directly. eg.: `Model(T32L8)` or `Model(default_speedy_coords)` or something?

docs/source/developer.rst
```diff
@@ -147,11 +147,12 @@ To profile the model and identify performance bottlenecks:
 .. code-block:: python
 
    import jax.profiler
+   from jcm.physics.speedy.speedy_coords import get_speedy_coords
 
    # Start a trace and create a Perfetto trace file
    jax.profiler.start_trace("./tensorboard_logs", create_perfetto_trace=True)
 
-   model = Model(time_step=30.0)
+   model = Model(time_step=30.0,coords=get_speedy_coords())
```
duncanwp: Oh right, sorry you did this here 😅 Perfect. Let's make it the first argument though to keep it clear

docs/source/getting_started.rst
```diff
@@ -106,14 +111,17 @@ You can customize various aspects of the model:
...
 .. code-block:: python
 
-   from jcm.geometry import Geometry
+   from jcm.terrain import TerrainData
+   from jcm.physics.speedy.speedy_coords import get_speedy_coords
 
    # Higher resolution: T85 (256x128 grid)
-   geometry = Geometry.from_spectral_truncation(spectral_truncation=85)
+   coords = get_speedy_coords(spectral_truncation=85)
+   terrain = TerrainData.aquaplanet(coords=coords)
```
duncanwp: This is much nicer

jcm/data/bc/interpolate.py
```diff
@@ -125,13 +125,13 @@ def upsample_terrain_ds(ds: xr.Dataset, target_resolution: int) -> xr.Dataset:
...
-    ds_interp = _upsample_ds(ds, target_resolution)
+    ds_interp = _upsample_ds(ds, grid)
     ds_interp['lsm'] = ds_interp['lsm'].clip(0.0, 1.0)
     # not clamping orog to avoid erasing real areas below sea level, but this might allow bad extrapolated values at the extreme latitudes
     return ds_interp
 
-def interpolate(target_resolution, output_dir=None):
-    f"""Interpolate T30 forcing and terrain data to target resolution and save to output directory.
+def interpolate(grid, output_dir=None):
+    f"""Interpolate T30 forcing and terrain data to target resolution (inferred from grid) and save to output directory.
```
duncanwp: 'defined by grid' no?

jcm/data/bc/interpolate.py
```diff
@@ -144,6 +144,7 @@ def interpolate(target_resolution, output_dir=None):
 
     Args:
         target_resolution: Target spectral truncation number. Must be one of: {VALID_TRUNCATIONS}.
+        grid: Dinosaur HorizontalGridTypes instance for the target resolution.
```
duncanwp: At some point we might have different grid types (I found a cool project developing another JAX dycore), but this is fine for now

jcm/model.py
```diff
@@ -194,39 +198,37 @@ def outer_step(carry):
...
-        
+
         Args:
             time_step:
                 Model time step in minutes
-            geometry: 
-                Geometry object describing the model grid and orography of the model
+            terrain:
+                TerrainData object describing boundary conditions (orography, land-sea mask, etc.)
             coords:
```
duncanwp: Can you move this to the top

jcm/physics/speedy/speedy_physics.py
```diff
@@ -87,49 +90,62 @@ def __init__(self,
 
         self.terms = physics_terms if not checkpoint_terms else [jax.checkpoint(term, static_argnums=static_argnums.get(term, ()) + (4,)) for term in physics_terms]
     
+    def cache_coords(self, coords: CoordinateSystem):
+        """Store model coordinate system for SpeedyCoords calculation in compute_tendencies"""
+        self.coords = coords
```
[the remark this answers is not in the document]
eldavenport: Yes… that would make sense

PR #377, opened 2026-02-06
Enable command-line arguments for non-interactive Docker runs

### LINKED ISSUE #376: Make Docker image accept command arguments for non-interactive simulations

I think something like this:

```FROM python:3.10-slim

WORKDIR /app

COPY requirements.txt .

COPY . .

```

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: jcm-deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app: jcm
  template:
    metadata:
      labels:
        app: jcm
    spec:
      containers:
      - name: jcm-container
        # env:
        # - name: CMD_ARGS
        resources:
          limits:
            memory: "128Mi"
            cpu: "500m"
        ports:
```

VanshikaSalekar:
## Describe your changes
This PR updates the Dockerfile to support passing command-line arguments to JCM simulations.

The Docker image now uses the ENTRYPOINT + CMD pattern, allowing arguments to be overridden at runtime.  
Documentation has also been updated with Docker and Kubernetes usage examples to enable non-interactive execution.

Fixes #376  
https://github.com/VanshikaSalekar/jax-gcm/issues/376

- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,84 @@
+FROM python:3.11-slim
+
+WORKDIR /app
+
+# Install jcm and dependencies
+RUN pip install --upgrade pip && \
+    pip install jcm hydra-core omegaconf
+
+# Create config directory and config file
+RUN mkdir -p /app/config && \
+    cat > /app/config/config.yaml << 'EOF'
+model:
+  time_step: 10
+  save_interval: 10
+  total_time: 10
+  start_date: "2026-02-01"
+
+hydra:
+  run:
+    dir: outputs/${now:%Y-%m-%d}/${now:%H-%M-%S}
+EOF
+
+# Create custom main script
+RUN cat > /app/run_model.py << 'EOF'
+import hydra
+from omegaconf import DictConfig, OmegaConf
+from jcm.model import Model
+from hydra.core.hydra_config import HydraConfig
+from pathlib import Path
+import jax_datetime as jdt
+
+@hydra.main(version_base=None, config_path="config", config_name="config")
+def main(cfg: DictConfig):
+    """Run JCM Model with adjustable parameters"""
+    
+    # Allow command-line overrides
+    OmegaConf.set_struct(cfg, False)
+    
+    # Convert string date to jax_datetime.Datetime
+    start_date = jdt.to_datetime(cfg.model.start_date)
+    
+    print(f"🚀 Starting JCM simulation with:")
+    print(f"   Time step: {cfg.model.time_step} minutes")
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -140,3 +140,21 @@ This project is licensed under the Apache License 2.0 - see the [LICENSE](LICENS
...
 
 For questions or collaboration inquiries, please open an issue or contact dwatsonparris@ucsd.edu.
+
+## Running Non-Interactive Simulations in Docker
+
+The Docker image supports passing command-line arguments for non-interactive runs.
+
+### Docker Example
+```bash
+docker run --rm jcm-test start_date=2026-02-06 db.user=k8s_user
```
duncanwp: I don't think the db.user example is relevant so remove that. It was copy/pasted into the example sorry

Dockerfile
```diff
@@ -16,13 +16,18 @@ RUN apt-get update && apt-get install -y  \
...
 WORKDIR /app
 
+# Install Python dependencies
 COPY requirements.txt .
 RUN pip install --upgrade pip setuptools wheel
 RUN pip install --no-cache-dir -r requirements.txt
 
+# Copy project files
 COPY . /app
 RUN pip install -e .
```
duncanwp: Actually, I think we can simplify all these pip installs now since we can just install jcm directly from pypi with: `pip install jcm`

duncanwp (commented):
This is great, thanks! A couple of minor fixes attached. Would you mind actually checking it works locally? I can push it to a Docker registry if needed - just let me know

VanshikaSalekar:
Thanks for the feedback!

I've simplified the Dockerfile to install jcm directly from PyPI as
suggested and removed the irrelevant db.user example. I also added a custom
entry point to properly handle Hydra configuration overrides via
command-line arguments.

Tested locally and confirmed working. Let me know if you'd like me to push
the image to a registry or make any other adjustments.

VanshikaSalekar:
Thanks for the feedback! I've updated the PR as suggested:

- Removed the irrelevant `db.user` example from README.md
- Simplified the Dockerfile by installing `jcm` directly from PyPI instead
of multiple pip steps
- Updated the README to correctly show non-interactive runs using Hydra
syntax (`+parameter=value`)

I also tested locally:
```bash
docker run --rm jcm-test +time_step=20 +layers=4 +total_time=30
```

and it runs successfully with the expected outputs.

Let me know if there are any other tweaks needed before merging.

PR #378, opened 2026-02-07
Docker cmd args

### LINKED ISSUE #376: Make Docker image accept command arguments for non-interactive simulations

VanshikaSalekar:
## Describe your changes
- Updated the Dockerfile to use `ENTRYPOINT ["python", "-m", "jcm.main"]` so the container can accept Hydra configuration overrides via command-line arguments.
- Updated README.md with a “Running Non-Interactive Simulations in Docker” section showing the correct `+parameter=value` syntax for Hydra overrides.
- Removed irrelevant examples (e.g., `db.user`) that were copied from prior examples.
- Simplified installation instructions in Dockerfile by installing `jcm` from PyPI instead of building from source.

- Resolves #376 — Make Docker image accept command-line arguments for non-interactive simulations

### DIFF
--- a/jcm/__init__.py
+++ b/jcm/__init__.py
@@ -1,1 +1,3 @@
+import logging
+logging.basicConfig(format='%(name)s: %(asctime)s %(levelname)s: %(message)s')
 __version__ = "1.0.0"
--- a/jcm/model.py
+++ b/jcm/model.py
@@ -24,6 +24,10 @@ from jcm.utils import DYNAMICS_UNITS_TABLE_CSV_PATH, get_coords
 from jcm.diffusion import DiffusionFilter
 import pandas as pd
 from functools import partial
+import logging
+
+# logging.basicConfig(format='%(name)s: %(asctime)s %(levelname)s: %(message)s')
+logger = logging.getLogger(__name__)
 
 _LEGACY_SCAN_API = version.parse(flax_version) < version.parse("0.10.0")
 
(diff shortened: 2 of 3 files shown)


jcm/main.py
```diff
@@ -8,9 +8,7 @@
 def main(cfg: DictConfig):
     """Run Speedy Model with adjustable parameters"""
     model = Model(
-        time_step=cfg.model.time_step,
-        layers=cfg.model.layers
-    )
+        time_step=cfg.model.time_step)
```
duncanwp: Why did you remove the layers argument?

duncanwp (commented):
Looks good, but it looks like the change to main is still there, can you clean that up then I'll merge

duncanwp:
@VanshikaSalekar Do you think you would be able to pick this up? It was very close, and we've now updated hydra to make the CLI more complete

VanshikaSalekar:
Hey @duncanwp,

Thanks for the ping! I've gone ahead and picked this back up.

The branch is updated with the following changes:

   -

   Restored the layers argument in jcm/main.py so the original model logic
   remains entirely untouched.
   -

   Updated main.py to utilize hydra_cfg.runtime.output_dir, replacing the
   old manual string-parsing block to take full advantage of the recent Hydra
   update.
   -

   Added clear examples to the README demonstrating how to override
   parameters with Docker and how to use volume mounting to access the
   model_state.nc output files.

Let me know if everything looks good to merge now!

Best, Vanshika

PR #389, opened 2026-03-02
Fix issue with coords setup

jvmadan:
Since the nodal_shape argument overrides the spectral_truncation argument, nodal_shape should default to None (the existing value was redundant anyway). This way, a user-specified spectral_truncation OR nodal_shape will actually be used, rather than only the latter.


duncanwp (approved):
Yeah, this is definitely an improvement

PR #405, opened 2026-03-13
Make p1 nonzero

jvmadan:
p1 needs to be nonzero for the random_seed argument to do anything. I made a quick plot (n=10 seeds) of what the variance between different seeds looks like over time, for different values of p1--TLDR, the value of p1 hardly matters. It can be very small without slowing down the physically realistic divergence between different seeds.
image
For completeness here's the variance over a longer duration (to get a sense of how large the initial noise perturbation is compared to the eventual internal variability).
image
Note that output_averages is off for these plots so the save_interval doesn't affect the variance between different seeds.

### DIFF
--- a/jcm/model.py
+++ b/jcm/model.py
@@ -244,6 +244,7 @@ class Model:
             coords=self.coords,
             physics_specs=self.physics_specs,
             p0=p0*units.pascal,
+            p1=0.01*p0*units.pascal,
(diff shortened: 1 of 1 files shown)


duncanwp:
So, just to be clear, the value or 0.01 you chose in the PR reflects the green line?

PR #406, opened 2026-03-13
Fix hfluxn calculation over sea

### LINKED ISSUE #402: The calculation of hfluxn reverses the sign of sensible and latent heat fluxes over ocean
The calculations of `surface_flux.hfluxn` over land and ocean do not seem to be consistent with each other in `jcm/physics/speedy/surface_flux.py`. 

For the land surface:
https://github.com/climate-analytics-lab/jax-gcm/blob/3609a953015b7cf995ad07c6d0cd99d2ae49a803/jcm/physics/speedy/surface_flux.py#L187-L190

For the ocean surface:
https://github.com/climate-analytics-lab/jax-gcm/blob/3609a953015b7cf995ad07c6d0cd99d2ae49a803/jcm/physics/speedy/surface_flux.py#L263

The `hfluxn` is positive downward while the sensible and latent heat fluxes are positive upward. Therefore, there should be a minus sign in the surface flux calculation, as is the case on land surface flux calculation. However, this sign seems to be missing over the ocean.

jvmadan:
This fixes the signs of shf and evap for the calculation of hfluxn over ocean to reflect the fact that they are positive-up, as can be verified by the preceding lines where shf is proportional to SST - t1 and evap is proportional to qsat - q1

This was a bug in the original speedy.f90 code as well.

Closes #402


duncanwp (approved):
Good catch, thanks

PR #419, opened 2026-04-03
Add w-wind to output

jvmadan:
So far I haven't found a dinosaur library function that gets w-wind or omega directly; there's one that produces d(sigma)/dt but it would internally run compute_diagnostic_state, which is redundant with the fact that we already do that earlier in dynamics_state_to_physics_state.

* Added w-wind to PhysicsState
* Added calculation of w-wind in dynamics_state_to_physics_state
* Switched from positional to keyword args in PhysicsState initializations so they still work
* Added switch in post_process to handle the fact that dinosaur.primitive_equations.compute_diagnostic_state does not properly handle states with a time dimension
* Fixed bug in physics_interface_test

### DIFF
--- a/jcm/model.py
+++ b/jcm/model.py
@@ -410,8 +410,12 @@ class Model:
         from jcm.physics_interface import verify_state
         jax.debug.callback(lambda t: logger.info("Post processing: %s simulated seconds", t), state.sim_time)
 
+        if output_averages:
+            dynamics_fn = jax.vmap(lambda s: dynamics_state_to_physics_state(s, self.primitive))
+        else:
+            dynamics_fn = lambda s: dynamics_state_to_physics_state(s, self.primitive)
         predictions = Predictions(
-            dynamics=dynamics_state_to_physics_state(state, self.primitive),
+            dynamics=dynamics_fn(state),
             physics=None,
             times=None
         )
--- a/jcm/physics_interface.py
+++ b/jcm/physics_interface.py
@@ -247,14 +247,15 @@ def dynamics_state_to_physics_state(state: State, dynamics: PrimitiveEquations) 
     # Compute vertical velocity (code adapted and extended from dinosaur.primitive_equations.compute_vertical_velocity)
     sigma_dot_full_boundaries = jnp.pad(
         nodal_state.sigma_dot_full,
-        [(1, 1) if i == nodal_state.sigma_dot_full.ndim - 3 else (0, 0) for i in range(nodal_state.sigma_dot_full.ndim)]
+        [(1, 1) if i == nodal_state.sigma_dot_full.ndim - 3 else (0, 0) 
+         for i in range(nodal_state.sigma_dot_full.ndim)]
     ) # pad vertical dimension with boundary conditions
     sigma_dot_full_centers, sigma_centers = (
-        (a[1:] + a[:-1])/2
-        for a in (sigma_dot_full_boundaries, dynamics.coords.vertical.boundaries)
+        (a[..., 1:, :, :] + a[..., :-1, :, :]) / 2
+        for a in (sigma_dot_full_boundaries, dynamics.coords.vertical.boundaries[:, jnp.newaxis, jnp.newaxis])
     ) # get values at layer centers
     w = - (rgas * t / grav) * (
-        sigma_dot_full_centers / sigma_centers[:, jnp.newaxis, jnp.newaxis] 
+        sigma_dot_full_centers / sigma_centers
         + dynamics.nodal_log_pressure_tendency(nodal_state)
     ) # convert from sigma_dot to vertical velocity using the hydrostatic relation and ideal gas law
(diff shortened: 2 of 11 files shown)


jcm/model.py
```diff
@@ -410,8 +410,12 @@ def _post_process(self, state: primitive_equations.State, forcing: ForcingData,
         from jcm.physics_interface import verify_state
         jax.debug.callback(lambda t: logger.info("Post processing: %s simulated seconds", t), state.sim_time)
 
+        if output_averages:
```
duncanwp, line 413: Why is this necassary now?

jcm/physics_interface.py
```diff
@@ -237,7 +244,22 @@ def dynamics_state_to_physics_state(state: State, dynamics: PrimitiveEquations)
     t += dynamics.reference_temperature[:, jnp.newaxis, jnp.newaxis]
     q = dynamics.physics_specs.dimensionalize(q, units.gram / units.kilogram).m
 
-    return PhysicsState(u, v, t, q, phi, jnp.squeeze(sp, axis=-3))
+    # Compute vertical velocity (code adapted and extended from dinosaur.primitive_equations.compute_vertical_velocity)
```
duncanwp, line 247: Can we pull this out in to a function and make some unit tests? It could do with some more inline comments to explain the operations too

jcm/physics_interface_test.py
```diff
@@ -19,7 +19,7 @@ def test_initial_state_conversion(self):
         v = jnp.ones((kx, ix, il)) * -0.5
         q = jnp.ones((kx, ix, il)) * 0.5
         phi = jnp.ones((kx, ix, il)) * 5000
-        sp = jnp.ones((kx, ix, il))
+        sp = jnp.ones((ix, il))
```
duncanwp, line 22: These are good catches if they were wrong before, but try and pull them into their own PR to keep things focussed. Was this broken before? Were the tests wrong?

jcm/physics_interface_test.py
```diff
@@ -36,7 +36,7 @@ def test_initial_state_conversion(self):
             coords,
             PHYSICS_SPECS)
 
-        state = PhysicsState.zeros((kx, ix, il), u, v, temp, q, phi, sp)
+        state = PhysicsState.zeros((kx, ix, il), u_wind=u, v_wind=v, temperature=temp, specific_humidity=q, geopotential=phi, normalized_surface_pressure=sp)
```
duncanwp, line 39: Why do we need all these now?

duncanwp:
Thanks! Some comments in the review

jcm/physics_interface_test.py
jvmadan: this one was wrong before but didn't cause errors until now

jcm/model.py
jvmadan: with output averages off, post process gets run on individual dynamics states, but with it on, post process gets run on a stack of dynamics states. compute_diagnostic_state doesn't broadcast correctly over modal fields that also have a time dimension, so it has to be vmapped. i don't think this is a particularly bad change since it anyway makes it a little clearer that in one case it's running on a 4D state vs the other case is 3D. I'll add a comment

jcm/physics_interface_test.py
jvmadan: I added w to the PhysicsState, which messed up the positional arguments. I could add some filler values for w and keep positional args but switching to kwargs minimized changes to code logic

duncanwp:
Sorry, this makes me realize; why do we add w to the physics state if it's a diagnosed quantity?

jvmadan:
* It didn't make sense to add it to PhysicsData because I'm computing it at the top level physics interface (rather than in any particular physics module). This also lets it be available for all physics schemes.
* With your PR that makes the PhysicsData modular, we could have one sub-struct of physicsdata that is defined by the physics interface itself and always included, for storing diagnostic quantities computed by the physics interface. This would be useful especially if we add more quantities like this.
* PhysicsState is currently forced to serve the above purpose at the same time as being the nodal space version of the dinosaur state.
* This seems fine to me except for one thing: we do accept PhysicsStates from users as a valid format for initial state. Having new diagnostic variables in PhysicsState is definitely confusing because they will have no impact on the state when it's converted to modal and fed to the model.
* The thing is, I think this was already the case for geopotential--it's in the dynamics state, but is diagnostic within the dycore, so it should have no effect as part of an initial state.
* I agree this is a problem though, until your composability PR is merged I think the solution would be either creating separate structs for InputPhysicsState and OutputPhysicsState, or putting a redundant w-wind calculation and extra field into the PhysicsData for each physics scheme where we'd want w-wind.

duncanwp:
Hmm, OK. Can this just be calculated offline since we return the full physics state? Maybe we just make it a util that can be used for post-processing until we have a physics scheme that actually needs it?

PR #439, opened 2026-04-22
Fix ICON hybrid coordinate support and loosen NaN guards

duncanwp:
# Fix ICON hybrid coordinate support and loosen NaN guards

## Summary

- Fixes ICON `HybridCoordinates` end-to-end (model dispatch, geopotential, pressure
  calculation, xarray output) — was crashing within 1-2 steps due to
  sigma-only code paths and unit mismatches.
- Loosens the NaN guards we accumulated while chasing CFL problems. The
  strict caps were masking bugs in convection / cloud physics that we now
  plan to tackle separately.
- Adds 18 regression tests (icon_coords, hybrid physics interface,
  verify_state / verify_tendencies).

## Hybrid coord fixes

| File | Change |
|------|--------|
| `jcm/model.py` | Dispatches `PrimitiveEquationsHybrid` for hybrid coords, with `hpa_quantity=units.pascal` (ICON stores `a` in Pa, dinosaur defaults to hPa). Also skips the `log_surface_pressure` normalization by p0 on the hybrid path, so `exp(log_sp) = P_s_actual` matches the units of `a_boundaries`. |
| `jcm/physics_interface.py` | `dynamics_state_to_physics_state` now calls `compute_diagnostic_state_hybrid` and `get_geopotential_on_hybrid` with the actual spatially-varying P_s (not a reference scalar). For the hybrid path, `sp` is divided by `p0_nondim` so `PhysicsState.normalized_surface_pressure` stays on the `P_s/p0 ≈ 1` scale that downstream physics expects. |
| `jcm/physics/icon/icon_coords.py` | Stores `a_full/b_full/a_half/b_half` and exposes `calculate_pressure_full/half` methods implementing `p = a + b·P_s` (sigma → a=0, b=sigma). Keeps `fsg`/`hsg` as the sigma-equivalent at a reference p0 for legacy consumers. |
| `jcm/physics/icon/unit_conversions.py` + `icon_physics.py` | Use the new coord methods instead of `sigma × P_s`. |
| `jcm/utils.py:data_to_xarray` | Falls back to `vertical.get_sigma_centers(p0)` when `.centers` isn't exposed (HybridCoordinates). |

## Loosened NaN guards

Keep only math-safety: divide-by-zero, log/sqrt/exp overflow. No physical-range bounds.

- `verify_state`: `q >= 0` only, no upper cap.
- `verify_tendencies`: `q_next >= 0` only, no upper cap.
- `radiation_scheme.py`: positivity for pressure/density/thickness/cloud fields; `q ∈ [0, 0.99]` for the `q/(1-q)` conversion (required for math safety only).
- T clips in `gas_optics.py`, `tiedtke_nordeng.py`, `icon_physics.py`, `turbulent_fluxes.py` widened from [100, 400]K or [150, 350]K to [50, 500]K — only catches truly pathological values.
- Two-stream denom / `tau/mu0` overflow guards: kept.
- Tridiagonal matrix pivot / log guards: kept.

## Test plan

- [x] `pytest jcm/physics_interface_test.py jcm/physics_interface_hybrid_test.py jcm/physics/icon/icon_coords_test.py` — 18 passed
- [x] `pytest jcm/physics/icon/` (fast subset) — 408 passed, 1 skipped
- [x] T85 hybrid + JW ref T + dt=3min: 90 days clean with realistic T (210-296K), q (0-17 g/kg), |u| (0-128 m/s)
- [x] T31 hybrid 1-day unit test: no NaN
- [x] Hybrid 3-year run: 270 days clean before an extreme wind event (|u|=120 m/s) triggers CFL — separate convection/diffusion work to follow

## Known limitations

- Still no precipitation / CAPE in RCE-style output — traced to missing iterative saturation adjustment, organized entrainment, and dynamic LNB in the Tiedtke-Nordeng port. Being fixed on `fix/convection-rce`.
- `SpeedyPhysics` and `HeldSuarezPhysics` still assume sigma coords (`vertical.centers`). Out of scope for this PR.

### DIFF
--- a/jcm/physics/icon/icon_coords.py
+++ b/jcm/physics/icon/icon_coords.py
@@ -3,30 +3,52 @@
 This module defines the IconCoords struct that caches precomputed
 coordinate data needed by ICON physics parameterizations, following
 the same pattern as SpeedyCoords for SPEEDY physics.
+
+The coordinates are stored in the hybrid (a, b) form so that the
+hybrid-coordinate pressure relation
+
+    p(k, col) = a(k) + b(k) * P_s(col)
+
+can be evaluated exactly at every column. Pure sigma coordinates are a
+special case with a = 0 and b = sigma. `fsg`/`hsg` are kept as convenience
+attributes equal to the sigma-equivalent at a reference surface pressure,
+for existing call sites that only need a rough sigma profile.
 """
 
 import jax.numpy as jnp
+import numpy as np
 import tree_math
 from dinosaur.coordinate_systems import CoordinateSystem
+
+
+# Reference surface pressure used to compute sigma-equivalent for hybrid coords.
+_P_S_REF_PA = 101325.0
 
 
 @tree_math.struct
 class IconCoords:
     """ICON-specific coordinate system data.
 
-    This struct caches precomputed coordinate data needed by ICON physics.
     All fields are constant during a simulation.
 
     Attributes:
         nodal_shape: Grid dimensions (nlev, nlon, nlat)
-        fsg: Sigma coordinates at level centers (nlev,)
-        hsg: Sigma coordinates at half levels / boundaries (nlev+1,)
+        a_full: Hybrid 'a' coefficient at level centers (Pa) [nlev]
+        b_full: Hybrid 'b' coefficient at level centers [-] [nlev]
+        a_half: Hybrid 'a' coefficient at half levels (Pa) [nlev+1]
+        b_half: Hybrid 'b' coefficient at half levels [-] [nlev+1]
+        fsg: Sigma-equivalent at level centers (a_full/P_s_ref + b_full) [nlev]
+        hsg: Sigma-equivalent at half levels (a_half/P_s_ref + b_half) [nlev+1]
         lat: Latitude in radians
         lon: Longitude in radians
 
     """
 
     nodal_shape: tuple
+    a_full: jnp.ndarray
+    b_full: jnp.ndarray
+    a_half: jnp.ndarray
+    b_half: jnp.ndarray
     fsg: jnp.ndarray
     hsg: jnp.ndarray
     lat: jnp.ndarray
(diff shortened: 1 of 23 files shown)


jcm/physics/icon/icon_coords.py
```diff
@@ -3,30 +3,51 @@
...
+for existing call sites that only need a rough sigma profile.
 """
 
 import jax.numpy as jnp
 import tree_math
 from dinosaur.coordinate_systems import CoordinateSystem
 
 
+# Reference surface pressure used to compute sigma-equivalent for hybrid coords.
+_P_S_REF_PA = 101325.0
```
duncanwp: This should probably use the constants.py reference?

jcm/physics/icon/radiation/constants.py
```diff
@@ -2,32 +2,29 @@
...
 """
 
-# Enhanced spectral resolution for improved accuracy
-N_SW_BANDS = 6  # Shortwave bands (UV, visible, near-IR)
-N_LW_BANDS = 8  # Longwave bands (far-IR, window, near-IR)
-N_BANDS_TOTAL = N_SW_BANDS + N_LW_BANDS  # Total bands
+# Spectral bands — must match RadiationParameters.default()
+N_SW_BANDS = 2  # Shortwave bands
+N_LW_BANDS = 3  # Longwave bands
+N_BANDS_TOTAL = N_SW_BANDS + N_LW_BANDS
```
duncanwp: Can we somehow have them derive from the radiation parameters then?

jcm/physics/icon/radiation/rce_test.py
```diff
@@ -138,8 +145,10 @@ def test_radiative_equilibrium_converges(self):
         pressure = atm["pressure_levels"]
         pressure_interfaces = atm["pressure_interfaces"]
 
-        dt = 86400.0  # 1 day
-        n_steps = 15
+        # dt = 30 min keeps forward-Euler stable at the thin TOA layers under
+        # the current 3-band LW radiation tuning. See class docstring.
+        dt = 1800.0
```
duncanwp, line 150: THis might now need to be marked as slow

jcm/physics/icon/icon_coords.py
```diff
@@ -36,21 +57,88 @@ def xarray_additional_coords(self):
...
+
         return cls(
             nodal_shape=coords.nodal_shape,
-            fsg=jnp.asarray(coords.vertical.centers),
-            hsg=jnp.asarray(coords.vertical.boundaries),
+            a_full=a_full,
+            b_full=b_full,
+            a_half=a_half,
+            b_half=b_half,
+            fsg=fsg,
```
duncanwp: I don't actually think we need fsg and hsg any more do we?

jcm/utils.py
```diff
@@ -217,12 +217,21 @@ def _infer_dims_shape_and_coords(
...
     lon_k, lat_k = coords.horizontal.modal_axes  # k stands for wavenumbers
     lon, sin_lat = coords.horizontal.nodal_axes
+
+    # HybridCoordinates uses `get_sigma_centers(p_ref)` and doesn't expose
+    # a .centers attribute; fall back to that for xarray's level axis.
+    vertical = coords.vertical
+    if hasattr(vertical, 'centers'):
+        level_coords = vertical.centers
+    else:
+        level_coords = np.asarray(vertical.get_sigma_centers(101325.0))
```
duncanwp: Use the reference surface pressure constant

jcm/physics/icon/icon_coords.py
duncanwp: Done — now imports `p0` from `icon.constants.physical_constants`. Also deleted the field entirely since it was only used to populate the unused `fsg`/`hsg` attributes (see separate reply below).

jcm/physics/icon/radiation/constants.py
duncanwp: Done. `SW_BAND_LIMITS` / `LW_BAND_LIMITS` in this file are now the single source of truth — `N_SW_BANDS = len(SW_BAND_LIMITS)` and `RadiationParameters.default()` imports from here. Keep the counts as Python ints because they gate static shapes/loop unrolls in gas_optics/planck/cloud_optics.

jcm/physics/icon/radiation/rce_test.py
duncanwp: The class is already marked with `@pytest.mark.slow` at the decorator line just above the class (line 130). With the reduced dt each of the 4 tests now runs in a few seconds on CPU, but I've left the slow marker because first-call JAX compilation still pushes the file to 7+ min on a cold CI runner.

jcm/physics/icon/icon_coords.py
duncanwp: You're right — grepped and nothing in ICON reads `.fsg`/`.hsg` (SPEEDY's `speedy_coords` is a separate struct). Removed the fields and the `_P_S_REF_PA` constant that was only used to compute them.

jcm/utils.py
duncanwp: Done — uses `p0` from `jcm.physics.icon.constants.physical_constants` (= 1e5 Pa, ICON's reference). Only affects the xarray level-axis labels so the small numerical difference vs. 101325 Pa is cosmetic.

Issue #173, opened 2025-07-04
Create a parameter configuration file


duncanwp:
Add the parameters file to Hydra to allow easy changing of the physics parameters per run.

This should be easy, but it's not obvious how this works once we offer different physics packages...

duncanwp:
Fixed by #447

Issue #375, opened 2026-02-06
Update the command line interface to accept all Model and run arguments


duncanwp:
Users should be able to specify all Model params, especially e.g. boundary and terrain files via the CLI. This will enable easier scripting of simulations across different compute environments and make longer runs easier.

Note, Physics parameters still need to be specified through a config file probably: https://github.com/climate-analytics-lab/jax-gcm/issues/173

duncanwp:
Closed by #447

Issue #192, opened 2025-07-25
Single Column Model


duncanwp:
It would be useful to have a single column `Model` that just runs the parameterizations over one column with a prescribed large-scale forcing (dynamic tendencies). This is commonly used for calibrating physics over specific regions of interest or observation sites and would be be useful for testing. It shouldn't be too hard, it's just a separate model class without the dycore, so it needs its own timestepping, but I also need the large-scale tendencies to drive it. 

A key requirement would be to output the large-scale tendencies from a vanilla JCM run to be able to use that to drive a single column.

It would be _nice_ to be able to run multiple (independent) columns in parallel if it's easy.

duncanwp:
Closed by #447

PR #447, opened 2026-04-29
Address #445: expand main.py CLI, modernise SCM (PR #366), refactor utils/ runners

### LINKED ISSUE #445: Update the command line interface
A bunch of my integration tests have ended up working around /reengineering the command line interface. Add the various switches and choices to the CLI to make it more useful for integration testing (which will test it more, and make it more useful for users)

duncanwp:
## Summary

Bundles three coupled changes that issue #445 asked for, plus the SCM revival
the user explicitly called out:

1. **Hydra config groups** under \`jcm/config/{physics,grid,run,init,terrain,
   diffusion}\` so the CLI covers the option surface that \`utils/run_*.py\`
   had grown around. Examples in \`jcm/main.py\` and the regenerated
   \`hydra/help/custom_help.yaml\`.
2. **\`jcm.runners\`** -- builders (\`build_coords\`, \`build_physics\`,
   \`build_terrain\`, \`build_diffusion\`, \`build_model\`) plus
   \`inject_jw_profile\` and \`run_chunked\` (chunked long-run with health
   checks). Importable by tests/notebooks without going through Hydra.
3. **PR #366 brought forward**: \`SingleColumnModel\` and
   \`PrescribedStateModel\` ported to the current
   \`CoordinateSystem\` + \`TerrainData\` API (\`Geometry\` is gone). Helpers
   \`load_states_from_xarray\`, \`create_single_column_state\`,
   \`create_initial_tracers\` added to \`jcm.utils\`.
4. **Harness cleanup**: \`utils/run_hs_90d.py\` removed (CLI replaces it),
   \`utils/run_icon_simulation.py\` reduced to plot helpers,
   \`utils/run_icon_longrun.py\` rewritten to compose a Hydra cfg and call
   \`run_chunked\`. \`jcm.diagnostics\` houses \`check_health\` /
   \`print_report\`.

## Tests

* \`jcm/runners_test.py\` -- composition + builder + smoke-run tests
  (Held-Suarez T31 and SPEEDY default).
* \`jcm/single_column_model_test.py\` -- Held-Suarez and ICON-grey on a
  small sigma grid; new SCM helpers.
* \`jcm/prescribed_state_model_test.py\` -- Held-Suarez vmap path.
* \`jcm/diagnostics_test.py\` -- happy + failure modes for \`check_health\`.

Local: 34 pass + 1 skipped on \`runners_test, diagnostics_test,
single_column_model_test, prescribed_state_model_test, model_test\` (T31 SPEEDY
and Held-Suarez).

## CLI shape

\`\`\`bash
# Default 10-day SPEEDY aquaplanet run
python -m jcm.main

# ICON physics on T85x47 hybrid coords with grey radiation
python -m jcm.main physics=icon grid=icon_t85_l47_hybrid

# Held-Suarez 30-day integration
python -m jcm.main physics=held_suarez grid=held_suarez_t31_l8 \\
    run.total_time=30 run.save_interval=1

# 1-day smoke test (CI-friendly)
python -m jcm.main run=smoke

# JW initial state with sponge enabled
python -m jcm.main physics=icon grid=icon_t85_l47_hybrid init=jw \\
    run.sponge.levels=5

# Multirun sweep
python -m jcm.main -m run.time_step=10,20,30
\`\`\`

## Test plan

- [ ] \`pytest -n 12 --cov=jcm --cov-fail-under=80\` passes on CI
- [ ] \`python -m jcm.main --help\` shows the new help template
- [ ] \`python -m jcm.main physics=held_suarez grid=held_suarez_t31_l8\`
      writes a netCDF
- [ ] No regressions in existing \`jcm/model_test.py\`

Closes #445.
Supersedes #366.


.github/workflows/run_test.yaml
```diff
@@ -31,14 +31,14 @@ jobs:
...
         run: |
-          pytest -v -s -m "not slow" --cov=jcm --cov-fail-under=90
+          JAX_PLATFORMS=cpu pytest -n auto -m "not slow" --cov=jcm --cov-fail-under=90
 
-      # run slow tests when triggered by a push to an open PR
+      # run the full suite (fast + slow) when triggered by a PR
       - name: Test with pytest (include slow tests on PR)
         if: github.event_name == 'pull_request'
         run: |
-          pytest -v -s -m "slow" --cov=jcm --cov-fail-under=80
```
duncanwp, line 44: Why did you remove the -m "slow" flag, we don't need to run the fast tests again, this setup runs the fast and slow tests in parallel on making a PR

jcm/diagnostics.py
```diff
@@ -0,0 +1,129 @@
...
+        report["sfc_T_mean"] = float(np.nanmean(sfc))
+
+    if "convection.precip_conv" in ds:
+        precip = ds["convection.precip_conv"].isel(time=-1).values
+        report["precip_conv_mean_mmday"] = float(np.nanmean(precip)) * 86400
+        report["precip_conv_max_mmday"] = float(np.nanmax(precip)) * 86400
+
+    ok = True
+    reasons: list[str] = []
+    if report.get("T_nan_frac", 0) > 0.1:
```
duncanwp: This should be greater than zero shouldn't it?

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+        f"Unknown grid.vertical={vertical!r}; "
+        "expected one of: speedy, held_suarez, icon_hybrid, sigma"
+    )
+
+
+# ---------------------------------------------------------------------------
+# Physics
+# ---------------------------------------------------------------------------
+
+def build_physics(cfg: DictConfig):
```
duncanwp, line 119: This is neat, but we should allow the hydra config to also define all the relevant physics parameters too. The config shouldn't redefine them - it should just rely on the respective Parameter definitions and allow the user to specify them individually (or not)

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+        ``held_suarez`` -- Held-Suarez sigma boundaries
+        ``icon_hybrid`` -- ICON hybrid (a + b·P_s) coordinates
+        ``sigma``       -- equidistant sigma coordinates
+    """
+    grid = cfg.grid
+    layers = grid.layers
+    truncation = grid.spectral_truncation
+
+    vertical = grid.vertical
+    if vertical == "speedy":
```
duncanwp: The vertical coordinate system should, as far as possible be independent of the physics scheme, it should just be sigma or hybrid and then a number of layers. SPEEDY only works with a specific number of layers, but it should be up to the physics scheme to raise an error if it doesn't like the number of layers. 

For ICON, there are a couple of 'pre-configured' hybrid coordinates that we should use if we're given the correct number of levels, but really we should just make the coordinate system using whichever number of levels is specified and if it's not a 'standard' number then print something to log.INFO

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Initial state injection (JW-style lapse-rate profile)
+# ---------------------------------------------------------------------------
+
+def inject_jw_profile(model: Model) -> None:
+    """Replace the default isothermal init with a JW-style lapse-rate profile.
+
+    Mirrors ``utils/run_icon_simulation.py:inject_realistic_profile`` so that
```
duncanwp: We will remove that profile so remove that reference from this comment and make it more explicit about what it does

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+    from dinosaur.hybrid_coordinates import HybridCoordinates
+    from dinosaur.scales import units
+
+    model._final_modal_state = model._prepare_initial_modal_state(
+        physics_state=None, random_seed=0,
+    )
+    state = model._final_modal_state
+
+    nlon, nlat = model.coords.horizontal.nodal_shape
+    p0_pa = 101325.0
```
duncanwp: This should be pulled from constants probably (or the coordinate system itself?)

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+    if isinstance(model.coords.vertical, HybridCoordinates):
+        sigma = jnp.asarray(model.coords.vertical.get_sigma_centers(p0_pa))
+    else:
+        sigma = jnp.asarray(model.coords.vertical.centers)
+    nlev = sigma.size
+
+    # Standard-atmosphere T(sigma); cap the cold tail so the semi-implicit
+    # reference temperature stays close.
+    p = sigma * p0_pa
+    T_sfc, gamma = 288.0, 6.5e-3
```
duncanwp: I think we define a surface temperature somewhere in constants/parameters

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+    z = 8400.0 * jnp.log(p0_pa / p)
```
duncanwp: What is 8400?

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+    T_profile = jnp.maximum(T_sfc - gamma * z, 250.0)
+
+    # Hydrostatic-balance the surface pressure when there's nontrivial
+    # orography, otherwise the isothermal-rest init produces air below ground.
+    orog = jnp.asarray(model.terrain.orog)
+    if jnp.any(orog > 1.0):
+        Rd, grav, T_ref_avg = 287.04, 9.80665, 260.0
```
duncanwp: Lots of magic numbers here. Again, pull from constants / parameters

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+
+    T_ref = jnp.asarray(model.primitive.reference_temperature)
+    T_var_profile = T_profile - T_ref
+    T_var_nodal = jnp.broadcast_to(
+        T_var_profile[:, None, None], (nlev, nlon, nlat)
+    ).astype(state.temperature_variation.dtype)
+    state.temperature_variation = model.coords.horizontal.to_modal(T_var_nodal)
+
+    # Humidity: 60% RH below 200 hPa
+    es = 611.2 * jnp.exp(17.67 * (T_profile - 273.15) / (T_profile - 29.65))
```
duncanwp: Explain some of these numbers. Pull constants out and maybe make RH a parameter to the function?

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+        time_step=cfg.run.time_step,
+        log_level=log_level,
+    )
+
+
+# ---------------------------------------------------------------------------
+# Run + save
+# ---------------------------------------------------------------------------
+
+def run(cfg: DictConfig, model: Model | None = None) -> ModelPredictions:
```
duncanwp: Add the option to pass in a specified forcing to the run. The user should be able to specify a file to use (which can be read using the forcing.from_file method)

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+        inject_jw_profile(model)
+        return model.resume(
+            save_interval=cfg.run.save_interval,
+            total_time=cfg.run.total_time,
+            output_averages=cfg.run.output_averages,
+        )
+    raise ValueError(f"Unknown init.kind={cfg.init.kind!r}")
+
+
+def run_chunked(
```
duncanwp, line 520: This should also accept a specified forcing

jcm/single_column_model.py
```diff
@@ -0,0 +1,355 @@
...
+            Listed prognostic variables (``u_wind``, ``v_wind``, ``temperature``,
+            ``specific_humidity``) are nudged toward the prescribed state with
+            timescale ``tau`` while still receiving their physics tendency.
+
+    """
+
+    def __init__(
+        self,
+        physics: Physics,
+        coords: CoordinateSystem,
```
duncanwp: By definition, this shouldn't need a horizontal coordinate system (it's just one column). It needs a vertical coordinate system, but that's all. It would be nice to run multiple columns in parallel, but that should happen one layer up because the columns don't need to be adjacent

jcm/runners.py
```diff
@@ -0,0 +1,378 @@
...
+        "specific_humidity": model.coords.horizontal.to_modal(q_nodal),
+    }
+    model._final_modal_state = state
+
+
+# ---------------------------------------------------------------------------
+# Top-level model construction
+# ---------------------------------------------------------------------------
+
+def build_model(cfg: DictConfig) -> Model:
```
duncanwp, line 318: Since we're bringing in the SCM in this PR , let's also add the option of running prescribed state or single column mode too. For the prescribed state the user will have to specify a file to read the state from (we can assume came from a JCM run), and for the SCM they will additionally have to specify which column(s) (location[s]) to run (which can happen in parallel). These will need their own runners.

jcm/config/config.yaml
duncanwp, line 1: Also need to add a way of speciyfing the SPMD mesh (for sharding). Not sure at what level that best lives

jcm/runners.py
duncanwp: You already have the util helpers for actually extracting the state from an xarray object which is a good start

utils/run_icon_longrun.py
duncanwp, line 1: This could (should?) now be able to just be run from the CLI, so lets just add a couple of relevant examples to the documentation and delete this script

utils/run_icon_simulation.py
duncanwp, line 1: This is almost all redundant now, lets delete it. If you want to you can keep the plot helpers locally

duncanwp:
Closes #445 and #375 and #192

duncanwp:
Closes #173 too

Issue #437, opened 2026-04-21
No time-varying plume amplitude in ICON Simple Plumes scheme


aoifemcquaid:
**Bug / Missing Feature**                                                                                                          
                                                                                                                                 
  The JCM MACv2-SP aerosol scheme (simple_aerosol.py) does not support time-varying plume amplitudes, unlike the original Fortran
   scheme (mo_simple_plumes_v1.f90) it is meant to translate.
                                                                                                                                 
  What the original does:                                                                                                        
   
  The Fortran scheme reads two time-varying fields from MACv2.0-SP_v1.nc:                                                        
  - year_weight — shape (251, 9): scales each of the 9 plumes by year (1850–2100)
  - ann_cycle — shape (2, 52, 9): applies a seasonal cycle (52 weekly bins × 2 features)                                         
                                                                                        
  At each call to sp_aop_profile(year_fr, ...) it computes:                                                                      
  iyear = FLOOR(year_fr) - 1849    ! index into year_weight                                                                      
  iweek = FLOOR(frac(year_fr)*52)+1 ! index into ann_cycle                                                                       
  time_weight(f,p) = year_weight(iyear,p) * ann_cycle(f,iweek,p)                                                                 
  The plume amplitude cw_an = time_weight * ftr_weight * spatial_gaussian * aod_spmx thus varies by both year and season.        
                                                                                                                                 
  What the JCM version does:                                                                                                     
                                                                                                                                 
  - ann_cycle and year_weight are stored as flat 1D arrays of shape (9,), initialized to all-ones in AerosolParameters.default().
  - get_anthropogenic_aod() and get_background_aod() have no year_fr argument and no time-indexing logic.                        
  - Plume amplitudes are static — identical at every model timestep regardless of year or season.     
  
https://github.com/climate-analytics-lab/jax-gcm/blob/076e2e0075f51d5f3c3c7ea408c42facd8b136b5/jcm/physics/icon/aerosol/aerosol_params.py#L186-L188                           
                                                                                                                                 
  **Impact:**                                                                                                                        
                                                                                                                                 
  Simulations will use present-day emission levels year-round, with no representation of:                                        
  - Historical emission trends (e.g. the strong aerosol loading in the mid-20th century)
  - Seasonal cycles (e.g. biomass burning peaks, wintertime heating emissions)                                                   
  - Future emission scenarios (RCP projections encoded in year_weight)   
  
  Cannot use plume amplitude for gradient descent (aim of future work).     
[...]

Issue #410, opened 2026-03-24
Make timing of solar forcing accurate


jvmadan:
Currently jcm computes the current phase of the seasonal cycle from the fraction of year elapsed, without taking into account what year it is. This is correct for a no-leap-days calendar, but now that we have leap days, we would need to pass the year to the solar function as well, and use the year mod 400, mod 100, mod 4 etc to compute the minor offset to the solar radiation due to leap days. A better solution might be to draw on whatever NeuralGCM uses, which if it computes accurate solar forcing from a jax.datetime would be ideal for our needs.

meteorologytoday:
I happened to bump into the same problem playing around with idealized setups. I asked ClaudeAI, and it suggested a JAX-compatible code (see below). The most important thing is the Fliegel & Van Flandern algorithm:

- Ref 1 (original paper): [Fliegel, H. F., & Van Flandern, T. C. (1968). Letters to the editor: a machine algorithm for processing calendar dates. Communications of the ACM, 11(10), 657.](https://dl.acm.org/doi/pdf/10.1145/364096.364097) 
- Ref 2: [A One-Line Algorithm for Julian Date by Fernie, J. D.](https://adsabs.harvard.edu/full/1983IAPPP..13...16F)
- Ref 3: [Converting Between Julian Dates and Gregorian Calendar Dates](https://aa.usno.navy.mil/faq/JD_formula)

# AI-generated code (needs verification)
```
import jax
import jax.numpy as jnp
import jax_datetime as jdt

# Unix epoch is 1970-01-01, which is day 2440588 in Julian Day Number
UNIX_EPOCH_JDN = 2440588
SECONDS_IN_DAY = 86_400

def days_to_ymd(days_since_epoch):
    """Convert days since Unix epoch to (year, month, day), JAX-compatible."""
    jdn = days_since_epoch + UNIX_EPOCH_JDN

    # Julian Day Number -> Gregorian calendar (Fliegel & Van Flandern algorithm)
    l = jdn + 68569
    n = (4 * l) // 146097
    l = l - (146097 * n + 3) // 4
    i = (4000 * (l + 1)) // 1461001
    l = l - (1461 * i) // 4 + 31
    j = (80 * l) // 2447
    day = l - (2447 * j) // 80
    l = j // 11
    month = j + 2 - 12 * l
    year = 100 * (n - 49) + i + l

    return year, month, day

def is_leap_year(year):
[...]

Issue #363, opened 2026-01-14
Provide an option to align forcing and boundary conditions based on date


eldavenport:
Right now we ignore the dates associated with forcing (i.e. forcing is technically 1982 but we treat it like its just a generic set of 365 days). It should be fine to do something like this so that we can do idealized simulations repeating the same annual forcing, but we probably want an option to actually align these things if you want to do more specific simulations on shorter timescales.

eldavenport:
As it is right now, we default the model date to 2000-01-01 and the forcing dates are 1982-01-01. So we completely ignore the fact that there are dates in both and they don't agree. If you run for long enough, the model will do leap days and the forcing only has 365 values so you are then forcing Jan 01 with Jan 02. In an idealized case like speedy it probably doesn’t matter. But I was thinking for the future it would be nice to clarify or fix it.

We could get rid of the forcing dates, and say that forcing days are just assumed to align with the model days. That gets kind of annoying for uses when you do care about specific calendar dates in the forcing. Then you always have to slice out the right days to be sure that the start date/end of the model is the same start/end date in the forcing.  

If we have the option to get forcing values based on the model date then you can have 10 years of ERA5 data and start your model on which ever day you want with no changes. Also helps if you have multiple data sets/sources. They just need to cover the desired time period, and you don’t have to reslice them all for each new experiment.

duncanwp:
Closed by #448

Issue #352, opened 2025-12-19
Core: DateData design - dt should be required parameter


duncanwp:
## Description
The DateData class design has a FIXME suggesting dt should be required with tyear and model_year derived from it.

## Location
`jcm/date.py:21`

## Current Code
```python
# FIXME: dt should be required and tyear and model_year derived from it (as properties)
```

## Suggested Change
Make `dt` a required parameter and compute `tyear` and `model_year` as derived properties.

## Impact
Would simplify the interface and ensure consistency between time representations.

duncanwp:
Fixed by #451 and #448

PR #448, opened 2026-04-30
Date-aware forcing: physics no longer cares about the calendar

### LINKED ISSUE #285: Move co2 concentration into boundary conditions
Currently it's included as an optional 1% increase but this should just be a boundary condition that can be provided as either a timeseries or constant

duncanwp:
## Summary

Refactor the JAX-GCM forcing/date system so physics terms stop reading `DateData` directly (apart from the `model_step` counter), and the `Model` aligns time-varying forcing files to its calendar. Six focused commits, each individually reviewable; all 770 unit tests pass.

Closes #285, #308, #352, #363, #410, #437. Adds calendar-string durations + xarray resample as a small follow-up. #287 (CFTime calendars) is intentionally out of scope — the issue is replaced with a `'365_day'` / `'gregorian'` switch on `Model`.

## What changed, by commit

1. **`forcing: add TimeSeries / SolarGeometry / co2_vmr foundation`** — pure plumbing.
   - `TimeSeries` leaf wrapper carries `values + time_seconds + align_mode` (`WRAP_YEAR` vs `BY_DATE`).
   - `SolarGeometry` precomputes `tyear / orbital_phase / synodic_phase`.
   - `ForcingData.select(date, calendar)` collapses every TimeSeries leaf to the current step's slice and populates `solar`. No-op for static-only forcing.
   - `Model` gains a `calendar` arg (`'365_day'` default for SPEEDY, `'gregorian'` available); calls `forcing.select` once per step inside `_get_step_fn_factory` and `_post_process`.

2. **`CO2 in BCs`** (#285) — SPEEDY `ablco2` and ICON CO₂ both read `forcing.co2_vmr` (linear-in-ratio mapping anchored at 360 ppmv). `ForcingParameters.increase_co2 / co2_year_ref` deleted. Regression test pins the legacy `exp(0.005 * dyears)` ramp by reproducing it with the equivalent CO₂ trajectory.

3. **`multi-year, date-aligned files; pre-slice physics view`** (#308 + #363) — `from_file` accepts arbitrary time lengths and wraps time-varying variables as `TimeSeries`. `auto` align mode picks `wrap_year` for ≤1-year files and `by_date` for longer spans. Day-of-year slicing in `SpeedyForcing.__call__` deleted (Model already pre-sliced); five `if forcing.X.ndim == 3 ...` shape branches in ICON deleted.

4. **`solar carve-out — physics is now date-free`** — radiation schemes (grey-two-stream, NN emulator, RRTMGP, SPEEDY shortwave) take `SolarGeometry` instead of `Datetime`. After this commit no physics term references `DateData` calendar fields; the only date-y behavior left is the `model_step` counter.

5. **`restore MACv2-SP feature axis`** (#437) — `get_anthropogenic_aod` / `get_background_aod` accept either 1-D `(nplumes,)` (legacy placeholder, unchanged path) or 2-D `(nfeatures, nplumes)` (proper post-time-slice shape). Per-feature plume gaussian extracted as a helper. Wiring the multi-time MACv2.0-SP_v1.nc loader plus its `(nyears, nplumes)` / `(nfeatures, nweeks, nplumes)` `TimeSeries` is left to a follow-up; the file is not bundled (obtain from MPI-M).

6. **`calendar-string durations + xarray resample helper`** — small ergonomic follow-up.
   - `Model.run(save_interval='1 month', total_time='1 year')` works; strings parsed via `parse_duration_days(value, calendar)`.
   - `predictions.resample('1MS').mean()` returns a calendar-aligned monthly trajectory by post-resampling daily output. Integrator stays fixed-cadence (its `nnx.scan` lengths are static at JIT time).

## Tests

- `JAX_PLATFORMS=cpu pytest -n 12 --ignore=jcm/physics/radiation/rrtmgp_test.py`: 770 passed, 13 skipped. (`rrtmgp_test.py` skipped because the optional `rrtmgp` Python package isn't installed in this env — pre-existing.)
- New tests:
  - `forcing_test.py`: `TimeSeries` `WRAP_YEAR`/`BY_DATE` indexing, JIT compatibility, multi-year `from_file`.
  - `physics/forcing/speedy_forcing_test.py`: `ablco2 = f(co2_vmr)` regression vs the legacy formula.
  - `physics/aerosol/macv2_sp_test.py`: 1-D vs 2-D `ann_cycle` agreement; per-feature zeroing reduces AOD.
  - `date_test.py`: calendar-string parsing.
  - `model_test.py`: end-to-end `model.run('2 months')` and `predictions.resample('1MS').mean()`.

## Test plan

- [ ] Spot-check the SPEEDY ablco2 mapping form (`ablco2 ∝ co2_vmr / 360 ppmv`) — happy to swap to log-in-ratio if you'd prefer; the regression test pins whichever shape we land on.
- [ ] Confirm the MACv2-SP feature-axis collapse (`ftr_weight`-weighted reduction over the feature dim) matches your intent — the original Fortran multiplies feature-by-feature inside the spatial sum, which we approximate here because the JAX port collapses features in the spatial gaussian via `ftr_weight` already. Equivalent under uniform `ftr_weight`; the open detail is what to do under non-uniform feature weights.
- [ ] Decide whether to follow up with the bundled MACv2.0-SP_v1.nc loader (separate PR) or leave that for a downstream user contribution.

### DIFF
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -226,42 +226,39 @@ duration of the run.
 Long forcing time-series and chunked runs
 -----------------------------------------
 
-Time-varying forcing is captured at JIT compile time as a closure constant
-(see ``Model._get_step_fn_factory``). Multi-year hourly forcing is therefore
-tractable for runs of a few simulated years; longer or finer-cadence
-forcing should either be resampled before loading, or fed to the model in
-chunks via ``Model.resume``. The chunked pattern avoids inflating the JIT
-closure and lets you stream output to disk year by year:
+For multi-year forcing files, it's often convenient to run the model one
+year at a time. This keeps memory bounded and lets you save output as you
+go. Use ``xarray.Dataset.groupby('time.year')`` to slice the forcing,
+then ``Model.run`` for the first year and ``Model.resume`` for subsequent
+years to continue from the previous state:
 
 .. code-block:: python
 
    import xarray as xr
    from jcm.forcing import ForcingData
 
-   # Multi-decade ERA5-style forcing with real timestamps.
    ds = xr.open_dataset('era5_1980_2010.nc')
+   yearly_outputs = []
 
-   # Run one calendar year at a time.
-   for year in range(1980, 2011):
-       year_slice = ds.sel(time=str(year))
-       year_slice.to_netcdf(f"_year_{year}.nc")
-       forcing = ForcingData.from_file(f"_year_{year}.nc", coords=coords)
+   year_iter = iter(ds.groupby('time.year'))
 
-       if year == 1980:
-           preds = model.run(forcing=forcing, save_interval='1 day',
-                             total_time='1 year')
-       else:
-           preds = model.resume(forcing=forcing, save_interval='1 day',
-                                total_time='1 year')
+   year, year_ds = next(year_iter)
+   forcing = ForcingData.from_dataset(year_ds, coords=coords)
+   preds = model.run(forcing=forcing, save_interval='1 day',
+                     total_time='1 year')
+   yearly_outputs.append(preds.to_xarray())
 
-       preds.to_xarray().to_netcdf(f"output_{year}.nc")
+   for year, year_ds in year_iter:
+       forcing = ForcingData.from_dataset(year_ds, coords=coords)
+       preds = model.resume(forcing=forcing, save_interval='1 day',
+                            total_time='1 year')
+       yearly_outputs.append(preds.to_xarray())
 
-Each ``run``/``resume`` call rebuilds the JIT-compiled step function for
-the new forcing closure (so the first iteration is slow as JAX compiles,
-subsequent iterations reuse the same compilation cache key when the
-forcing's pytree structure is unchanged). For really long campaigns,
-prefer this chunked pattern over loading the entire forcing record into
-one ``ForcingData``.
+   trajectory = xr.concat(yearly_outputs, dim='time')
+
+xarray's lazy loading means each year's slice only pulls the data it
+actually needs from disk, so this stays memory-efficient even for very
+long forcing records.
 
 
 Multi-Device Parallelization
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -205,22 +205,63 @@ each "month" is a fixed 365/12-day chunk, not aligned to calendar month
 boundaries — so this is mostly an ergonomic shortcut.
 
 For *calendar-aligned* monthly / annual statistics, run the model at a
-daily ``save_interval`` and post-resample the trajectory:
+daily ``save_interval`` and post-resample the trajectory using xarray's
+standard ``resample`` API. The trajectory's ``time`` coord is real
+``datetime64``, so xarray's resampler does the calendar bookkeeping:
 
 .. code-block:: python
 
    predictions = model.run(save_interval='1 day', total_time='1 year')
+   ds = predictions.to_xarray()
 
-   # Calendar-aligned monthly means against the trajectory's real
-   # `datetime64` time coord. Returns an xarray.Dataset.
-   monthly = predictions.resample('1MS').mean()
+   # Calendar-aligned monthly means.
+   monthly = ds.resample(time='1MS').mean()
 
    # Daily total precipitation summed into calendar months, etc.
-   monthly_precip = predictions.resample('1MS')['precipitation'].sum()
+   monthly_precip = ds['precipitation'].resample(time='1MS').sum()
 
-This works because the trajectory carries actual ``datetime64`` timestamps,
-so xarray's resampler does the calendar bookkeeping. The cost is keeping
-daily output in memory for the duration of the run.
+The cost of this pattern is keeping daily output in memory for the
+duration of the run.
+
+Long forcing time-series and chunked runs
+-----------------------------------------
+
+Time-varying forcing is captured at JIT compile time as a closure constant
+(see ``Model._get_step_fn_factory``). Multi-year hourly forcing is therefore
+tractable for runs of a few simulated years; longer or finer-cadence
+forcing should either be resampled before loading, or fed to the model in
+chunks via ``Model.resume``. The chunked pattern avoids inflating the JIT
+closure and lets you stream output to disk year by year:
+
+.. code-block:: python
+
+   import xarray as xr
+   from jcm.forcing import ForcingData
+
+   # Multi-decade ERA5-style forcing with real timestamps.
+   ds = xr.open_dataset('era5_1980_2010.nc')
+
+   # Run one calendar year at a time.
+   for year in range(1980, 2011):
+       year_slice = ds.sel(time=str(year))
+       year_slice.to_netcdf(f"_year_{year}.nc")
+       forcing = ForcingData.from_file(f"_year_{year}.nc", coords=coords)
+
+       if year == 1980:
+           preds = model.run(forcing=forcing, save_interval='1 day',
+                             total_time='1 year')
+       else:
+           preds = model.resume(forcing=forcing, save_interval='1 day',
+                                total_time='1 year')
+
+       preds.to_xarray().to_netcdf(f"output_{year}.nc")
+
+Each ``run``/``resume`` call rebuilds the JIT-compiled step function for
+the new forcing closure (so the first iteration is slow as JAX compiles,
+subsequent iterations reuse the same compilation cache key when the
+forcing's pytree structure is unchanged). For really long campaigns,
+prefer this chunked pattern over loading the entire forcing record into
+one ``ForcingData``.
 
 
 Multi-Device Parallelization
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -107,11 +107,17 @@ For a more realistic simulation with orography and time-varying boundary conditi
    terrain_file = data_dir / "terrain.nc"
    terrain = TerrainData.from_file(terrain_file, coords=coords)
 
-   # Load realistic forcing data (SST, sea ice, soil moisture, etc.) interpolated to T31 grid
+   # Load realistic forcing data (SST, sea ice, soil moisture, etc.) interpolated to T31 grid.
+   # Time-varying variables are wrapped as `TimeSeries` leaves; the Model
+   # picks the right slice each step via `forcing.select(date)`. By default
+   # `from_file` auto-detects climatology vs date-aligned mode from the
+   # netCDF time axis (one-year files wrap, multi-year files align by date).
    forcing_file = data_dir / "forcing.nc"
    forcing = ForcingData.from_file(forcing_file, coords=coords)
 
-   # Create model with realistic configuration
+   # Create model with realistic configuration. SPEEDY assumes a 365-day
+   # no-leap calendar by construction; pass `calendar='gregorian'` if you
+   # want the model clock to advance against real Gregorian timestamps.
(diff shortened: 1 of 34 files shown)


jcm/physics/speedy/params.py
```diff
@@ -28,22 +28,12 @@ def default(cls):
...
-        return cls(
-            increase_co2 = False,
-            co2_year_ref = 1950,
-        )
-
-    def isnan(self):
-        self.increase_co2 = 0
-        self.co2_year_ref = 0
-        return tree_util.tree_map(jnp.isnan, self)
+# ForcingParameters used to carry `increase_co2` and `co2_year_ref` for the
```
duncanwp: Do NOT reference previous code / patterns - it leads to confusion

jcm/forcing.py
```diff
@@ -1,29 +1,117 @@
...
+    `values` carries the data with a time axis at index 0; `time_seconds`
+    is a 1-D coordinate (seconds since `MODEL_EPOCH`) used by `BY_DATE`
+    indexing. `align_mode` is `WRAP_YEAR` or `BY_DATE`. The Model collapses
+    every `TimeSeries` leaf to its current-step slice via
+    `ForcingData.select(date)` before handing the forcing to physics, so
+    physics terms always see the leading-time axis already removed.
+    """
+
+    values: jnp.ndarray
+    time_seconds: jnp.ndarray
```
duncanwp, line 44: I wonder if this might neatly be done using coordax: https://github.com/neuralgcm/coordax. We already have an issue open to explore using that library here: https://github.com/climate-analytics-lab/jax-gcm/issues/214

jcm/forcing.py
```diff
@@ -55,24 +147,36 @@ def ones(cls,nodal_shape,
...
             aerosol_year_weight=aerosol_year_weight if aerosol_year_weight is not None else jnp.ones(nplumes),
             aerosol_ann_cycle=aerosol_ann_cycle if aerosol_ann_cycle is not None else jnp.ones(nplumes),
+            solar=solar if solar is not None else SolarGeometry.zero(),
         )
-    
+
     @classmethod
-    def from_file(cls, filename: str, coords: CoordinateSystem = None):
-        """Initialize forcing data from a file.
+    def from_file(cls, filename: str, coords: CoordinateSystem = None,
```
duncanwp, line 156: This is neat, but might also be easier using the coordax<~>xarray functionality

jcm/forcing.py
```diff
@@ -92,61 +196,212 @@ def from_file(cls, filename: str, coords: CoordinateSystem = None):
...
         )
 
     def isnan(self):
         return tree_util.tree_map(jnp.isnan, self)
 
     def any_true(self):
         return tree_util.tree_reduce(lambda x, y: x or y, tree_util.tree_map(jnp.any, self))
-    
+
+    def select(self, date: DateData, calendar: str = DEFAULT_CALENDAR) -> "ForcingData":
```
duncanwp, line 288: Yeah OK, I think a lot of the below might already be dealt with by coordax

jcm/model.py
```diff
@@ -85,6 +85,22 @@ def physics(self):
     def times(self):
         return self._predictions.times
 
+    def resample(self, freq):
```
duncanwp: Maybe this is just better in the documentation / readme as a useful pattern? It's a lot of lines of code for wrapping a simple function call

jcm/model.py
```diff
@@ -276,12 +294,27 @@ def __init__(self, coords: CoordinateSystem, time_step=30.0, terrain: TerrainDat
...
+                Calendar used for `tyear` / `model_year` / forcing-axis alignment.
+                ``"365_day"`` (no leap years) matches SPEEDY's climatology
+                tables and solar lookup; ``"gregorian"`` (365.2425 days/year)
+                is available for ICON-style runs that align against real
+                Gregorian timestamps in their forcing files.
             log_level:
                 (int) indicates what level of messages will be output, use logging.INFO (20) for verbose (defaults logging.CRITICAL)
 
+        Notes:
+            Time-varying forcing is captured at JIT compile time as a closure
```
duncanwp: Again, let's make this a documentation thing rather than a code comment. Include an example of how to simply loop over a long forcing time-series running e.g. 1 year of simulation at a time

duncanwp (commented):
This looks good and I think it's a neat solution. I've left some inline coments, and the only other major comment is that we should be able to remove the forcing slicing logic from the speedy physics here now: https://github.com/climate-analytics-lab/jax-gcm/blob/dc08a46b6e00ff8b74167a88d4d5e3a9e93b8b57/jcm/physics/speedy/speedy_terms.py#L212

I haven't looked thoroughly but I suspect we can also simplify the ICON physics. It looks like it's physics data still has a date object which it should no longer need. Please do a check of other possible redundencies

duncanwp:
Thanks for the review! Pushed 3e2f3f9 with everything addressed:

- **`physics/speedy/params.py:31`** — migration comment block deleted; the deletion stands without a pointer to the old code.
- **`forcing.py:45 / 157 / 281` (coordax)** — agreed it's the right shape for that library, but the migration is wider than this PR. Captured as #450, tied to #214.
- **`model.py:88` (`ModelPredictions.resample`)** — removed. The pattern (`predictions.to_xarray().resample(time='1MS').mean()`) is now documented in `getting_started.rst` instead.
- **`model.py:307` (JIT-closure `Notes:` block)** — moved out of the docstring into a new "Long forcing time-series and chunked runs" section in `getting_started.rst`, with an example loop running one calendar year at a time via `Model.resume`.
- **PR-level: speedy slicing logic at `speedy_terms.py:212`** — confirmed already gone in the current head; the link in the review pointed at the pre-PR base (`dc08a46`).
- **PR-level: ICON `physics_data` still has a `date` object** — good catch. After this PR's earlier commits the only fields physics ever reads off it were `model_step` and `dt_seconds`, neither of which are really 'date' info. Replaced `PhysicsData.date: DateData` with first-class `model_step` / `dt_seconds` scalar fields on both SPEEDY and ICON `PhysicsData`. Wall-clock info remains exclusively on `forcing.solar`.

770 unit tests pass; lint clean.

docs/source/getting_started.rst
```diff
@@ -187,6 +193,77 @@ You can customize various aspects of the model:
...
+   # Daily total precipitation summed into calendar months, etc.
+   monthly_precip = ds['precipitation'].resample(time='1MS').sum()
+
+The cost of this pattern is keeping daily output in memory for the
+duration of the run.
+
+Long forcing time-series and chunked runs
+-----------------------------------------
+
+Time-varying forcing is captured at JIT compile time as a closure constant
```
duncanwp: Make this explanation a bit simpler, this is aimed at general users who don't need to know the technical details of JAX

docs/source/getting_started.rst
```diff
@@ -187,6 +193,77 @@ You can customize various aspects of the model:
...
+.. code-block:: python
+
+   import xarray as xr
+   from jcm.forcing import ForcingData
+
+   # Multi-decade ERA5-style forcing with real timestamps.
+   ds = xr.open_dataset('era5_1980_2010.nc')
+
+   # Run one calendar year at a time.
+   for year in range(1980, 2011):
```
duncanwp: I think this would be neater using `groupby` year?

docs/source/getting_started.rst
```diff
@@ -187,6 +193,77 @@ You can customize various aspects of the model:
...
+       year_slice = ds.sel(time=str(year))
+       year_slice.to_netcdf(f"_year_{year}.nc")
+       forcing = ForcingData.from_file(f"_year_{year}.nc", coords=coords)
```
duncanwp: Why are you round-tripping to disk? We can just do it in memory can't we? Xarray's lazy loading takes care of memory overheads

docs/source/getting_started.rst
```diff
@@ -187,6 +193,77 @@ You can customize various aspects of the model:
...
+
+       if year == 1980:
```
duncanwp: Just do the first year outside the loop?

docs/source/getting_started.rst
```diff
@@ -187,6 +193,77 @@ You can customize various aspects of the model:
...
+           preds = model.run(forcing=forcing, save_interval='1 day',
+                             total_time='1 year')
+       else:
+           preds = model.resume(forcing=forcing, save_interval='1 day',
+                                total_time='1 year')
+
+       preds.to_xarray().to_netcdf(f"output_{year}.nc")
+
+Each ``run``/``resume`` call rebuilds the JIT-compiled step function for
```
duncanwp: resume doesn't compile again I don't think?

duncanwp: Again, make this easier for non-experts to understand

docs/source/speedy_physics.rst
```diff
@@ -424,24 +424,25 @@ Forcing and Boundary Conditions
...
-     - Default
-   * - ``increase_co2``
-     - Enable time-varying CO₂
-     - True
-   * - ``co2_year_ref``
-     - Reference year for CO₂
-     - 1950
+- CO₂ mixing ratio
+
+All time-varying boundary conditions live on :py:class:`jcm.forcing.ForcingData`.
```
duncanwp: Again, this is too technical for user documentation. The details are in the code comments, the docs should be understandable by a grad student who knows some python

duncanwp:
Round 2 addressed in f6040cd:

- **`getting_started.rst:229`** — dropped the JIT-internals aside; the chunked-forcing intro is now plain user-facing prose.
- **`getting_started.rst:245` (use `groupby` year)** — done. Loop is now `for year, year_ds in ds.groupby('time.year'):`.
- **`getting_started.rst:248` (no disk round-trip)** — done. Added `ForcingData.from_dataset(ds, coords=...)` so the in-memory `year_ds` can go straight into the model. `from_file` is now a one-line wrapper around `from_dataset`.
- **`getting_started.rst:250` (first year outside the loop)** — done. `next(year_iter)` for year one, `for ... in year_iter` for the rest. No `if year == start_year`.
- **`getting_started.rst:259` (`resume` doesn't recompile)** — you're right, dropped that claim.
- **`getting_started.rst:259` (non-expert)** — rewrote the section using simple words; mentions xarray lazy loading without going into JAX internals.
- **`speedy_physics.rst:429`** — cut the implementation-level walkthrough (`TimeSeries`, `forcing.solar`, pre-slicing). The user docs now just say what `forcing.co2_vmr` is and what the default is. Implementation details stay in the code comments.

Tests pass, lint clean.

duncanwp:
Folded #352 and #410 into fa83593:

- **#352 (DateData design)** — `dt` is now the only stored time field. `tyear` / `model_year` are methods that derive from `self.dt` on demand. Override kwargs removed from `zeros`/`ones`/`copy`/`set_date`, so there's nothing that can drift out of sync. Method form lets the caller pass `calendar` per call (the model passes its own).
- **#410 (accurate solar timing)** — under `'gregorian'`, `fraction_of_year_elapsed` now returns *true* day-of-year / actual-year-length, with leap years honoured. Uses the Fliegel & Van Flandern (1968) integer Y/M/D extraction (`gregorian_ymd_from_days`) so it stays JAX-traceable. The smeared 365.2425 approximation is gone. `'365_day'` calendar unchanged — every year is exactly 365 days by design.

`_solar_from_date` and `_wrap_year_index` both call `date.tyear(calendar)`, so the orbital phase now tracks the actual day-of-year and the climatology wrap indexing is leap-aware too.

PR description updated to list #352 and #410 in the closed-issues line. 772 tests pass; lint clean.

PR #458, opened 2026-05-04
ECHAM physics over real terrain: surface + moisture-balance fixes

duncanwp:
## Summary

Take the T63L47 hybrid + real terrain + balanced_isothermal + ECHAM physics + sponge run from **NaN at step 34** to **stable for 30+ days at dt=12 min** on GPU. Five fixes (one bug + four supporting), all isolated by GPU bisection on the actual production CLI config.

## The fixes (in dependency order)

| Commit | Fix | Effect |
|---|---|---|
| `0070ee4` | Sign flip in `surface/echam/{land,sea_ice}.py`: `delta = surf - atm` (matches `ocean.py` and `apply_surface`'s positive-upward convention) | NaN @34 → @603 |
| `00c94c5` | Implicit-Euler damping `1/(1+K·dt/dz)` on bottom-level surface tendencies; read `ocean_temp`/`land_temp` directly from forcing (not the upstream-blended `surface_temperature`); `ComposableEchamPhysics.__add__` overrides parent so `+UpperSponge` preserves type | @603 → @773 (no sponge), past 21 days |
| `571489b` | Cap convective T tendency at ±5 K/hr in `apply_convection` — bounds the residual high-orography column runaway (TN scheme + 2× supersat at L40 over Tibet) | @1796 → 30+ days |
| `a352892` | **The moisture-balance bug**: `filter_tendencies` was only hyperdiffusing `divergence`. Vorticity, T', log_ps and tracers (including `q`) passed through unfiltered. Now route each prognostic through its proper `DiffusionFilter` setting | worst negative-q from `-1e-3 kg/kg` → `-4e-6 kg/kg` (8 orders of magnitude) |

## Tests (10 GPU-gated regressions, 16m32s on one H100)

`echam_t63_land_repro_test.py` (6 tests) — production CLI config:
- ``test_aquaplanet_t63l47_baseline`` (control)
- ``test_real_terrain_stable_for_24h`` (originally NaN @34)
- ``test_real_orog_zero_fmask_does_not_nan`` (orog × surface bisection: confirms it isn't land-tile fluxes)
- ``test_real_terrain_with_sponge_stable_5_days`` (catches the ``__add__`` regression — without it NaNs @95)
- ``test_real_terrain_with_sponge_stable_30_days`` (full production wiring, 30 sim days)
- ``test_real_terrain_minus_surface_survives`` (kept as bisection control)

`echam_column_moisture_test.py` (4 tests) — single-column moisture invariants on the Tibetan column:
- ``test_no_surface_means_no_q`` (sanity)
- ``test_q_stays_subsaturated`` (Sundqvist must condense supersat)
- ``test_q_negatives_stay_below_1_percent_of_q_max`` (catches return of the pre-fix g/kg-scale negatives)
- ``test_full_physics_q_grows_smoothly`` (catches the original sign-flip's >5 g/kg jumps)

`echam_with_land_test.py` — small T31L47 hybrid bisection control on CPU.

All gated behind ``JCM_RUN_GPU_INTEGRATION_TESTS=1``.

## Two known follow-ups (documented in test docstrings)

1. **Convection cap is a workaround.** ECHAM has the same problem and bounds it via the mass-flux limiter ``zmfub1 = (CAPE × zmfub) / (zheat × cmftau)`` in ``mo_cumastr.f90:580``. Porting that limiter would let us drop the cap.
2. **µg/kg-scale negative q at noise floor.** Even with proper hyperdiffusion the spectral round-trip of horizontally-advected `q` from neighbouring evap-active grid cells leaves O(1 µg/kg) negatives — physically zero but not bit-exact. Eliminating them entirely needs positive-definite tracer advection in dinosaur.

## Test plan
- [x] Lint clean (``ruff check``)
- [x] GPU sweep: 10/10 in ``echam_t63_land_repro_test.py`` + ``echam_column_moisture_test.py`` (16m32s)
- [x] CLI: ``CUDA_VISIBLE_DEVICES=4 python -m jcm.main physics=echam grid=echam_t63_l47_hybrid terrain=from_file forcing=from_file init=balanced_isothermal run=longrun run.total_time=30`` completes 3 chunks, no NaN, sensible climate (T 226-331 K, mean 269 K)
- [ ] Run a >30-day climatology and check the precip / temperature spinup looks sane


docs/source/echam_physics.rst
```diff
@@ -420,6 +466,15 @@ Surface Physics
      - Number of soil layers
      - 4
 
+**Boundary-condition workarounds**
```
duncanwp: I think this documentation is out of date now that we have proper BCs. Discuss the fields that are read from the BC files and the utils available for creating them from the original ECHAM files. 

I think we removed these workarounds in a previous commit, but they really do still exist in the code we should remove them. Maybe the ECHAM physics should check it actually has sensible values in the BCs it gets though?

PR #462, opened 2026-05-07
refactor(physics): scheme-named PhysicsTerm migration (Phases 0-3)

duncanwp:
## Summary

Phases 0–3 of the scheme-named-terms refactor planned in `docs/design/composable_physics.md`. Each ECHAM physics scheme that was previously a model-named `Echam<X>` term wrapping an `apply_<x>` function inside `echam_physics.py` is now a self-contained `PhysicsTerm` living next to its underlying numerical implementation.

- **Phase 0** — Bit-exact reference trajectory tests (T31L8 ECHAM + T21L8 SPEEDY 1-day). Frozen `.npz` references in `jcm/data/test/composable_physics_regression/`. Acts as the safety net for every subsequent phase.
- **Phase 1** — Promote `_prepare_common_physics_state` and `apply_forcing_data` to real `PhysicsTerm`s (`MoistAirColumnState`, `EchamBoundaryConditions`). Moist-air diagnostics flip to public top-level keys (`pressure_full`, `air_density`, …).
- **Phase 2** — `TiedtkeConvection` template migration. `ConvectionData` moves to live with the scheme.
- **Phase 3** — Migrate the remaining 13 schemes: `SimpleGwd`, `HinesGwd`, `LottMillerSso`, `SimpleChemistry`, `Macv2SpAerosol`, `SundqvistCloudFraction`, `Echam1MMicrophysics`, `GreyTwoStreamRadiation`, `RRTMGPRadiation`, `NNEmulatorRadiation`, `TteTkeVerticalDiffusion`, `EchamSurface`, `Lohmann2MMicrophysics`. Each scheme owns its own `PhysicsTerm` class, holds its native `Parameters` as `nnx.Param`, reads/writes the diagnostics dict via public keys (`"radiation"`, `"clouds"`, `"vertical_diffusion"`, `"surface"`, `"aerosol"`, `"chemistry"`), and lives in the same file as the numerical implementation. The radiation triplet shares `radiation_should_compute` / `cached_radiation_tendency` helpers extracted to `grey_two_stream/radiation_scheme.py`.
- The `apply_*` layer in `echam_physics.py` shrinks from ~1647 lines to ~140 lines (just the still-used `_prepare_common_physics_state` plus migration markers). The `EchamTermBase` / `_data_from_diagnostics` / `_diagnostics_from_data` translation helpers remain as a bridge during Phase 4 cleanup.
- `IMPLEMENTATION_ROADMAP.md` repurposed as a living housekeeping checklist for the remaining cleanup (docs, dead test rewrites, `ComposableEchamPhysics` + isinstance gates, deletion of dead files).

## What's intentionally NOT in this PR

- Phase 4 (drop `ComposableEchamPhysics`, `apply_timestep`, three isinstance gates in `model.py` / `single_column_model.py` / `prescribed_state_model.py`).
- Phase 5 (delete the now-empty `echam_physics.py` / `echam_physics_data.py` / `parameters.py` / `forcing.py`).
- Phase 6 (plugin-author docs).
- SPEEDY scheme function signatures stay unchanged.

## Surfaced bugs / oddities

The `requires`/`provides` validator caught a real ordering bug while migrating the radiation triplet: the default ECHAM factory runs radiation **before** cloud fraction, so radiation reads the previous step's `cloud_fraction` (or zeros on step 1). This is the legacy behaviour preserved via `prev_physics_data`, so the bit-exact regression passes — but it's almost certainly not what we want physically. Worked around by dropping `"clouds"` from the radiation terms' `requires` and noting the zero-fallback for step 1; tracked in `IMPLEMENTATION_ROADMAP.md` for follow-up. Other oddities listed there too.

## Test plan

- [x] `JAX_PLATFORMS=cpu pytest -n 12 -m "not slow"` → 822 passed, 22 skipped, 0 failed
- [x] Bit-exact T31L8 ECHAM 1-day regression
- [x] Bit-exact T21L8 SPEEDY 1-day regression
- [x] `test_echam_2m_composable_model_run` (slow) passes
- [x] `ruff check` clean across all touched files
- [ ] (Reviewer) Spot-check one or two scheme migrations against the `apply_*` they replaced
- [ ] (Reviewer) Decide whether the cloud_fraction / radiation ordering inconsistency should be fixed in this PR or a follow-up


jcm/data/test/composable_physics_regression/echam_t31l8_1day.npz
duncanwp, line 1: We already have a harness for this sort of test: generate_default_stats.py. Please use / extend that script to generate appropriate (5 or 30 day) test statistics for ECHAM physics, and also include 1 day snapshots to be able to test closer reproducibility. Bit reproducibility is not possible given the different hardware we run and test on

docs/design/composable_physics.md
```diff
@@ -579,3 +579,146 @@ output (written to xarray); keys prefixed with `_` (e.g. `_radiation`,
 `_convection`, `_chemistry`) are internal inter-term state and are
 filtered out of the user-facing output by `data_struct_to_dict`.
 
+---
+
+## Planned next phase: scheme-named terms (in progress)
```
duncanwp: Rewrite these documentation additions in the present tense assuming the new infrastructure is done, otherwise we'll have to rewrite tomorrow. Don't refer to how it was - just describe our new architecture and the flexibility it enables (including an example)

docs/design/composable_physics.md
```diff
@@ -579,3 +579,146 @@ output (written to xarray); keys prefixed with `_` (e.g. `_radiation`,
...
+
+physics = echam_physics().replace("convection", MyScheme())
+```
+
+`ComposablePhysics.__init__` validates the new term list at construction
+time: every `requires` has an upstream `provides`; no key is provided
+twice; declared `required_tracers()` are seeded into the initial state
+by `Model`.
+
+### Migration phases
```
duncanwp: Drop this section

jcm/physics/chemistry/simple_chemistry.py
```diff
@@ -283,12 +283,89 @@ def initialize_chemistry_tracers(
...
-    )
\ No newline at end of file
+    )
+
+
+# ---------------------------------------------------------------------------
+# Composable physics term wrapper
+# ---------------------------------------------------------------------------
+
+from typing import ClassVar  # noqa: E402
```
duncanwp, line 366: This is OK for now, but when you tidy up the old code, this should move to the top. Don't work around the linter

jcm/physics/clouds/lohmann_2m.py
```diff
@@ -3579,3 +3579,154 @@ def _flux_coupled_step(carry, level_in):
...
+
+    Must be composed downstream of ``SundqvistCloudFraction`` and
+    (because it reads TKE) downstream of ``TteTkeVerticalDiffusion``.
+    """
+
+    name: ClassVar[str] = "lohmann_2m_microphysics"
+    category: ClassVar[str] = "clouds"
+    requires: ClassVar[tuple[str, ...]] = (
+        "pressure_full", "air_density", "layer_thickness",
+        "clouds", "aerosol",
```
duncanwp, line 3628: Shouldn't this also depend on vertical_diffusion? Check for other missing requires

jcm/physics/radiation/grey_two_stream/radiation_scheme.py
```diff
@@ -479,5 +479,258 @@ def radiation_scheme(
...
+from jcm.physics_interface import PhysicsState, PhysicsTendency  # noqa: E402
+from jcm.terrain import TerrainData  # noqa: E402
+
+
+def _column_vector(value: jnp.ndarray, ncols: int) -> jnp.ndarray:
+    """Return a vmapped scalar diagnostic as one value per column."""
+    return jnp.reshape(value, (ncols,))
+
+
+def radiation_should_compute(
```
duncanwp: Maybe this should live up in /radiation.__init__.py instead then?

duncanwp (commented):
Looking good

duncanwp:
Pushed review fixes in [f48651c](https://github.com/climate-analytics-lab/jax-gcm/pull/462/commits/f48651c). Per-comment:

- **`generate_default_stats.py` harness** — done. Added `jcm/data/test/echam/generate_default_stats.py` mirroring the SPEEDY harness. New slow test `test_echam_model_default_statistics` in `model_test.py` runs default `echam_physics()` for 5 days and asserts each variable's global mean falls inside the saved `mean ± 3·std` climatology band (`jcm/data/test/echam/default_statistics.nc`). 5 days because longer aquaplanet runs still hit the T31L8 NaN you're debugging on `debug/echam-2m-micro-stability`. The 1-day `.npz` snapshot test stays as a tighter complementary check, but switched from bit-exact to `np.testing.assert_allclose(rtol=1e-3, atol=1e-6)` to tolerate cross-hardware XLA drift.
- **Doc rewrite (line 584)** — done. Section retitled "Scheme-named terms", rewritten in present tense, drops all "today every parameterisation lives behind three layers" framing.
- **Drop migration-phases section (line 704)** — done.
- **`simple_chemistry.py:301` import-at-top** — deferred to Phase 4 cleanup as you suggested.
- **Lohmann2M missing requires + audit (line 3624)** — done. `Echam1MMicrophysics` now requires `aerosol`; the three radiation terms now require `surface`. `Lohmann2MMicrophysics` *keeps* `vertical_diffusion` out of `requires` with an explicit comment — the vdiff term runs after microphysics in the default factory order, so the TKE comes from the previous step's `prev_physics_data`.
- **Move `radiation_should_compute` / `cached_radiation_tendency` to `radiation/__init__.py`** — done.

duncanwp:
Pushed [3c1b073](https://github.com/climate-analytics-lab/jax-gcm/pull/462/commits/3c1b073). The ECHAM stats test now exercises the production wiring you wanted:

* T63L47 hybrid coords + real ECHAM terrain + forcing under \`jcm/data/bc/t63\`
* \`echam_physics(grey) + UpperSponge\` (the same composition the \`debug/echam-2m-micro-stability\` work runs against)
* Resumes from a saved 5-day spun-up state in \`jcm/data/test/echam_t63l47/spinup_state.nc\` (2 MB, prognostic state only, netCDF deflate-compressed) — so each test invocation only pays the 5-day post-spinup cost, not a fresh cold start
* New harness in \`jcm/data/test/echam_t63l47/generate_default_stats.py\` mirrors \`t30/generate_default_stats.py\`. \`generate()\` regenerates both the spun-up state and the climatology stats together when the physics changes intentionally.
* Test gated on \`JCM_RUN_GPU_INTEGRATION_TESTS=1\` — T63L47 is too heavy for CPU CI. Verified locally on GPU 6 in ~2 minutes wall-clock.

One workaround flagged in the commit message: the assertion uses daily *snapshots* (mean over 5 days) rather than \`output_averages=True\` daily means. \`output_averages=True\` on hybrid coords trips a pre-existing shape-broadcast bug in \`compute_diagnostic_state_hybrid\` (existing T63L47 tests don't exercise that path). The bug is unrelated to the refactor; happy to file a separate issue if you want it tracked.

PR #464, opened 2026-05-08
Make Docker image accept Hydra CLI args (#376)

### LINKED ISSUE #376: Make Docker image accept command arguments for non-interactive simulations

duncanwp:
## Summary

- Replaces the Dockerfile `CMD ["bash"]` with `ENTRYPOINT ["python", "-m", "jcm.main"]` (and an empty `CMD []`) so that anything appended to `docker run <image>` is forwarded to the Hydra CLI. This enables non-interactive simulations on Kubernetes, GCP, NRP and similar batch/queued clusters, and unlocks parameter sweeps via Hydra `-m` multirun (relevant once #173 / #375 land).
- Adds a "Running in Docker" section to `README.md` covering the default run, Hydra overrides (config groups + dotted overrides), multirun, persisting outputs via volume mounts, dropping into an interactive shell, and a Kubernetes pod example.

## Context

Picks up #378, which was abandoned with no diff in its final state. The CLI has since been restructured (Hydra config groups under `jcm/config/{physics,grid,run,...}`), so the README examples here use the current `physics=icon` / `run.total_time=30` syntax rather than the `+time_step=20` shape from the original PR draft.

Closes #376.

## Test plan

- [ ] `docker build -t jcm .` succeeds
- [ ] `docker run --rm jcm` runs the default 10-day SPEEDY aquaplanet config
- [ ] `docker run --rm jcm physics=held_suarez grid=held_suarez_t31_l8 run.total_time=2` honours Hydra overrides
- [ ] `docker run --rm -v "$(pwd)/outputs:/app/outputs" jcm run.total_time=1` writes the output netCDF to the host outputs/ directory
- [ ] `docker run --rm -it --entrypoint bash jcm` drops into a shell

Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,54 @@
-FROM python:3.11-slim
+# GPU build of the JCM image. Uses an NVIDIA CUDA runtime base and JAX's
+# bundled CUDA 12 wheels. JAX falls back to CPU if no GPU is visible at
+# runtime, so the same image works on hosts without `--gpus all`, just
+# without acceleration.
+#
+# Build:
+#     docker build -t jcm .
+#
+# Run on a GPU host:
+#     docker run --rm --gpus all jcm physics=icon run.total_time=2
+FROM nvidia/cuda:12.6.3-cudnn-runtime-ubuntu22.04
 
-# Install dependencies
-RUN apt-get update && apt-get install -y  \
-    git \
-    curl \
-    build-essential \
-    libopenblas-dev \
-    liblapack-dev \
-    libffi-dev \
-    libglib2.0-0 \
-    libsm6 \
-    libxext6 \
-    libxrender-dev \
+# Ubuntu 22.04 ships Python 3.10; pull 3.11 from deadsnakes to satisfy
+# JCM's >=3.11 requirement. DEBIAN_FRONTEND=noninteractive prevents tzdata
+# (a transitive dep of build-essential) from prompting at build time.
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        software-properties-common \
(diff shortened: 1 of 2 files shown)


duncanwp:
Smoke-tested locally on macOS arm64 (no GPU, JAX falls back to CPU as expected).

- `docker build -t jcm .` — succeeds (~5 min, image ~6 GB / 16.5 GB on disk due to bundled CUDA libs)
- `docker run --rm --entrypoint bash jcm -c 'python --version'` → `Python 3.11.15 in /app`
- `docker run --rm jcm` (default 10-day SPEEDY aquaplanet) — completes, writes `outputs/2026-05-08/19-47-26/model_state.nc`
- `docker run --rm -v /tmp/jcm-test-outputs:/app/outputs jcm physics=held_suarez grid=held_suarez_t31_l8 run.total_time=2 run.save_interval=1` — Hydra overrides applied, 1.5 MB netCDF lands on the host

JAX picks up `CpuDevice(id=0)` on this Mac and warns about CUDA init failing (`Unknown CUDA error 303` — i.e. no device), which is the expected fallback path. Real GPU validation will happen once the NRP cluster access is sorted.

Issue #471, opened 2026-05-11
Operator-split physics from the IMEX-RK dynamical core


duncanwp:
### Context

Today JCM puts physics inside the *explicit* part of the IMEX-RK SIL3 dynamics step (`jcm/model.py::_get_step_fn_factory` builds `physics_forcing_eqn = ExplicitODE.from_functions(_step_tendencies)` and folds it into `_composed_eqn`). SIL3 has 3 explicit-stage evaluations per dynamics timestep (`dinosaur/time_integration.py::imex_rk_sil3`, `a_ex` is 3×3), so `_step_tendencies` — and therefore `physics.compute_tendencies` — is called **3 times per `dt`**, each time on a different intermediate RK state.

Mathematically this is "correct" for SIL3 to retain its 3rd-order accuracy on explicit terms. Operationally it has two costs that are starting to bite:

1. **3× the physics work per timestep.** Most expensive terms (radiation, microphysics, convection, surface, vertical diffusion) re-run on every stage.
   - Radiation is partly mitigated by `radiation_should_compute` sub-cycling that re-emits a cached heating tendency on most stages — so full RRTMGP/grey optics actually fire much less often. But every other term runs full-strength 3×.
2. **The cache plumbing for cross-step physics state is fragile.** With multi-stage RK, "previous state" is ambiguous (previous substage vs previous full step). PR #469 had to add explicit `physical_step` gating to `physics_data_cache` writes so the cache reflects the previous *timestep* and not previous *substage* — an outright correctness fix that wouldn't even need to exist if physics ran once per `dt`. Issue #470 traces the long-duration averaged-mode NaNs back to this same architectural asymmetry between snapshot and averaged paths.

### Proposal

Refactor to **operator-split** physics from the IMEX-RK dynamics step. Conceptually:

```python
def step_with_physics(state, dt):
    physics_tend = physics.compute_tendencies(state)       # ONE evaluation per dt
    state_dyn = imex_rk_dynamics_only(state, dt)           # 3 RK stages, no physics
    state_new = apply_tendency(state_dyn, physics_tend, dt)
    return state_new
```

This is the convention used by ECHAM (the model JCM ports from), CAM, IFS, and basically every operational GCM. The cost is a small splitting error on the explicit-physics part — it's a Lie / Strang splitting between dynamics and physics — which is well-studied and acceptable for climate-rate integrations.

### What this would buy us

[...]

Issue #468, opened 2026-05-11
Grey two-stream: day-14/16 convective-instability NaN in long averaged runs


duncanwp:
## Grey two-stream radiation: convective-instability NaN around day 14-16 in long averaged integrations

When running `physics=echam grid=echam_t63_l47_hybrid` with `output_averages=true`, the integration is stable for ~14 days then **explosively NaN's the entire atmosphere at day 16**. Snapshot mode dodged this for a long time because saved-state-sampling occasionally misses the bad column, but a 30-day averaged run reliably blows up.

### Reproduction

```bash
python -m jcm.main \
  physics=echam grid=echam_t63_l47_hybrid \
  run.total_time=30 run.save_interval=2 run.time_step=12 \
  run.output_averages=true \
  run.sponge.levels=5 \
  init=balanced_isothermal \
  terrain=from_file terrain.file=jcm/data/bc/t63/terrain.nc \
  forcing=from_file forcing.file=jcm/data/bc/t63/forcing.nc
```

Day-by-day check (saves every 2 days, `output_averages=true`):

| Day | T_NaN | T range (K)   | TKE max (m²/s²) | OLR (W/m²) | SW down (W/m²) |
|----:|------:|---------------|-----------------|------------|----------------|
|  2  |  0.0% | 253.4 – 298.4 |   2.7           | 372.5      | 324.6          |
|  4  |  0.0% | 262.6 – 299.0 |   1.8           | 356.9      | 323.8          |
|  6  |  0.0% | 264.0 – 299.4 |   3.8           | 337.8      | 322.9          |
|  8  |  0.0% | 254.9 – 299.4 |   5.4           | 319.8      | 322.0          |
| 10  |  0.0% | 248.1 – 299.4 |  16.5           | 302.7      | 321.0          |
| 12  |  0.0% | 241.5 – 299.5 |  29.0           | 288.5      | 320.0          |
| 14  |  0.0% | 234.8 – 299.6 |  30.5           | 277.7      | 318.9          |
| 16  |**100%**| all NaN      | NaN             | NaN        | 317.7          |
| 18+ | 100%  | all NaN       | NaN             | NaN        | smooth decay   |

[...]

Issue #388, opened 2026-02-27
Have the option to use the pysces dycore


duncanwp:
It would be awesome to have the option of running with a new JAX (or numpy) based spectral finite element dycore called [pysces](https://github.com/OkayHughes/pysces) developed by @OkayHughes. 

It shouldn't be too difficult, but there will be a choice about how to abstract the current `Model` class that assumes using the dinosaur dycore. It might make sense to have a separate Model class for each dycore, or push the dycore specific details into the PhysicsInterface.

duncanwp:
## Implementation plan (for provenance)

Verbatim copy of the plan file approved before implementation began. Lives at `~/.claude/plans/zippy-booping-wind.md` on the dev host; kept here so the chain from issue → plan → PR #489 is self-contained.

<details><summary>Click to expand the plan</summary>

# Pluggable Dynamical Cores — pysces integration (issue #388)

## Context

`jcm.Model` is hard-wired to the `dinosaur` spectral dycore. `_get_dynamics_step_fn`,
`_get_op_split_step_fn`, the three spectral diffusion-filter closures, the global-mean
ps-conservation filter, `_prepare_initial_modal_state`, `_post_process`, the
hybrid-vs-sigma `PrimitiveEquations*` dispatch, and `truncated_modal_orography` all live
inside `Model.__init__` and reach into `dinosaur` symbols directly. `physics_interface.py`
holds the gridpoint↔spectral conversion that wires the only currently-supported dycore
to physics. `nudging.py` is the lone physics-like effect still inside the IMEX-RK
substages via `compose_equations`. Beyond `Model`, modal transforms also leak through
`TerrainData.from_coords`/`from_file` (orography spectral smoothing) and the xarray
output path in `utils._infer_dims_shape_and_coords` (which hard-codes `lon_mode`/
`lat_mode` axes).

Issue #388 asks for a second backend, `pysces` (https://github.com/OkayHughes/pysces) —
a spectral-finite-element cubed-sphere dycore from Okay Hughes (installable from PyPI
as `pyses`; **not** the unrelated systems-biology `pysces` package on PyPI). v2 also
asks that this be a **demonstration of plugging in any dycore**: clean interface, clear
APIs, docs that let an outsider add a third backend without reading `Model.run`.

Approach (confirmed):
- **DynamicalCore protocol**; one user-facing `Model` that delegates to a swappable
  dycore.
[...]

PR #472, opened 2026-05-11
Operator-split physics (#471 phases 0-3): physics outside the RK stage loop

### LINKED ISSUE #470: RRTMGP + output_averages=true NaN over long integrations — integration trajectory should not depend on output mode
### Observation

After the fixes in #469 (mixing-length stability factor, Thomas `_safe`, analytic TKE source step, RRTMGP eager init, IMEX-substage cache gating):

| Configuration | 3 hours | 30 days |
|---|---|---|
| Grey + `output_averages=false` (snapshot) | clean | **clean (full year, see prior runs)** |
| Grey + `output_averages=true` (averaged) | clean | NaN at day 16 → #468 |
| RRTMGP + `output_averages=false` (snapshot) | not tested today, was clean previously | not yet tested today |
| RRTMGP + `output_averages=true` (averaged) | clean (0/97 NaN, T 272-293 K) | **100 % NaN at every saved time-step** |

Reproduction (RRTMGP, 30-day averaged):
```bash
python -m jcm.main \
  physics=echam-rrtmgp grid=echam_t63_l47_hybrid \
  run.total_time=30 run.save_interval=2 run.time_step=12 \
  run.output_averages=true \
  run.sponge.levels=5 \
  init=balanced_isothermal \
  terrain=from_file terrain.file=jcm/data/bc/t63/terrain.nc \
  forcing=from_file forcing.file=jcm/data/bc/t63/forcing.nc
```

The run "completes" (no traceback, `Run completed.` logged at sim time ~29 days), but the saved monthly averages are 100 % NaN in T / u / v / q / TKE / OLR / surface fluxes. Many static-from-forcing fields (surface T, AOD plumes, cos zenith) are correctly preserved — confirming the saving path itself works, the integration produced NaN at some point and the running mean caught it.

### The conceptual problem

**The integration step shouldn't depend on whether we're saving averages or snapshots.** Today it does:

- Snapshot mode (`Model._run_from_state` with `output_averages=False`) uses `dinosaur.time_integration.trajectory_from_step`. The integrator runs the pure dynamical core + physics with no extra side-channel state — `diagnostics_collector=None`, so `prev_physics_data=None` always.
- Averaged mode (`output_averages=True`) uses `averaged_trajectory_from_step`. That scan threads a `DiagnosticsCollector` carry through every step, holding `physics_data_cache` across timesteps. `get_physical_tendencies` then feeds the previous step's physics_data into each term as `prev_physics_data`.

The cross-step cache feedback is the difference. It's not the averaging math: snapshot mode could just as easily compute and discard a running mean. The asymmetry is that averaged mode plumbs extra state through the integration that snapshot mode doesn't, and that extra state (a) destabilises the integration over long times, and (b) means the trajectory under `output_averages=true` is provably not the same trajectory as under `output_averages=false`, even at floating-point precision.

That's wrong on its face. The output mode should only decide *how the saved arrays are aggregated*; the in-memory model trajectory should be identical.

### Suggested architecture

Option A — **decouple the cache from the integration**. Make `prev_physics_data=None` the contract for every term, in both modes. Terms that genuinely need a previous-step quantity (grey/RRTMGP's `cached_radiation_tendency`, TTE-TKE's prev TKE) should manage their own per-step state via a mechanism that's identical in both modes — e.g. by carrying it on `state.tracers` or by recomputing fresh every step. The `DiagnosticsCollector` then has only one job: running-mean accumulation over inner_steps.

Option B — **always run the collector**, snapshot mode included. Then both paths have the same trajectory; `output_averages=true` just additionally writes the accumulated mean to the saved netCDF. Snapshot mode discards the accumulator. This is the smaller architectural change but it changes snapshot-mode behaviour from "no cache" to "cache" — needs careful re-validation.

Either way, **all my evidence so far says the long-duration averaged failures (#468 and the RRTMGP one above) are downstream of this architectural asymmetry**, not bugs in radiation per se. With #469's IMEX-substage cache-write gating in place, the cache is at least consistent across substages within a single step — but it still propagates across steps in averaged mode and not in snapshot mode.

### Why this matters for production runs

We want `output_averages=true` to be the *recommended* path for long climate integrations (it cleanly captures the diurnal cycle that snapshot mode aliases). Today snapshot mode is the only path that completes a year cleanly. That's backwards. Fixing the architectural decoupling will likely also reveal the underlying physics issues that grey-radiation has at day 16 (#468) and whatever RRTMGP-averaged-specific failure shows up at 30 days.

### Related

- #468 — grey day-16 NaN (a likely physics issue, but the asymmetry described here is what made it impossible to even reach day 16 a week ago).
- #469 — the PR that landed the immediate fixes; this issue is the deeper architectural concern that PR doesn't address.
- #463 — the prior partial fix that addressed only the broadcasting crash in `compute_diagnostic_state_hybrid`; the resulting test (now hardened in #469) was the canary that surfaced the rest.

duncanwp:
## Summary

Implements phases 0-3 of the operator-split physics refactor described in #471. Physics is lifted out of the IMEX-RK SIL3 stage loop and called exactly once per `dt` against the start-of-step state; the tendency is applied as a forward-Euler add before a pure-dynamics IMEX-RK step (Lie split (a)). This mirrors ECHAM/CAM/E3SM/IFS at the dycore/physics boundary and closes the structural asymmetry between snapshot and averaged output modes (#470).

The cross-step physics carry (radiation flux for sub-cycling, previous-step TKE for the analytic source update, aerosol AOD, …) flows through the integration scan as a first-class pytree carry — the `DiagnosticsCollector.physics_data_cache` plumbing PR #469 had to gate around is gone.

## Phase-by-phase scope

**Phase 0 — Scaffold the `PhysicsCarryState` API.** New `PhysicsTerm.initial_carry_state(coords)` (default `{}`) + `ComposablePhysics.initial_carry_state` that aggregates per-term slots. No call-site changes.

**Phase 1 — Lie split path behind `use_op_split` flag.** New `Model._get_op_split_step_fn` / `_get_dynamics_step_fn`; physics carry threads through `lax.scan` as an explicit pytree. `compute_physics_step` helper. `physics_tendency_to_dynamics_tendency` produces `log_surface_pressure` tendency with the leading vertical=1 axis the op-split `tree_math` add requires.

**Phase 2 — Default to op-split.** `use_op_split=True` everywhere; legacy path warns at trace-time.

**Phase 3 — `DiagnosticsCollector` cleanup + first-class carry.** `PhysicsCarryState` type alias in `physics_interface`. Drop `physics_data_cache` / `physical_step` from `DiagnosticsCollector`; rename `accumulate_if_physical_step` → `accumulate`. `get_physical_tendencies` no longer reads/writes the cache. Add `initial_carry_state` overrides on `EchamBoundaryConditions`, `TTETKEVerticalDiffusion`, `SundqvistCloudFraction`, `Macv2SpAerosol`. Fix `VerticalDiffusionData.km/.kh` shape mismatch.

Phases 4 (delete legacy path) and 5 (Strang upgrade) are follow-ups; see the updated `docs/design/operator_split_physics.md` for the full plan.

## Why this is the right architecture

- **Closes #470.** Integration trajectory no longer depends on output mode — the carry flows the same way in snapshot and averaged modes.
- **Cheaper.** Physics goes from 3× per `dt` (one per IMEX-RK SIL3 substage) to 1× per `dt`. Radiation sub-cycling now has one timeline to honour, not three.
- **Cleaner.** \`DiagnosticsCollector\` has one job (running-mean accumulation). Substage cache gating is gone. The cross-step physics state has a name and a type.
- **Decoupled from the dycore.** \`Model._get_dynamics_step_fn\` is the only thing #388 (pysces) needs to swap.

Justification, references (Ullrich & Jablonowski 2012; Santos, Caldwell & Bretherton 2021), and the ECHAM-leapfrog caveat (Strang/Lie vs SIL3 substages) are in \`docs/design/operator_split_physics.md\`.

## Test plan

- [x] \`JAX_PLATFORMS=cpu pytest -n 12 -q -m \"not slow\"\` — 855 passed, 13 slow-skipped, 0 failed.
- [x] \`ruff check jcm/ docs/\` — clean.
- [x] 11 new op-split tests in \`model_test.py::TestOperatorSplitPhysics\`:
  - Snapshot + averaged SPEEDY + ECHAM-hybrid finite-state checks.
  - Op-split single-step is \`jit\`-able and the returned carry has stable pytree shape.
  - Op-split global-mean T agrees with the legacy path to < 1 % over a short run.
  - Legacy path emits the deprecation warning.
- [x] 4 new \`initial_carry_state\` tests in \`composable_physics_test.py::TestInitialCarryState\`:
  - Default \`{}\`, aggregation, collision detection, no zero-state probe.
  - ECHAM composition: \`initial_carry_state(coords)\` shapes are a subset of \`get_empty_data(coords)\` and agree where they overlap.
- [x] 5 new \`DiagnosticsCollector\` cleanup tests in \`model_test.py::TestDiagnosticsCollectorCleanup\`:
  - The class no longer has \`physics_data_cache\` or \`physical_step\`.
  - \`accumulate()\` divides by \`steps_to_average\` and writes into \`data[i]\`.
  - \`PhysicsCarryState\` alias is importable.
  - Repo-grep regression: no production code identifier references \`physics_data_cache\`.

## Refs

- Closes architectural concern in #470.
- Implements phases 0-3 of #471.
- Builds on PR #469 (the surgical fixes that surfaced #470/#471).
- Prerequisite for #388 (pluggable dycore — pysces).
- May improve #468 (grey day-16 NaN) once op-split is the default — TBD on long-time stability.


jcm/data/test/t30/regenerate_default_stats.py
duncanwp, line 1: I'm confused. This should just use (or update as needed) the generate_default_stats.py

jcm/physics/aerosol/macv2_sp.py
```diff
@@ -421,6 +421,18 @@ def cache_coords(self, coords) -> None:
         self._lons = nnx.Variable(lon_2d.reshape(-1))
         self._coords_cached = True
 
+    def initial_carry_state(self, coords) -> dict:
```
duncanwp: This should just inherit a generic zeros method on the PhysicsTerm class if it's not doing anything special

duncanwp: I'm also not a fan of the need of coords to be passed through just to determine the size. Can't we get these a neater way?

jcm/physics/echam/echam_physics_test.py
```diff
@@ -120,9 +120,22 @@ def test_echam_physics_integration_3_timesteps(self):
         self.assertGreater(sp_min, 0.3, f"Minimum surface pressure {sp_min} (normalized) seems too low")
         self.assertLess(sp_max, 2.0, f"Maximum surface pressure {sp_max} (normalized) seems too high")
         
-        # Specific humidity should be non-negative
+        # Specific humidity should be near-positive. Small negative
```
duncanwp: Undo this and just add a filter to the list of filters in the Model constructor

jcm/physics/forcing/echam_boundary_conditions.py
```diff
@@ -94,6 +94,24 @@ class EchamBoundaryConditions(PhysicsTerm):
     def __init__(self):
         """No tunables; nothing to initialise."""
 
+    def initial_carry_state(self, coords) -> dict:
+        """Seed the radiation/surface/chemistry sub-structs as zeros.
+
+        The first ``compute_tendencies`` call overwrites the boundary
```
duncanwp: As above

jcm/model.py
```diff
@@ -165,27 +165,188 @@ def _model_predictions_unflatten(aux_data, children):
...
+                # actually consumed — recomputing physics at save time
+                # with a freshly-seeded carry would zero out radiation
+                # on non-radiation outer steps (default 2-hour
+                # ``radiation_interval``).
+                return (
+                    (x_final, ps_final),
+                    post_process_fn(x_final, ps_final),
+                )
+
+            (x_final, ps_final), preds = jax.lax.scan(
```
duncanwp: This is the same as the output_averages path, why not just move this outside the if block and write it once?

jcm/data/test/t30/regenerate_default_stats.py
duncanwp: Merged into the existing `generate_default_stats.py` — `generate()` function added there, separate file removed.

jcm/physics/aerosol/macv2_sp.py
duncanwp: Done — moved to a declarative `carry_slots: ClassVar[dict[str, type]]` class attribute. The base `PhysicsTerm.initial_carry_state` walks it; the three boilerplate overrides here / EchamBoundaryConditions / SundqvistCloudFraction go away. TteTkeVerticalDiffusion keeps its override since it floors TKE at 0.01 m²/s² (non-zero seed).

duncanwp: Coords still flows in because the typed sub-structs (`RadiationData`, `CloudData`, `AerosolData`, `VerticalDiffusionData`, …) fundamentally need grid shape to size themselves — there's no way around that. What's gone is the *per-term* boilerplate; subclasses just declare the slot mapping and the base method handles shape extraction once. Going further (no coords arg) would mean caching `(ncols, nlev)` on the term via `cache_coords` and demanding every `cache_coords` override call `super` — happy to do it if you want, but it touches 8 subclasses for a marginal API win. Let me know.

jcm/physics/echam/echam_physics_test.py
duncanwp: Done a bit differently — see [the commit](https://github.com/climate-analytics-lab/jax-gcm/pull/472/commits/5c9b832) and my reply on the original PR thread. The TL;DR: a spectral round-trip filter on every `dt` would cost two extra spherical-harmonic transforms per non-negative tracer; instead, the post-processors clamp via existing `verify_state` at the modal→nodal output boundary (a single `max`), and the latent oversight where `_post_process` built a clamped state but never assigned it back to `predictions.dynamics` is fixed. Physics input was already clamped by `verify_state` in `compute_physics_step`, so the integration state can carry small Gibbs ringing while output and physics input are non-negative. Test bound restored to `q >= 0`.

jcm/physics/forcing/echam_boundary_conditions.py
duncanwp: Done — same fix as above, replaced by declarative `carry_slots` on the term.

jcm/model.py
duncanwp: Deduplicated. The outer `lax.scan` over `(state, physics_state)` and the `(x_final, ps_final, preds)` return now live once; the snapshot vs averaged difference is isolated to per-mode `outer_step` builders (`_snapshot_outer_step` / `_averaged_outer_step`).

duncanwp:
@codex review

jcm/model.py
```diff
@@ -551,139 +600,223 @@ def _date_from_sim_time(self, sim_time) -> DateData:
...
+        structure on iteration 1 (within-step diagnostic keys terms
+        write are zero-filled). ``get_empty_data`` is internal-only
+        in this role: it discovers the post-step output pytree
+        structure via a one-shot probe — never used as live state.
+        """
+        template = self.physics.get_empty_data(self.coords)
+        initial_carry = self.physics.initial_carry_state(self.coords)
+        if isinstance(initial_carry, dict) and isinstance(template, dict):
+            return {**template, **initial_carry}
+        return initial_carry or template
```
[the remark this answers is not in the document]
duncanwp: Good catch — fixed in [`a4ac742`](https://github.com/climate-analytics-lab/jax-gcm/pull/472/commits/a4ac742). Replaced `return initial_carry or template` with `return template if initial_carry is None else initial_carry` so non-dict carry types (JAX arrays etc.) pass through without triggering ambiguous-truth evaluation.

PR #479, opened 2026-05-12
feat(#128): persistent model-state checkpointing for preemptible runs

### LINKED ISSUE #128: Model state checkpointing (serialization)
We now have a lot of compute time available that is 'preemtable' e.g. might be killed with little notice. Therefore, it would be nice to make it easy to save model states to disk occasionally, and start a simulation from a specified checkpoint file, something like:

```checkpoint_path = get_checkpoint_path(task_name)
if file_exists(checkpoint_path):
    state = restore_state(checkpoint_path)
else:
    state = start_from_scratch()
```

I think we have everything in place to do this, so really we just need a wrapper to make it neat.

duncanwp:
## Summary

Adds save/load primitives so a long chunked integration that gets killed (spot preemption, OOM, manual abort) can be re-launched with the same config and pick up where it left off, instead of restarting.

`Model._final_modal_state` and `_final_physics_state` together carry everything `Model.resume()` needs to keep stepping; this PR just persists them between chunks.

- **`jcm/checkpoint.py`** — `save_checkpoint(model, path, *, elapsed_days)` flattens the two state pytrees to lists of arrays (flax's msgpack codec doesn't natively serialize the `tree_math` structs that `dinosaur` uses for the modal state), msgpacks them alongside the elapsed sim-day count, and writes via tmpfile + atomic rename so a kill mid-write leaves the previous checkpoint intact. `load_checkpoint(model, path) → elapsed_days` deserializes against the destination model's bootstrapped templates and overwrites the live states.
- **`Model.bootstrap_state(initial_state=None)`** lifts the prep block out of `Model.run` into a public method so checkpoint restore (and any other caller wanting the seed pytrees without integrating) doesn't have to spin up a zero-length integration to materialize templates. It also eagerly builds the physics carry now, equivalent in behavior because `resume` skips its lazy branch when the carry is already populated.
- **`run_chunked` wiring** via `cfg.run.checkpoint_path`. When the file exists at startup, the runner bootstraps state, loads the checkpoint, and resumes at the recorded chunk boundary. The loop is now elapsed-day driven rather than `i`-indexed so resume just falls into place. Throughput reporting now uses the post-resume window so the displayed sim-days/hr reflects work actually done on this host. After every chunk (restored or fresh) the runner saves a new checkpoint.

## Usage

```bash
python -m jcm.main \
  physics=echam-rrtmgp grid=echam_t63_l47_hybrid \
  run=longrun run.checkpoint_path=/scratch/job123.ckpt
```

If the run is killed at sim-day 270, a re-launch with the same command picks up at day 270 and runs the remaining chunk(s).

## Test plan
- [x] `pytest jcm/checkpoint_test.py -v` → 4 passed (round-trip fidelity, error paths, resumption equivalence: continuous 4-day Held-Suarez run matches `(2 d, ckpt, fresh model, load, 2 d)` split to < 1e-5)
- [x] `pytest jcm/runners_test.py::TestModeDispatch -v` → 3 passed, including a new test that drives `run_chunked` once with `total_time=1`, then re-invokes with `total_time=2` and the existing checkpoint — the second invocation runs only the remaining chunk
- [x] `pytest jcm/model_test.py jcm/runners_test.py -m "not slow"` → 36 passed, 1 skipped (regression check on the `bootstrap_state` refactor)
- [ ] CI green on the standard fast + coverage gates

Closes #128.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,12 @@ python -m jcm.main physics=echam +physics.terms.tiedtke_convection.params.entrpe
 # Long ECHAM + RRTMGP run with chunked health checks
 python -m jcm.main physics=echam-rrtmgp grid=echam_t63_l47_hybrid run=longrun
 
+# Same, but resumable across preemptions: model state is saved to
+# ``run.checkpoint_path`` after each chunk; if the file already exists at
+# launch, the run picks up at the recorded sim-day.
+python -m jcm.main physics=echam-rrtmgp grid=echam_t63_l47_hybrid run=longrun \
+    run.checkpoint_path=/scratch/$JOB_ID.ckpt
+
 # Single-column physics evolution from a saved JCM run
 python -m jcm.main run.mode=scm run.state_file=path/to/state.nc \
     run.column.lat_deg=0 run.column.lon_deg=180
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -11,3 +11,4 @@ API
    jcm.physics_interface.PhysicsState
    jcm.physics_interface.PhysicsTendency
    jcm.physics_interface.Physics
+   jcm.checkpoint
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -261,6 +261,45 @@ actually needs from disk, so this stays memory-efficient even for very
 long forcing records.
 
 
+Checkpointing for preemptible runs
+----------------------------------
+
+Multi-day integrations on preemptible compute (spot instances, Slurm
+``--requeue`` queues, NRP Nautilus) can be killed at short notice. Set
+``run.checkpoint_path`` to make a chunked run resumable: after each
+chunk the runner persists the modal + physics state and the elapsed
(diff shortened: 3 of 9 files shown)


duncanwp:
Thanks @codex — real P1, fixed in 6ad2fc9.

The resume-from-checkpoint branch in `run_chunked` was calling `inject_jw_profile` / `inject_balanced_isothermal_profile` (which only populate `_final_modal_state`) and then `load_checkpoint` (which guards on both states being non-None as deserialization templates). For `init.kind` of `jw` or `balanced_isothermal`, that raised before doing any actual loading.

Fix: after the inject branch, build the physics carry from `_build_initial_physics_carry` if it's still `None`, so the load has a template to deserialize against. The original regression test used the default `init=isothermal` which routed through `bootstrap_state` (which does eagerly populate both pytrees) so it missed this. Added `test_chunked_resume_with_balanced_isothermal_init` which exercises the inject path end-to-end and would have caught the bug.

PR #487, opened 2026-05-15
feat(clouds): merge condensation into ECHAM column-sweep; restore moisture cycle

duncanwp:
## Summary

- Merge saturation adjustment (Newton step + pass-2 cleanup, ported from Sundqvist) into the ECHAM 1M column-sweep microphysics, matching ECHAM `mo_cloud.f90` structure.
- `Echam1MMicrophysics` switches from the per-level scheme to `cloud_microphysics_column_sweep`. Rain and snow now propagate downward as fluxes within `dt`, Rotstayn rain-evap moistens dry sub-cloud air, and large-scale precip can reach the surface.
- `SundqvistCloudFraction` becomes a pure cloud-fraction diagnostic (zero q/qc/qi tendencies). Condensation lives with the microphysics it feeds.

## Why

Per-term Q-tendency diagnosis on the `analyticoz_full` day-360 spun-up state (see memory `moisture_loop_diagnosis_260515.md`) traced the missing large-scale precip and low LWP to the per-level `cloud_microphysics`:

- Rain/snow reinitialised to zero each level → no inter-level rain flux, no surface large-scale precip, no rain-evap moistening of dry layers.
- Sundqvist evaporated qc/qi back to vapour at +2.1 mm/day, faster than ECHAM 1M could autoconvert → cloud water never accumulated.
- Ice sedimentation removed qi with no flux to the layer below → 0.44 mm/day hidden sink.

PR #458 had added a column-sweep variant but reverted it because Sundqvist's separate pre-step condensation + the column-sweep's rain-evap formed a positive feedback (rain evap → moistens dry layer → next-step Sundqvist condenses → latent heat → more convection → more rain) that crashed the T63L47 30-day test from ~day 30 back to day 7. The planned fix in that PR was *"a coupled implicit Newton step between rain-evap and condensation"*.

This PR realises that fix by putting both inside one column sweep so the rain-evap → re-condensation loop closes within a single `dt` instead of swinging across timesteps.

## Per-layer order (matches ECHAM `mo_cloud.f90` lines 260-1080)

1. Snow melt of incoming `zsfl` at T > tmelt
2. **Pre-microphysics saturation adjustment** — two-pass Newton step (`_saturation_adjustment_layer`) on `(T, q, qc, qi)`
3. Autoconversion / accretion / riming / aggregation on the **post-condensation** `(qc', qi')`
4. Rotstayn rain-evap using **post-condensation** `q'` (existing `0.99·(qs - q')` cap)
5. **Post-rain-evap cleanup** — second Newton pass absorbs any residual oversat in the same `dt`
6. `zrfl` / `zsfl` / `zclcpre` flux update for the next layer

## 2M scheme side-effect

`cloud_microphysics_2m` already does its own condensation internally via `mixed_phase_deposition_and_corrections` (initialises `pcnd = 0`, computed from scratch). Sundqvist's pre-condensation was *double-counting* against the 2M scheme today. Stripping that double-count needs no further code change. A CPU smoke test confirms `Lohmann2MMicrophysics` runs cleanly under the new composition.

## Tests

- **New**: `test_column_water_budget_closes` — locks the mass-conservation invariant `∫(dq+dqc+dqi) dm/dt + precip_surface = 0` to 1e-3 relative tolerance. The per-level scheme could not satisfy this.
- **Updated**: `test_subsaturated_column_evaporates_rain` — putting the cloud at near-saturation with dry layers *below* (the previous setup put the cloud *in* subsaturated air, which is now correctly evaporated by the in-sweep saturation adjustment before autoconv can fire).
- All 24 existing `sundqvist_test.py` tests pass (they exercise the underlying functions, not the term wrapper).
- Full `jcm/physics/clouds` + `jcm/physics/echam` sweep: **137 passed, 1 skipped** (`pytest -m "not slow"`).

## 5-day T63L47 smoke test (GPU 4, dt=12 min, JW-dry init)

|                       | Baseline year-end | Day-5 with merge |
|-----------------------|------------------:|-----------------:|
| Total precip          |   0.54 mm/day     |  **2.54** mm/day |
| Convective precip     |   0.53            |  2.57            |
| Large-scale precip    |   0.005           |  0.00 (spin-up)  |
| LWP + IWP             |   20 g/m²         | **161 g/m²**     |
| CWV                   |   21 kg/m²        |  10 → growing    |

No NaN over 5 days; T 232-296 K. Large-scale precip is still 0 at day 5 because the troposphere is too dry (rain evaporates entirely in the sub-cloud air), but the channel is live and should produce surface precip as the column wets up — a 90-day spin-up run on GPU 2 is in flight now to validate equilibrium-state climatology.

## Test plan

- [ ] 90-day T63L47 ECHAM-RRTMGP integration on GPU 2 — verify no NaN, sensible total precip + LWP at month 3
- [ ] Compare to baseline year-end climatology (LWP, surface precip breakdown, TOA energy)
- [ ] If happy, follow up with a full 365-day run to update production climatology


duncanwp:
@codex review

duncanwp:
2m large scale precipitation looks healthy, but the 1m is still too low. I suspect this might be due to the CDNC we prescribe rather than something structural, and this could/should be tuned. Leaving for future calibration work

PR #489, opened 2026-05-17
feat(dycore): introduce DynamicalCore protocol + wrap dinosaur (#388 phase 0+1)

duncanwp:
## Summary

First slice of the pluggable-dycore work for issue #388. Introduces a `DynamicalCore` protocol, wraps the existing dinosaur spectral path behind it, and thins `Model` to delegate every state-touching operation to `self.dycore`. The dinosaur path is preserved bit-for-bit (within rtol=1e-10) by a captured pre-refactor baseline.

* **Phase 0 — protocol scaffold.** `jcm/dycore/{base.py, registry.py, base_test.py}`: the ABC, the name→factory registry, and a contract test that rejects incomplete implementations.
* **Phase 1 — dinosaur backend behind the protocol.** `jcm/dycore/dinosaur/{dycore.py, state_bridge.py}` own everything spectral; `Model.__init__` thins out; `physics_interface.py` drops the three dinosaur-specific conversion helpers and gains a `compute_physics_step_gridpoint` that the dycore is no longer in the loop for.
* **Layout canary.** `jcm/dycore/_fake_cubed_sphere/` + `jcm/dycore/protocol_test.py`: a deliberately-minimal cubed-sphere fake exercises the protocol on a `(nlev, nelem, gll, gll)` layout to prove nothing in `Model` / `physics_interface` silently assumes lat/lon. This is the canary for Phase 2's real `PyscesDycore`.

## Design choices worth a second look

* **Gridpoint-on-both-sides on the dycore's own horizontal layout** — not lat/lon. Routing through lat/lon would force a forward *and an inverse* tendency regrid every dt for an SE backend, and the inverse is the scientifically sensitive direction (conservation, vertical alignment, surface-pressure tendency, tracer mass). Native-layout dodges this entirely: physics is already column-local under `ComposablePhysics(vectorize_columns=True)`, and the small number of horizontally non-local schemes broadcast against the dycore-supplied `coords.horizontal.latitudes`. See the canary test for the proof.
* **Coordinate-system unification deferred.** v2 reuses `dinosaur.CoordinateSystem` as the common type — physics reads only `latitudes`/`longitudes`/`nodal_shape` off it, and pyses' adapter will stub the modal transforms. A `GridSpec` ABC owned by `jcm/dycore/` is the right long-term shape and is tracked as follow-up work.
* **Nudging stays inside the dinosaur backend for v2.** `DinosaurDycore` takes `nudging`/`start_date`/`calendar` at construction and composes the relaxation tendency into its IMEX-RK substages exactly as before. The clean move (promote `Nudging` to a `PhysicsTerm`) requires moving `NudgingTarget` from modal to gridpoint space, which is its own focused refactor — tracked as a follow-up.
* **v2 = major version bump, no shims.** `_final_modal_state`→`_final_dycore_state` is a hard rename. The on-disk checkpoint format changed (`modal_leaves`→`dycore_leaves`); pre-v2 checkpoints will not load. The three lifted conversion functions moved to `jcm.dycore.dinosaur.state_bridge`; in-repo callers updated in this same PR.

## Deferred to follow-up PRs (intentionally out of scope)

* Move terrain spectral-smoothing out of `TerrainData.{from_coords,from_file}` into `DinosaurDycore.build_terrain` (currently a thin shim around the existing factories).
* Move xarray-output dispatch out of `utils._infer_dims_shape_and_coords` into `DinosaurDycore.to_xarray` (currently calls `data_to_xarray` unchanged).
* Delete `utils.get_coords` and relocate to `jcm/dycore/dinosaur/coords.py`.
* Promote `Nudging` to a `PhysicsTerm`.

Doing all of the above in this PR would expand the diff substantially without changing the protocol shape or the numerics; each move is a focused follow-up that can be reviewed independently.

## Test plan

- [x] `pytest jcm/dycore/dinosaur/regression_test.py` — 1-day T21L8 SPEEDY trajectory matches captured baseline at rtol=1e-10, atol=1e-12 (same-host, single-precision XLA backend).
- [x] `pytest jcm/dycore/` — 19 protocol-shape tests pass, including the cubed-sphere layout canary.
- [x] `JAX_PLATFORMS=cpu pytest -n 12 -m "not slow"` — 920 passed, 13 skipped, 0 failed.
- [x] `JAX_PLATFORMS=cpu pytest -n 12 -m slow` — 43 passed, 17 skipped, 0 failed.
- [ ] 3-month T63L47 production-style run on GPU as a soak test (kicking off now; will report back).

Refs #388.

---

## Phase 2 status (added 2026-05-20)

Phase 2 (`PyscesDycore`) is **parked pending an upstream pySEs install fix**: `pip install git+https://github.com/OkayHughes/pySEs.git` produces a broken environment due to relative-imports-beyond-top-level (root cause + fix proposal in [OkayHughes/pySEs#1](https://github.com/OkayHughes/pySEs/issues/1)). Per discussion on this PR, we're filing the bug upstream and waiting rather than vendoring or forking. Phase 2 work will resume here once pySEs ships a fixed release; Phase 1's protocol surface is unaffected and stays mergeable.


jcm/checkpoint.py
```diff
@@ -86,25 +86,25 @@ def load_checkpoint(model, path) -> float:
...
-    modal_leaves_template = _flatten_arrays(model._final_modal_state)
+    dycore_leaves_template = _flatten_arrays(model._final_dycore_state)
     physics_leaves_template = _flatten_arrays(model._final_physics_state)
     template = {
         "elapsed_days": 0.0,
-        "modal_leaves": modal_leaves_template,
+        "dycore_leaves": dycore_leaves_template,
         "physics_leaves": physics_leaves_template,
     }
     payload = flax.serialization.from_bytes(template, Path(path).read_bytes())
```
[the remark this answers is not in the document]
duncanwp: Acknowledging but intentionally not adding the dual-key fallback. v2 is a major version bump (release notes / PR body call this out) and the agreed policy for this work is _no legacy/compat shims_ — pre-v2 checkpoints don't roll forward.

What I did do in eadf64a is improve the failure mode: `load_checkpoint` now detects the old-key payload, retries against the legacy template, and if THAT succeeds raises an explicit `"this is a pre-v2 checkpoint, regenerate it"` error rather than the cryptic `KeyError: 'dycore_leaves'` flax otherwise produces. Users with in-flight pre-v2 runs see a clear path forward instead of a confusing stack trace.

jcm/model.py
```diff
@@ -307,291 +286,129 @@ def integrate(x_initial):
...
+        # ``dt_si`` for sim-time bookkeeping and to pass to physics tendency
+        # capping (``verify_tendencies``).
+
+        self.physics = physics if physics is not None else speedy_physics()
+
+        if dycore is None:
+            if coords is None:
+                raise ValueError(
+                    "Model requires either an explicit ``dycore`` or a "
+                    "``coords`` argument (used to build the default "
```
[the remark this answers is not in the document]
duncanwp: Good catch — fixed in eadf64a. `DynamicalCore.tracer_specs` is now a public, documented attribute on the ABC, and `Model.__init__` writes `self.dycore.tracer_specs = {name: spec}` after the dycore is in place regardless of whether it was Model-built or caller-supplied. Renamed `DinosaurDycore._tracer_specs` -> `tracer_specs` (public) and updated `FakeCubedSphereDycore` + the trivial fake in `base_test.py` to match. Refactor-regression at rtol=1e-10 still passes.

jcm/model.py
```diff
@@ -86,38 +72,40 @@ def to_xarray(self):
...
 
-        pred_ds = data_to_xarray(dynamics_predictions.asdict() | physics_preds_dict,
-                                 coords=coords, serialize_coords_to_attrs=False,
-                                 times=times - times[0],
-                                 additional_coords=additional_coords)
+        pred_ds = data_to_xarray(
+            dynamics_predictions.asdict() | physics_preds_dict,
+            coords=coords, serialize_coords_to_attrs=False,
+            times=times - times[0],
+            additional_coords=additional_coords,
```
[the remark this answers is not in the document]
duncanwp: Yes, this is intentional for Phase 1 — called out in the PR body under _Deferred to follow-up PRs_. The dinosaur path needs no behaviour change here, so I left `ModelPredictions.to_xarray()` calling `data_to_xarray` directly to keep the diff scoped to the protocol introduction. `DynamicalCore.to_xarray(predictions, times)` is already on the ABC and `DinosaurDycore.to_xarray` is a thin shim around the existing helper — the contract is locked in. The follow-up PR will move `utils._infer_dims_shape_and_coords` and the axis dispatch into `DinosaurDycore.to_xarray`, after which `ModelPredictions.to_xarray()` will route through `self._dycore.to_xarray(...)`. Tracking under the existing follow-up todo on issue #388.

Cubed-sphere backends won't get a useful xarray export until that follow-up lands; `FakeCubedSphereDycore.to_xarray` here is intentionally a placeholder.

jcm/data/test/dycore_refactor_baseline/speedy_t21l8_1day.npz
duncanwp, line 1: This is too big for the repo, please remove it and just keep it locally for the duration of the work

jcm/model.py
```diff
@@ -1,61 +1,48 @@
+"""User-facing :class:`Model` class.
+
+The :class:`Model` orchestrates a simulation: forcing, run/resume/run_from_state,
+chunked op-split scan, post-processing, and xarray conversion. The actual
+dynamical-core operations (state initialisation, the per-``dt`` step, the
+gridpoint↔native bridge) live behind a :class:`DynamicalCore` (Phase-1 of
```
duncanwp: Don't refer to ongoing issues or PRs in the comments, it gets out of date. This mentions they dycore because that's what you're working on, but the physics is equally important, mention that too

jcm/model.py
```diff
@@ -307,291 +289,139 @@ def integrate(x_initial):
...
+                when ``dycore`` is ``None`` (so that a default
+                :class:`DinosaurDycore` can be constructed). To enable SPMD
+                parallelization, pass ``spmd_mesh`` to the coords helper
+                (e.g. :func:`get_speedy_coords`).
+            time_step: Model time step in minutes.
+            terrain: :class:`TerrainData` describing boundary conditions
+                (orography, land-sea mask, etc.). Defaults to an aquaplanet.
+            physics: :class:`Physics` describing the model physics. Defaults
+                to :func:`speedy_physics`.
+            diffusion: :class:`DiffusionFilter` describing horizontal diffusion
```
duncanwp: Remove this, it's dycore specific and should be specified by the Dinosaur constructor if wanted

jcm/model.py
```diff
@@ -307,291 +289,139 @@ def integrate(x_initial):
...
+                filter parameters. Used by the default :class:`DinosaurDycore`;
+                ignored when a non-dinosaur dycore is passed in.
+            start_date: ``jax_datetime.Datetime`` for the start of the run.
+            calendar: Calendar string for ``tyear`` / ``model_year`` /
+                forcing-axis alignment. ``"365_day"`` (no leap years) matches
+                SPEEDY's climatology tables; ``"gregorian"`` (365.2425
+                days/year) is available for ICON-style runs that align against
+                real Gregorian timestamps in their forcing files.
+            nudging: Optional :class:`jcm.nudging.Nudging` handle. When provided
```
duncanwp: If this belongs to the dycore it should be specified there, if it belongs in the physics it should go there, neither require it here. I thought we promoted nuding -> physics to Phase 1 so this should be in scope

jcm/model.py
```diff
@@ -307,291 +289,139 @@ def integrate(x_initial):
...
-        # no-op for other radiation backends). ``None`` means auto-detect.
+        # no-op for other radiation backends).
         if radiation_chunk_size is not None:
             from jcm.physics.radiation import rrtmgp as _rrtmgp_mod
             _rrtmgp_mod.set_chunk_size(radiation_chunk_size)
 
-        self.physics_specs = PHYSICS_SPECS
         self.dt_si = (time_step * units.minute).to(units.second)
-        self.dt = self.physics_specs.nondimensionalize(self.dt_si)
+        # Nondimensional dt is owned by the dycore; the Model only needs
```
duncanwp: Don't leave comments that refer to deleted code, it's confusing

jcm/checkpoint.py
```diff
@@ -86,25 +86,50 @@ def load_checkpoint(model, path) -> float:
...
+        # key ``modal_leaves`` -> ``dycore_leaves`` (PR #489). flax's msgpack
+        # codec raises on the key mismatch when an old checkpoint is loaded
+        # under the new code; surface the cause clearly rather than letting
+        # the underlying ``KeyError: 'dycore_leaves'`` confuse users.
+        old_template = {
+            "elapsed_days": 0.0,
+            "modal_leaves": dycore_leaves_template,
+            "physics_leaves": physics_leaves_template,
+        }
+        try:
```
duncanwp: This is nice, but not necassary. The checkpointing was never officially released pre v2. Drop backwards compatibility

jcm/dycore/dinosaur/dycore.py
```diff
@@ -0,0 +1,437 @@
...
+        calendar: str = "365_day",
+        diffusion: DiffusionFilter | None = None,
+    ):
+        """Initialise the dinosaur backend; see the class docstring for argument semantics."""
+        self.coords = coords
+        self.terrain = terrain
+        self.dt_seconds = float(dt_seconds)
+        self.start_date = start_date
+        self.calendar = calendar
+        self.nudging = nudging
```
duncanwp: We don't need nudging in the dycore if we're doing it through physucs

duncanwp: Especially since it brings with it a whole load of date handling - drop it and all the date stuff from the dycore

jcm/model.py
```diff
@@ -307,291 +289,139 @@ def integrate(x_initial):
...
+                real Gregorian timestamps in their forcing files.
+            nudging: Optional :class:`jcm.nudging.Nudging` handle. When provided
+                and ``dycore`` is ``None``, the default :class:`DinosaurDycore`
+                composes a Newtonian relaxation tendency into its IMEX-RK
+                substages.
+            radiation_chunk_size: Override the RRTMGP chunked-vmap chunk size
+                (cells per chunk). Default ``None`` auto-detects from the JAX
+                device's HBM. Has no effect when the radiation backend is not
+                RRTMGP.
+            dycore: An explicit :class:`DynamicalCore` instance. When provided,
```
duncanwp: Move dycore to the top as the first argument, it's funamental to the Model

jcm/dycore/dinosaur/dycore.py
```diff
@@ -0,0 +1,437 @@
...
+        *,
+        dt_seconds: float | None = None,
+    ) -> State:
+        """Advance ``state`` by one ``dt``.
+
+        Order: forward-Euler add of the physics dynamics-tendency →
+        IMEX-RK SIL3 dynamics step → spectral filters. Mirrors the original
+        :meth:`Model._get_op_split_step_fn` body.
+        """
+        if dt_seconds is not None and dt_seconds != self.dt_seconds:
```
duncanwp: So don't ask for it? We don't require it

jcm/dycore/dinosaur/dycore.py
```diff
@@ -0,0 +1,437 @@
...
+            raise NotImplementedError(
+                "DinosaurDycore.step does not currently support a per-call "
+                "dt override; the IMEX-RK step is built once at construction."
+            )
+        if physics_tendency is not None:
```
duncanwp, line 306: Again, this is unsupported, so remove passing tendencies through here

jcm/dycore/dinosaur/regression_test.py
duncanwp, line 1: Don't commit this file either, just keep it locally while we work on the refactor

jcm/physics_interface.py
```diff
@@ -1,710 +1,417 @@
...
+This module defines the gridpoint-space data structures that physics packages
+consume and produce — :class:`PhysicsState` and :class:`PhysicsTendency` — and
+the :class:`Physics` base class that they implement.
+
+It is **dycore-agnostic**: no spectral transforms or ``dinosaur`` symbols
+appear here. The actual dycore↔physics conversion is owned by each backend
+under ``jcm/dycore/<backend>/state_bridge.py`` (see
+:mod:`jcm.dycore.dinosaur.state_bridge` for the canonical example).
+
+Date originally 2/7/2024. Refactored 2026-05-17 to drop the dinosaur-specific
```
duncanwp: Remove the date comment

jcm/physics_interface.py
```diff
@@ -1,710 +1,417 @@
...
+        tracers=clipped_tracer_tends,
+    )
+
+
+def compute_physics_step_gridpoint(
+    physics_state: PhysicsState,
+    forcing: ForcingData,
+    terrain: TerrainData,
+    physics_state_carry,
+    *,
```
duncanwp, line 384: Why the asterix here?

jcm/physics_interface.py
```diff
@@ -1,710 +1,417 @@
...
+logger = logging.getLogger(__name__)
+
+
+@tree_math.struct
+class PhysicsState:
+    u_wind: jnp.ndarray
+    v_wind: jnp.ndarray
+    temperature: jnp.ndarray
+    specific_humidity: jnp.ndarray
+    geopotential: jnp.ndarray
```
duncanwp, line 51: What if the dycore physical state is in terms of vorticity and divergence?

jcm/data/test/dycore_refactor_baseline/speedy_t21l8_1day.npz
duncanwp: Done in 253b3b7 — baseline `.npz` and `regression_test.py` both untracked from the branch and added to `.gitignore`. Keeping the local copy so I can still self-check that the dinosaur path stays bit-identical (rtol=1e-10) through follow-up refactors; the cross-hardware `composable_physics_regression_test` stays as the shipped guardrail.

jcm/model.py
duncanwp: Rewritten in 253b3b7. Module docstring no longer mentions #388 / PR refs and now gives equal billing to physics and the dycore as the two component contracts the Model routes between.

jcm/model.py
duncanwp: Done — `diffusion` is off `Model.__init__` entirely. `runners.build_model` constructs the `DinosaurDycore` with the diffusion config it built, then passes the dycore in. Custom diffusion is now exclusively a `DinosaurDycore` constructor kwarg.

jcm/model.py
duncanwp: Done — `nudging` is off `Model.__init__` and now lives in physics as `NudgingTerm(PhysicsTerm)`. Callers do `physics + NudgingTerm(target, config)` or `with_nudging(physics, target, config)`. Phase-1 promotion as originally planned.

jcm/model.py
duncanwp: Removed.

jcm/checkpoint.py
duncanwp: Done — `load_checkpoint` is now a straight-line `flax.serialization.from_bytes` call; no legacy-key detection, no helpful error. v2 checkpoints only.

jcm/dycore/dinosaur/dycore.py
duncanwp: Done — `DinosaurDycore` is nudging-free. The relaxation tendency now flows through the physics path via `NudgingTerm`.

duncanwp: Done in the same change — `start_date`, `calendar`, and `_date_from_sim_time` are all gone from `DinosaurDycore`. The IMEX-RK builder is now a one-liner over `self._primitive`.

jcm/model.py
duncanwp: Done — `Model(dycore, *, coords=..., physics=..., ...)`. The dycore is the first positional argument; the implicit-default coords / terrain path is preserved for the convenience case (`dycore=None`).

jcm/dycore/dinosaur/dycore.py
duncanwp: Removed.

jcm/dycore/dinosaur/dycore.py
duncanwp: Removed — `DynamicalCore.step(state, physics_tendency)` is the only signature now.

jcm/dycore/dinosaur/regression_test.py
duncanwp: Done — `jcm/dycore/dinosaur/regression_test.py` untracked from the branch and gitignored. Kept locally.

jcm/physics_interface.py
duncanwp: Removed.

jcm/physics_interface.py
duncanwp: Was about the `*,` separator — I left it (kept the explanation in chat), per your reply. No code change here.

jcm/physics_interface.py
duncanwp: Per your follow-up, just added a header comment to `PhysicsState` explicitly noting that the dycore's *internal* representation can be vor/div (or anything else) and the conversion happens inside `dycore.to_physics_state`. `PhysicsState` is the physics-facing contract; it stays u/v.

Issue #496, opened 2026-06-02
Aerosol: gas-phase chemistry coupling beyond SimpleChemistry


duncanwp:
The harness uses the existing `SimpleChemistry` term for SO2→SO4 / DMS oxidation. Add a fuller gas-phase oxidant/chemistry coupling (prognostic or prescribed OH/O3/H2O2 fields driving aqueous + gas-phase sulfate production) for the online aerosol path.

Tracked as out-of-scope for the initial aerosol harness (#461). See the harness plan in that issue (comment) / `.claude/aerosol_harness_plan.md`.

duncanwp:
### MAM4-JAX-specific scope (from #490 / #510)

The MAM4-JAX microphysics core (#510) needs gas-phase **H₂SO₄** and **SOAG** as inputs to its condensation + binary H₂SO₄–H₂O nucleation. They are currently seeded to **zero** in the adapter (`jcm/physics/aerosol/jam/microphysics/mam4_jax.py`), so nucleation/condensation are driven only by the core's internal hardcoded `~1e-16` mol/mol/s H₂SO₄ production — i.e. decoupled from jcm's sulfur cycle. Closing this under the broader gas-phase chemistry work here means:

- **Prognostic gas tracers** in the JAM tracer layout: H₂SO₄ and SOAG (optionally SO₂, DMS).
- **Gas-phase chemistry** that produces gas H₂SO₄ (SO₂ + OH → H₂SO₄) and SOAG (VOC oxidation) — the substance of this issue (`SimpleChemistry` today only does ozone/methane).
- **Adapter**: pack these tracers into the MAM4 `q` gas slots (H₂SO₄ pcnst 6, SOAG 9, SO₂ 2, DMS 3) and read post-step gas changes back as tracer tendencies so the gas↔aerosol mass transfer closes.
- **Remove the core's internal hardcoded H₂SO₄ production** once real gas is supplied (avoid double-counting + the small non-conservative source noted in the wrapper docstring).

Acceptance: sulfate condensation/nucleation respond to jcm's SO₂/DMS emissions + oxidants; gas + aerosol sulfur budget closes; column + T21 runs stay finite. Refs #490, #510.

duncanwp:
Merged in #511

Issue #495, opened 2026-06-02
Aerosol: dust (and aerosol) optics in SW/LW radiation


duncanwp:
Couple the online aerosol population's optical properties (from `_ham_state` + species refractive indices) into the RRTMGP/grey radiation bands, so online aerosol has a direct radiative effect. Initially the harness only affects clouds via activation; this adds the direct effect, replacing MACv2-SP optics for the online path.

Tracked as out-of-scope for the initial aerosol harness (#461). See the harness plan in that issue (comment) / `.claude/aerosol_harness_plan.md`.

duncanwp:
Closed by #509

Issue #494, opened 2026-06-02
Aerosol: heterogeneous ice nucleation (dust/BC)


duncanwp:
Add heterogeneous freezing (immersion/deposition) driven by dust and BC populations, feeding the cloud microphysics ice number. Out of scope for the initial harness which only does ARG warm-cloud activation.

Tracked as out-of-scope for the initial aerosol harness (#461). See the harness plan in that issue (comment) / `.claude/aerosol_harness_plan.md`.

Issue #493, opened 2026-06-02
Aerosol: SOA volatility basis set (VBS)


duncanwp:
The placeholder/MAM4 path treats SOA simply. Add a volatility basis set treatment for secondary organic aerosol partitioning (biogenic + anthropogenic precursors), coupled to the emissions `biogenic_voc` source.

Tracked as out-of-scope for the initial aerosol harness (#461). See the harness plan in that issue (comment) / `.claude/aerosol_harness_plan.md`.

PR #497, opened 2026-06-02
Add online aerosol harness (HAM) — phases 0–7 (#461)

duncanwp:
## Summary

Builds out the online aerosol module from #461 as a HAMMOZ-style,
microphysics-agnostic **process harness** wrapping an interchangeable modal
**microphysics core**, expressed in JCM's composable `PhysicsTerm`
architecture. Shipped in 8 per-phase commits (0–7).

- **Population contract** shaped to the real MAM4 4-mode layout and named in
  AMBRS's modal vocabulary, so the swap point for a future MAM4-JAX core
  (#490) is a clean interface, not a rewrite. part2pop is intentionally *not*
  a dependency (numpy/scipy, non-differentiable) — a future diagnostics
  adapter (#492).
- **Process terms**: online emissions, κ-Köhler equilibrium placeholder
  microphysics, ARG activation (with an optional, off-by-default Ghosh 2025
  variant), Stokes sedimentation, resistance-model dry deposition, and
  in-/below-cloud wet deposition. All differentiable.
- **Wiring**: `echam_physics(aerosol_module="ham", cloud_scheme="2m")`. The
  HAM path *augments* MACv2-SP (kept for the aerosol optics/Twomey diagnostic
  radiation + clouds hard-require) and ARG's `activated_cdnc` is preferred by
  the 2M scheme over the SPA floor when present (#342).

Design doc: `docs/design/aerosol_harness.md`. Detailed plan:
`.claude/aerosol_harness_plan.md` + #461 comment.

## What reviewers should know

- **MAM4 constants** (modes/species/σ/density/κ) are transcribed from the
  E3SM Fortran (`rad_constituents.F90`, `modal_aero_data.F90`),
  cross-checked against MAM4-JAX @ `ccd872b`. They are physical constants,
  not GPL code.
- **Ghosh 2025 ARG coefficients** were *reconstructed* by fitting the paper's
  Table 3 (the inline-equation values I first extracted were garbled). The
  variant is gated off by default and flagged in-code as needing PDF
  verification. The `arg2000` default is the textbook scheme.
- **Best-effort vs verified**: per-term physics is unit-tested for
  self-consistency (mass conservation, monotonic trends, finite gradients,
  the ARG Table-3 oracle). Quantitative gates that need external reference
  data — AeroCom emission budgets, parcel-model activation, Slinn/Ganzeveld
  curves — are implemented to the published formulae but **not** validated
  against datasets we don't carry. Emission magnitudes are order-of-magnitude
  defaults.
- **Cloud-borne tracer mirror** (`mc_`/`nc_`) is declared for the future MAM4
  core but is currently inert (no term writes it).
- The real MAM4-JAX core is **not yet differentiable upstream**; the harness
  itself is. Wiring it in is #490.

Out-of-scope items were filed as issues: #490 (MAM4-JAX core), #491
(sectional/bulk), #492 (part2pop adapter), #493 (SOA VBS), #494 (het
freezing), #495 (aerosol optics in radiation), #496 (gas-phase chemistry).

## Test plan

- [x] `pytest jcm/physics/aerosol/ham -m "not slow"` — 64 unit tests
  (population, tracer layout, placeholder κ-Köhler, ARG + Ghosh Table-3
  oracle, sedimentation mass conservation, dry/wet deposition, emissions,
  factory wiring + full-stack ordering validation). Plus `check_jvp`/`grad`
  finiteness per term.
- [x] `pytest jcm/physics/clouds/lohmann_2m_test.py -m "not slow"` — 47
  tests, no regression from the `activated_cdnc` read change.
- [x] `pytest jcm/physics/aerosol/ham/ham_integration_test.py` (slow) — T21
  aquaplanet, `aerosol_module="ham"` + 2M, runs finite; HAM tracers seeded
  and physical; `ghosh2025` variant also runs.
- [ ] Reviewer: confirm the per-phase commit split reads cleanly.


jcm/physics/clouds/lohmann_2m.py
```diff
@@ -3690,14 +3690,19 @@ def __call__(
...
-            cloud_fraction=cloud_fraction,
-            prefactor=self._spa_prefactor.get_value(),
-            exponent=self._spa_exponent.get_value(),
-        )
+        # Activated CDNC source: prefer an explicit ``activated_cdnc`` from an
+        # upstream activation term (e.g. HAM's ARG, #461) if present; otherwise
+        # fall back to the inline SPA floor derived from the MACv2-SP Nccn.
+        # Both are differentiable and produce the same (nlev, ncols) field.
+        activated_cdnc = diagnostics.get("activated_cdnc")
+        if activated_cdnc is None:
```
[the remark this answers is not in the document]
duncanwp: Good catch — fixed in 7ab35d4. The 2M scheme now always computes the MACv2-SP SPA floor and uses `jnp.where(arg_cdnc > 1.0, arg_cdnc, spa_floor)`: it prefers the online ARG `activated_cdnc` where it is non-empty and falls back to the SPA floor wherever ARG is ≈0 (e.g. before the zero-seeded HAM tracers spin up). So the default HAM+2M run activates droplets from step 0, and ARG takes over once aerosol builds up. (Used a small threshold rather than `maximum` so legitimately clean, spun-up columns aren't floored.)

jcm/physics/echam/echam_terms.py
```diff
@@ -151,11 +166,30 @@ def echam_physics(
...
+            f"Unknown aerosol_module={aerosol_module!r}. Choose 'macv2sp' "
+            "or 'ham'."
+        )
+
     return ComposablePhysics(
         terms=[
             MoistAirColumnState(),
             EchamBoundaryConditions(),
-            Macv2SpAerosol(params=aerosol_p),
+            *aerosol_terms,
```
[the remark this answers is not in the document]
duncanwp: Agreed — fixed in 7ab35d4. `ham_aerosol_physics()` is now split in the factory: terms 1–5 (emissions → core → ARG activation → sedimentation → dry dep) stay in the pre-cloud aerosol block (activation must precede the cloud term that consumes `activated_cdnc`), and `HamWetDeposition` is spliced in immediately **after** the cloud microphysics term so it scavenges against the current step's precip/condensate rather than the previous-step carry. `_ham_state`/`activated_fraction` written earlier in the step persist in the diagnostics dict, so wetdep still sees them. Verified finite at T21 (HAM + 2M).

docs/design/aerosol_harness.md
```diff
@@ -0,0 +1,103 @@
+# Online aerosol harness (HAM)
```
duncanwp: I don't like actually calling this 'HAM'. It's inspired by HAM and (currently) shares many of the specific parameterizations, but as with the main physics packages, I want to decouple those specific implementations from the package itself for maximum flexibility. 

Rename the aerosol harness 'the Jax Aerosol Module (JAM)'. Rename the wet/dry deposition, and sedimentation after the actual schemes

docs/design/aerosol_harness.md
```diff
@@ -0,0 +1,103 @@
...
+
+The `jcm.physics.aerosol.ham` package is a HAMMOZ-style, microphysics-agnostic
```
duncanwp: We can still call it 'HAMMOZ style' in the docs, it's a useful comparison for people familiar with this implementation

docs/design/aerosol_harness.md
```diff
@@ -0,0 +1,103 @@
...
+  **zero tendency** on the real MAM4 4-mode population (`mam4_data.py`,
+  constants from E3SM `rad_constituents.F90` / `modal_aero_data.F90`,
+  cross-checked against MAM4-JAX). Replacing it with the real MAM4-JAX core
+  swaps only the per-step microphysics, not the contract (#490).
+- **Tracers** (`tracer_layout.py`): aerosol mass and number are ordinary
+  `state.tracers` entries (`m_`/`mc_`/`n_`/`nc_` for interstitial/cloud-borne
+  mass/number), so the dynamical core transports them and existing
+  diagnostics work. The cloud-borne mirror is carried for the future MAM4
+  core (currently inert).
+- **Inter-term state** (`ham_state.py`): the core writes a typed
```
duncanwp: rename jam_state etc

docs/design/aerosol_harness.md
```diff
@@ -0,0 +1,103 @@
...
+  and sedimentation read it.
+
+Every harness step is differentiable; scheme choice (placeholder vs real core,
+ARG variant) is a compose-time Python decision with no traced branching.
+
+## Term chain
+
+`ham_aerosol_physics()` returns the ordered list:
+
+1. `HamEmissions` — sea salt / dust / DMS / volcanic / biogenic surface
```
duncanwp: JamEmissions.

For clarity, this should be "Sources -> lowest-layer tracers tendencies"

We emit in physical (nodal space; presumably) and Some dycore don't have modal representations.

docs/design/aerosol_harness.md
```diff
@@ -0,0 +1,103 @@
...
+to the MACv2-SP SPA floor wherever the online source is ≈0 (e.g. before the
+prognostic HAM tracers spin up from a zero-seeded initial state), so the
+default HAM+2M run always activates droplets.
+
+## Status and caveats
+
+- **Source magnitudes** (emissions) are order-of-magnitude defaults, not
+  inventory-calibrated; the AeroCom-budget acceptance gate needs reference
+  data not carried in-repo.
+- **Re-evaporation** re-injection in wet deposition is deferred (no per-level
```
duncanwp: Make an issue to track this, or just add the diagnostic if it's an easy fix

jcm/physics/aerosol/ham/activation/arg.py
```diff
@@ -0,0 +1,152 @@
...
+_RGAS = 8.314462    # universal gas constant [J/mol/K]
+_RHOW = 1000.0      # water density [kg/m³]
+_LV = 2.501e6       # latent heat of vaporization [J/kg]
+_CP = 1004.0        # specific heat of air [J/kg/K]
+_G = 9.80665        # gravity [m/s²]
+_DV = 2.11e-5       # water-vapour diffusivity [m²/s]
+_KA = 0.024         # thermal conductivity of air [J/m/s/K]
+_SIGMA_W = 0.072    # surface tension of water [J/m²]
+
+_TINY = 1.0e-30
```
duncanwp: We already have an epsilon defined in PhysicalConstants, use this and/or update it to match 1e-30

jcm/physics/aerosol/ham/activation/arg.py
```diff
@@ -0,0 +1,152 @@
...
+is the only data-dependent switch, handled with ``jnp.where``.
+"""
+
+from __future__ import annotations
+
+import jax.numpy as jnp
+from jax.scipy.special import erf
+
+# Physical constants (SI).
+_MW = 0.018015      # molar mass of water [kg/mol]
```
duncanwp: Please move and consolidate these to PhysicalConstants. Ideally deriving from existing constants

jcm/physics/aerosol/jam/activation/arg.py
```diff
@@ -0,0 +1,152 @@
...
+_TINY = 1.0e-30
+# Ghosh et al. (2025) σ_acc validity range.
+_SIGMA_ACC_LO = 1.4
+_SIGMA_ACC_HI = 2.1
+
+
+def _saturation_vapor_pressure(temperature: jnp.ndarray) -> jnp.ndarray:
+    """Tetens saturation vapour pressure over liquid water [Pa]."""
+    t_c = temperature - 273.15
+    return 611.2 * jnp.exp(17.62 * t_c / (temperature - 30.03))
```
duncanwp, line 54: Where do these numbers come from? THey look magic. Are they defined in a paper, are they empirical or do they refer to physical constants (in which case they should derive from those)

jcm/physics/aerosol/jam/activation/arg.py
```diff
@@ -0,0 +1,152 @@
...
+    sigma_acc: float,
+    variant: str,
+):
+    """Return ``(f, g, p)`` for the chosen ARG variant.
+
+    ``f``/``g`` multiply the two ARG terms; ``p`` is the exponent on
+    ``ζ/η`` (3/2 in the original scheme). For ``ghosh2025`` the three are
+    functions of the accumulation-mode width ``sigma_acc`` (applied to all
+    modes, per the paper), and ``p`` switches in the kinetically limited
+    ``ζ/η > 1`` regime.
```
duncanwp, line 69: Maybe add a note that this is deliberately not jittable, since it's only run once (I guess?)

jcm/physics/aerosol/jam/emissions/emissions_term.py
```diff
@@ -0,0 +1,116 @@
...
+        dz = diagnostics["layer_thickness"]
+        nlev, ncols = state.temperature.shape
+
+        u10 = jnp.sqrt(state.u_wind[-1] ** 2 + state.v_wind[-1] ** 2)
+        land = self._land_fraction(terrain, ncols)
+        ocean = 1.0 - land
+
+        seasalt = ocean * p.seasalt_coeff * u10 ** p.seasalt_wind_exp
+        dust = land * p.dust_coeff * jnp.maximum(u10 - p.dust_u_threshold, 0.0) ** 3
+        dms_so4 = ocean * p.dms_coeff * u10
```
duncanwp: These seem very simple, are they direct ports from HAM? What scheme do they correspond to? What are the implict/explicit assumptions used?

jcm/physics/aerosol/jam/emissions/emissions_term.py
```diff
@@ -0,0 +1,116 @@
...
+        volc = jnp.full((ncols,), p.volcanic_so4)
+        bio = jnp.full((ncols,), p.biogenic_soa)
+
+        fluxes = [
+            ("ss", "acc", seasalt * p.seasalt_accum_frac),
+            ("ss", "cor", seasalt * (1.0 - p.seasalt_accum_frac)),
+            ("du", "acc", dust * 0.1),
+            ("du", "cor", dust * 0.9),
```
duncanwp: Where do these fractions come from? Reference the HAM / scheme

jcm/physics/aerosol/ham/microphysics/mam4_data.py
```diff
@@ -0,0 +1,76 @@
...
+``ccd872b6592d0f1ffa7ee43e51f7952b999fd57b``. These are data, not code.
+"""
+
+from __future__ import annotations
+
+from jcm.physics.aerosol.ham.population import AerosolMode, ModalAerosolSpec
+from jcm.physics.aerosol.ham.species import SPECIES
+
+#: Upstream MAM4-JAX commit the constants below were cross-checked against.
+MAM4_JAX_COMMIT = "ccd872b6592d0f1ffa7ee43e51f7952b999fd57b"
```
duncanwp: Remove this it's not needed

jcm/physics/aerosol/jam/sedimentation/sedi_term.py
```diff
@@ -0,0 +1,140 @@
...
+        params = self.params.get_value()
+        ham = diagnostics["_ham_state"]
+        air_density = diagnostics["air_density"]
+        dz = diagnostics["layer_thickness"]
+        pressure = diagnostics["pressure_full"]
+        temperature = state.temperature
+
+        tracer_tends: dict[str, jnp.ndarray] = {}
+        for i, mode in enumerate(self._spec.modes):
+            if not mode.sediments:
```
duncanwp, line 126: Are there any easy optimizations/vectorizations available for this nested loop? It looks slow...

jcm/physics/aerosol/ham/sedimentation/sedi_term.py
```diff
@@ -0,0 +1,140 @@
...
+            v = params.velocity_scale * stokes_velocity(
+                ham.r_wet[i], ham.rho[i], temperature, pressure,
+            )
+            # Same velocity transports every interstitial tracer of the mode.
+            names = [number_name(mode.short)] + [
+                mass_name(sp, mode.short) for sp in mode.species
+            ]
+            for nm in names:
+                q = state.tracers.get(nm)
+                if q is None:
```
duncanwp: Why would q be 'None' ?

jcm/physics/aerosol/ham/wetdep/wetdep_term.py
```diff
@@ -0,0 +1,165 @@
+"""``HamWetDeposition`` — in-cloud and below-cloud aerosol scavenging.
+
+Two removal pathways for interstitial aerosol, both differentiable and built
+only from diagnostics the cloud scheme already exposes (so the cloud
+microphysics terms are untouched):
+
+* **In-cloud nucleation scavenging** — the activated fraction of aerosol is in
+  cloud droplets and is removed at the rate cloud condensate converts to
+  precipitation. The per-level precip-formation rate is reconstructed by
+  distributing the column surface precip (``CloudData.precip_rain/snow``)
```
duncanwp: Is this what ECHAM does? We could just expose the per-level precip formatino from the microphysics?

jcm/physics/aerosol/ham/wetdep/wetdep_term.py
```diff
@@ -0,0 +1,165 @@
...
+  across the cloudy column weighted by in-cloud condensate.
+* **Below-cloud impaction scavenging** — falling precipitation collects
+  interstitial aerosol in clear air, with a size-dependent (∝ r²) collection
+  efficiency so coarse particles are scavenged far faster than accumulation
+  mode.
+
+Mirrors ``mo_hammoz_wetdep``. Precip re-evaporation re-injection is deferred
+(no per-level evaporation-rate diagnostic is exposed yet).
```
duncanwp: This would be pretty easy to add though? Let's just do it unless there are good reasons not to

jcm/physics/aerosol/ham/wetdep/wetdep_term.py
```diff
@@ -0,0 +1,165 @@
...
+    def __call__(self, state, diagnostics, forcing, terrain):
+        params = self.params.get_value()
+        ham = diagnostics["_ham_state"]
+        activated_fraction = diagnostics["activated_fraction"]
+        air_density = diagnostics["air_density"]
+        dz = diagnostics["layer_thickness"]
+        nlev, ncols = state.temperature.shape
+
+        clouds = diagnostics.get("clouds")
+        if clouds is not None:
```
duncanwp: We shouldn't need these blocks here. If we have aerosol we can assume we have clouds - and anyway we can just include it in the 'requires' list and then the constructor takes care of this for us

jcm/physics/aerosol/jam/wetdep/wetdep_term.py
```diff
@@ -0,0 +1,165 @@
...
+
+        p_form = precip_formation_rate(
+            precip_col, cloud_fraction, qc, air_density, dz,
+        )
+        rate_incloud = params.incloud_scale * in_cloud_rate(
+            activated_fraction, p_form, qc,
+        )
+
+        tracer_tends: dict[str, jnp.ndarray] = {}
+        for i, mode in enumerate(self._spec.modes):
```
duncanwp, line 148: Yeah, this nested pattern has cropped up again. It would be nice to be able give the compiler the chance to optimize this out. Perhaps with a different view on the tracer list?

jcm/physics/aerosol/ham/ham_state.py
```diff
@@ -0,0 +1,54 @@
...
+
+The active microphysics core writes per-step physical state needed by the
+downstream harness terms (activation, dry/wet deposition, sedimentation):
+per-mode dry/wet radius, particle density and hygroscopicity, plus the
+gathered per-mode mass and number. Stored under the underscored diagnostics
+key ``"_ham_state"`` (inter-term plumbing, auto-flattened to ``ham_state.*``
+in xarray output, mirroring the ``_radiation`` / ``_humidity`` pattern).
+
+All arrays are shaped ``(n_modes, nlev, ncols)`` except per-mode number
+which is ``(n_modes, nlev, ncols)`` too — every field carries the mode axis
```
duncanwp: This comment doesn't make sense

jcm/physics/aerosol/ham/ham_state.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+from __future__ import annotations
+
+import jax.numpy as jnp
+import tree_math
+
+
+@tree_math.struct
+class HamAerosolState:
```
duncanwp: Given we want to be able to support bin and modal (and mixed) representatinos, I think we can trivially make this more general by just renaming the dimensions to be generic 'n_aer' (or similar) and coming up with a name for the mass and number that is more agnostic. There's nothing in here that's intrinsically modal

jcm/physics/aerosol/ham/species.py
```diff
@@ -0,0 +1,48 @@
...
+# so the future MAM4 adapter can map tokens to Fortran species slots.
+SPECIES: dict[str, AerosolSpecies] = {
+    "so4": AerosolSpecies("so4", molar_mass=0.115, density=1770.0,
+                          hygroscopicity=0.507, long_name="sulfate"),
+    "nh4": AerosolSpecies("nh4", molar_mass=0.018, density=1770.0,
+                          hygroscopicity=0.507, long_name="ammonium"),
+    "no3": AerosolSpecies("no3", molar_mass=0.062, density=1770.0,
+                          hygroscopicity=0.507, long_name="nitrate"),
+    "poa": AerosolSpecies("poa", molar_mass=0.012, density=1000.0,
+                          hygroscopicity=0.010, long_name="p-organic"),
```
duncanwp: spell these out for the long-name: Parriculate-organic, secondary-organic, black-carbon, ...

jcm/physics/aerosol/jam/tracer_layout.py
```diff
@@ -0,0 +1,71 @@
+"""Tracer-name conventions for the HAM aerosol harness.
+
+Aerosol mass and number live as ordinary entries in ``state.tracers`` so the
+dynamical core transports them and the existing diagnostics infrastructure
+works unchanged. MAM4 carries both an *interstitial* population and a
+*cloud-borne* mirror, so every quantity has an interstitial and a
+cloud-borne key.
+
```
duncanwp, line 11: Again, I feel this could be made more general very easily. We will want to support modal and sectional schemes, some with in-cloud  aerosol and some not, so lets not over constrain ourselves in these scaffolding classes (even just in the doc-strings)

jcm/physics/echam/echam_terms.py
```diff
@@ -151,16 +166,46 @@ def echam_physics(
             f"Unknown cloud_scheme={cloud_scheme!r}. Choose '1m' or '2m'."
         )
 
+    # MACv2-SP always provides the ``aerosol`` optics/Twomey diagnostic that
```
duncanwp: *always* is too strong; we do this in-lieue of the proper radiation and microphsyics coupling  - it's a fudge for now, but once JAM (né HAM) is wired up we won't necassarily include MacSP contributions

jcm/physics/echam/echam_terms.py
```diff
@@ -151,16 +166,46 @@ def echam_physics(
...
+    # must precede the cloud term that consumes ``activated_cdnc``).
+    ham_post_cloud_terms: list[PhysicsTerm] = []
+    if aerosol_module == "macv2sp":
+        aerosol_terms = [Macv2SpAerosol(params=aerosol_p)]
+    elif aerosol_module == "ham":
+        from jcm.physics.aerosol.ham.ham_terms import ham_aerosol_physics
+        ham_terms = ham_aerosol_physics(
+            microphysics=ham_microphysics, arg_variant=ham_arg_variant,
+        )
+        ham_post_cloud_terms = [
```
duncanwp: This doesn't 'smell' great. I'll accept it for now but if we find other ordering dependencies or issues between terms we'll have to come up with a better solution

docs/design/aerosol_harness.md
duncanwp: Done in 5d3492a — package renamed `ham/` → `jam/` (the Jax Aerosol Module), and the terms renamed after their schemes: `StokesSedimentation`, `SlinnDryDeposition`, `WetScavenging`, plus `JamEmissions`. Design doc renamed to `docs/design/jam.md`.

docs/design/aerosol_harness.md
duncanwp: Kept — the doc still describes JAM as "HAMMOZ-style" as a familiar point of comparison, just decoupled from the name.

docs/design/aerosol_harness.md
duncanwp: Done — `ham_state.py` → `jam_state.py`, `HamAerosolState` → `JamAerosolState`, `_ham_state` → `_jam_state` (5d3492a).

docs/design/aerosol_harness.md
duncanwp: Done — `JamEmissions`, reworded to "Sources → lowest-layer tracer tendencies", with a note that emission is computed in gridpoint (nodal) space and the class split belongs to the microphysics core's population, so no modal representation is assumed on the dynamics side (8a97e62).

docs/design/aerosol_harness.md
duncanwp: Filed #499 — expose per-level precip formation + evaporation from the cloud schemes (and add re-evaporation re-injection). Deferred rather than done inline because it changes the return signatures of both stability-tuned warm-rain schemes (see the wetdep thread).

jcm/physics/aerosol/ham/activation/arg.py
duncanwp: The existing `epsilon` is 1e-12, which is too coarse here — aerosol volumes are ~1e-18 m³/kg, so 1e-12 would wrongly floor real values. Rather than change the global epsilon, I added a distinct `tiny = 1e-30` to `PhysicalConstants` and use that (107ba58).

jcm/physics/aerosol/ham/activation/arg.py
duncanwp: Done (107ba58) — moved to `PhysicalConstants`: added `r_universal`, `m_air` (= R*/Rd), `m_water` (= eps·M_air), `surface_tension_water`, `vapor_diffusivity`, `air_thermal_conductivity`; reuse `grav`/`cp`/`alhc`/`rhow`/`ak`. Same consolidation applied to the sedimentation/dry-dep constants (removed the resistances→sedi cross-import).

jcm/physics/aerosol/jam/activation/arg.py
duncanwp: They're the empirical Magnus-Tetens saturation coefficients (WMO / Alduchov & Eskridge 1996) — not derivable from fundamental constants. Cited and labelled as such in the docstring (107ba58).

jcm/physics/aerosol/jam/activation/arg.py
duncanwp: Clarified in the docstring: it actually *is* jittable. `variant` is a plain Python string fixed at compose time, so the `if variant ==` branch is resolved during tracing (compile-time static dispatch), not a runtime branch. Only the `jnp.where` on ζ/η is a traced switch (107ba58).

jcm/physics/aerosol/jam/emissions/emissions_term.py
duncanwp: They're deliberately *simplified placeholders*, not ports of a specific scheme — documented as such now (8a97e62): sea salt a bare u10 power law (cf. Monahan 1986 / Gong 2003 without the size-resolved whitecap form), dust a threshold-cubed wind law (cf. Marticorena & Bergametti 1995, no erodibility map), DMS→SO4 a linear wind proxy. The real prescribed CEDS path with HAMMOZ-grounded characteristics is #498.

jcm/physics/aerosol/jam/emissions/emissions_term.py
duncanwp: Documented as ad-hoc placeholder splits (8a97e62). The size-resolved, source-specific mode assignment grounded in HAMMOZ (cmr_ff/bb, 2.5%% primary-SO4 Aitken/accum split) lands with the CEDS work in #498.

jcm/physics/aerosol/ham/microphysics/mam4_data.py
duncanwp: Removed (8a97e62) — the provenance commit hash stays in the module docstring.

jcm/physics/aerosol/jam/sedimentation/sedi_term.py
duncanwp: Done (8a97e62) — the per-mode/per-tracer loop now stacks all interstitial tracers + their matching settling velocity and runs the donor-cell transport once via `jax.vmap`, so XLA gets one batched op instead of an unrolled op per tracer.

jcm/physics/aerosol/ham/sedimentation/sedi_term.py
duncanwp: Good question — it turned out to be load-bearing: `Model.get_empty_data` traces the physics once with a zero `PhysicsState` whose tracer dict is *empty*, so a declared tracer is genuinely absent there (only in that structural probe; never in a real run). I kept graceful handling via a zeros fallback and documented exactly this (3411b2e).

jcm/physics/aerosol/ham/wetdep/wetdep_term.py
duncanwp: Agreed that's the right approach — filed #499 to expose the true per-level precip formation (and evaporation) from the cloud schemes and consume it directly. The current column-precip reconstruction is documented as interim.

jcm/physics/aerosol/ham/wetdep/wetdep_term.py
duncanwp: It's a bit more than a quick add — it changes the return signatures of *both* stability-tuned warm-rain schemes (1M + 2M) to thread per-level formation/evaporation out of their vmapped column functions, so I'd rather do it as a focused PR with cloud-microphysics regression testing. Tracked in #499; the wetdep interface (clouds in `requires`, post-cloud ordering) is already in place so it's a drop-in once the diagnostics exist.

jcm/physics/aerosol/ham/wetdep/wetdep_term.py
duncanwp: Done (8a97e62) — `clouds` is now in `requires` and read directly; the None-fallback block is gone (wetdep runs after the cloud term, so clouds is upstream).

jcm/physics/aerosol/jam/wetdep/wetdep_term.py
duncanwp: Done (8a97e62) — same treatment as sedimentation: stack the tracers + per-tracer rate and apply the removal as one batched elementwise op.

jcm/physics/aerosol/ham/ham_state.py
duncanwp: Fixed (107ba58) — that "except … too" sentence was nonsense; the docstring now just says every field is shaped `(n_aer, nlev, ncols)` with the class axis first.

jcm/physics/aerosol/ham/ham_state.py
duncanwp: Done (107ba58) — `JamAerosolState` is now representation-agnostic: leading axis `n_aer` (mode *or* bin), fields `mass`/`number`, and the docstring no longer assumes modal.

jcm/physics/aerosol/ham/species.py
duncanwp: Done (107ba58) — long-names spelled out: particulate-organic, secondary-organic, black-carbon, sea-salt, marine-organic.

jcm/physics/aerosol/jam/tracer_layout.py
duncanwp: Done (107ba58) — `tracer_layout` and `population` docstrings now talk in terms of aerosol *classes* (mode or bin), note the cloud-borne mirror is optional per scheme, and point at sectional support (#491). Nothing modal is baked in.

jcm/physics/echam/echam_terms.py
duncanwp: Softened (5d3492a/8a97e62) — the comment now says MACv2-SP is kept *for now* as a temporary fudge in lieu of proper JAM aerosol↔radiation/microphysics coupling (#495), and need not be included once JAM supplies those.

jcm/physics/echam/echam_terms.py
duncanwp: Acknowledged — left as-is for now (splitting the chain so only wet dep runs post-cloud is the minimal correct fix). If more inter-term ordering dependencies crop up I agree we'll want a general mechanism (declared ordering constraints / a proper scheduler) rather than hand-placing terms.

duncanwp:
Thanks for the thorough review — all comments addressed across four commits:

- **`5d3492a`** — rename HAM → **JAM** (Jax Aerosol Module): `jam/` package, `_jam_state`, `aerosol_module="jam"`, and terms renamed after their schemes (`StokesSedimentation`, `SlinnDryDeposition`, `WetScavenging`, `JamEmissions`). "HAMMOZ-style" kept as a comparison.
- **`107ba58`** — generalize the scaffolding (`JamAerosolState` → `n_aer` axis + `mass`/`number`, representation-agnostic docstrings, spelled-out species long-names) and consolidate physical constants into `PhysicalConstants` (cite Magnus-Tetens; clarify the ARG variant branch is compile-time static dispatch / jittable).
- **`8a97e62`** — vectorize the sedimentation/wet-scavenging tracer loops (stack + batched op), `WetScavenging` now takes `clouds` in `requires` (dropped the None fallback), emissions provenance documented as simplified placeholders, `MAM4_JAX_COMMIT` removed, design doc → `docs/design/jam.md`.
- **`3411b2e`** — fix a probe regression the review surfaced: `Model.get_empty_data` traces the physics with an empty tracer dict, so the deposition/sedimentation gathers keep a zeros fallback (documented).

Two items deferred to focused follow-ups with reasoning in-thread:
- **#499** — expose true per-level precip formation/evaporation from the cloud schemes (+ re-evaporation re-injection). Touches both warm-rain schemes' return signatures, so it deserves its own PR with cloud regression tests; the wet-scavenging interface is already wired for it.
- **#498** — prescribed CEDS emissions with HAMMOZ-grounded, differentiable per-sector characteristics (replaces the placeholder source magnitudes/splits).

Verified: 64 fast unit tests + 2 GPU integration tests (T21 aquaplanet, JAM + 2M, `arg2000` + `ghosh2025`) green; existing Lohmann-2M tests unaffected; lint clean.

duncanwp:
Replaced the simplified natural-emission placeholders with faithful HAMMOZ scheme ports, each its own `PhysicsTerm` + calibratable `Parameters` (commits `d5de88b`, `ebeee73`):

- **`SeaSaltEmissions`** — full **Gong (2003)** (`seasalt_emissions_gong`, nseasalt=6). The size-only part of the Gong source function is wind-independent, so the 300-bin integral precomputes to two constants per mode (accum, coarse); the per-step flux is those × u10³·⁴¹ × open-water fraction `(1−land)(1−seaice)`. `SeaSaltParameters{scale, wind_exponent}`.
- **`DmsEmissions`** — **Nightingale (2000)** oceanic DMS: smooth piston velocity `kw=0.222u²+0.333u`, `(Sc/660)^-½` with the Andreae DMS Schmidt number, × prescribed seawater DMS field. No gas-phase sulfur tracer exists here, so the sulfur is emitted as primary sulfate into Aitken/accum (interim; full DMS→SO2→SO4 gas chemistry is #496). `DmsParameters{flux_scale, aitken_fraction}`.
- **`DustEmissions`** — **Tegen et al. (2002)** wind-erosion physics: threshold-friction-velocity saltation `G ∝ (ρ/g)u*³(1+u*t/u*)(1−(u*t/u*)²)` × sandblasting `α` × prescribed source/erodibility field. The 17-soil-type database is folded into the calibratable `source`/`α`/`u*t` knobs (`DustParameters`).

**Surface fields:** added `dms_seawater` and `dust_source` to `ForcingData` (trailing `None` defaults; read from the forcing file when present, **fall back to zero on read** — so DMS/dust are simply inert until the fields are supplied, as on the aquaplanet). The simplified `JamEmissions`/`EmissionParameters` are removed; `jam_aerosol_physics` now composes `[SeaSalt, Dms, Dust]` and the distributor is retained/reused.

All key parameters are differentiable for future calibration. Verified: 78 fast unit tests (incl. each scheme's physics, ocean/land/ice & threshold gating, plausible sea-salt magnitude, and finite gradients) + the 2 GPU integration tests (T21 aquaplanet, JAM + 2M — Gong sea salt actively emitting, stays finite). Lint clean.

PR #315, opened 2025-11-25
Implementing simplified Betts Miller convection

AndrewILWilliams:
## Describe your changes
WORK IN PROGRESS, nothing runs yet (thanks Claude).

Blows up after ≈ 10 hrs, generating 200% RH after a few hours in lower levels

## Issue ticket number and link
Linked to #314 

## Checklist before requesting a review
- [ ] I have performed a self-review of my code
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #508, opened 2026-06-09
feat(convection): Betts-Miller convective adjustment (Isca port, all 3 flavors) — closes #314

### LINKED ISSUE #314: simplified Betts-Miller convection scheme
As with most of SPEEDY's physical parameterizations, the convection scheme is 'hard-coded' to 8 vertical levels (you _can_ run with more, but it would not make much sense). In the idealized AGCM community, a standard 'simple convection scheme' which is flexible and somewhat physically-based is [Dargan Frierson's 2007 'Simplified Betts-Miller' scheme](https://journals.ametsoc.org/view/journals/atsc/64/6/jas3935.1.xml). 

Both [Isca](https://github.com/ExeClim/Isca/blob/master/src/atmos_param/qe_moist_convection/qe_moist_convection.F90) and [SpeedyWeather.jl](https://github.com/SpeedyWeather/SpeedyWeather.jl/blob/main/src/physics/convection.jl) implement this scheme. 

It's simpler than ECHAM6's convection scheme, but represents a useful stepping stone between that and SPEEDY.

duncanwp:
## Summary

Implements the **Betts-Miller / Frierson (2007) Simplified Betts-Miller** convective-adjustment scheme as a composable `PhysicsTerm`, ported fresh from Isca's [`betts_miller.f90`](https://github.com/ExeClim/Isca/blob/master/src/atmos_param/betts_miller/betts_miller.f90). Closes #314.

This supersedes the stale WIP #315 — a single-flavor SpeedyWeather.jl port on the pre-composable API (`jcm.geometry.Geometry`, `speedy_physics.py`) that never ran (it blew up to ~200% RH). Given the API drift and the broken state, I started fresh from the authoritative Fortran and matched the current term API; the new scheme's stability test directly guards that old failure mode.

## Design: one term, three flavors via a config field

All three Isca "flavors" share the entire deep precipitating core — parcel-ascent CAPE → moist-adiabatic `Tref` and `rhbm·qsat` `qref` → relaxation over `tau_bm` → energy-consistency correction — and differ **only** in how a column that would dry net-negative is handled. So they're one `BettsMillerConvection` term selected by a static config field (`params.shallow`), as discussed:

| `ShallowScheme` | Isca flag | behaviour |
|---|---|---|
| `SIMP` (default) | `do_simp` | Frierson 2007: shorten a relaxation timescale to keep precip ≥ 0, energy-consistent |
| `SHALLOWER` | `do_shallower` | raise the cloud top until column precip ≥ 0, trim the boundary layer |
| `CHANGEQREF` | `do_changeqref` | scale `qref`/`Tref` to force precip = 0 |
| `NONE` | (all flags off) | suppress when the deep adjustment would dry the column |

plus the `do_envsat` (reference RH wrt environment vs. parcel) and `do_taucape` (`tau ∝ CAPE^-1/2`) modifiers.

## What's here

- **`betts_miller.py`** — column physics: a `lax.scan` parcel ascent (dry→moist adiabat, CAPE/CIN, cloud-top index), reference profiles, the deep relaxation with the `precip`/`precip_t` energy match, and the three shallow branches; `vmap`-ed over the grid. Vectorized, jittable, **differentiable**; works on sigma or hybrid grids at any number of levels. Constants read via attribute access from `jcm.constants` (honours `set_constants`).
- **`params.py`** — static `BettsMillerParameters` (Isca `betts_miller_nml` defaults: `tau_bm=7200 s`, `rhbm=0.8`, `do_simp`). Flavor/modifier flags are static because they select genuinely different code paths.
- **`betts_miller_terms.py`** — `BettsMillerConvection` PhysicsTerm (`category="convection"`): builds `p_half = a + b·ps`, converts g/kg↔kg/kg, emits a `betts_miller_precip` diagnostic.

## Verification (`betts_miller_test.py`, 9 tests, all pass)

- No-CAPE column → zero tendencies.
- Moist unstable column → positive precip, warming + drying, and **energy consistency** (`precip_from_q == precip_from_T`).
- Every flavor × `do_envsat` × `do_taucape` across deep and shallow columns → finite, `precip ≥ 0`.
- Shallow flavors → non-precipitating and **conserve column moisture**.
- **Stability: 200-step column integration keeps RH < 1.05** (the #315 blow-up was ~200%) and temperatures physical.
- Vectorized driver matches the single-column routine; **gradients are finite**; the PhysicsTerm runs end-to-end on a grid.

## Usage

```python
from jcm.physics.convection.betts_miller import (
    BettsMillerConvection, BettsMillerParameters, ShallowScheme)

physics = ComposablePhysics(terms=[
    ...,
    BettsMillerConvection(BettsMillerParameters(
        shallow=ShallowScheme.SHALLOWER, do_envsat=True)),
])
```

## Notes / follow-ups

- Worked in SI specific humidity (Frierson 2007 / SpeedyWeather.jl convention, which the issue cites); Isca's mixing-ratio form differs only at second order, negligible for this idealized adjustment.
- A full multi-day aquaplanet integration with BM swapped in for SPEEDY/ECHAM convection would be a good follow-up validation, but is beyond a unit-test PR.

### DIFF
--- a/jcm/physics/convection/betts_miller/betts_miller.py
+++ b/jcm/physics/convection/betts_miller/betts_miller.py
@@ -18,16 +18,17 @@ Conventions
 * Physical constants are read by attribute access from :mod:`jcm.constants` so a
   ``set_constants`` override is honoured.
 
-The per-column routine :func:`betts_miller_column` is ``vmap``-ed over the
-horizontal grid by :func:`betts_miller_tendencies`; the flavor (``params.shallow``)
-and modifiers are *static* and resolved by Python branching at trace time.
+The routines are *broadcasting-native*: the vertical level is axis 0 and any
+trailing axes are horizontal and broadcast numpy-style, so the identical code
+runs on a single ``(kx,)`` column, a ``(kx, ncols)`` vectorized block, or a
+whole ``(kx, ix, il)`` grid — no ``vmap`` or reshaping, and no awareness of how
+the host arranges the horizontal dimension. The flavor (``params.shallow``) and
+modifiers are *static* and resolved by Python branching at trace time.
 """
 
 from __future__ import annotations
 
-import jax
 import jax.numpy as jnp
-import numpy as np
 from jax import lax
 
 import jcm.constants as c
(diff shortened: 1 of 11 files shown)


jcm/physics/convection/betts_miller/betts_miller.py
```diff
@@ -0,0 +1,366 @@
...
+from jax import lax
+
+import jcm.constants as c
+from jcm.physics.convection import saturation as _saturation
+from jcm.physics.convection.betts_miller.params import (
+    BettsMillerParameters,
+    ShallowScheme,
+)
+
+_T_FLOOR = 173.16  # K — below this the parcel ascent is presumed CAPE-free (Isca).
```
duncanwp: Put this in BettsMillerParameters - it's definitely tunable

jcm/physics/convection/betts_miller/betts_miller.py
```diff
@@ -0,0 +1,366 @@
...
+    All inputs are 1-D column profiles ordered top (0) -> surface (kx-1).
+
+    Returns:
+        tp: parcel temperature [K], (kx,)
+        cloud: bool mask, True on levels from the level of zero buoyancy down to
+            the surface (the layers the deep adjustment relaxes), (kx,)
+        cape: convective available potential energy [J/kg], scalar
+
+    """
+    kappa, hlv, cp, rv, rd = c.akap, c.alhc, c.cpd, c.rv, c.rd
```
duncanwp: Don't do this - just use our variables. This makes it harder to track the actual values used when skimming code

jcm/physics/convection/betts_miller/betts_miller.py
```diff
@@ -0,0 +1,366 @@
...
+    pstar = c.p0
```
duncanwp: Ditto

jcm/physics/convection/betts_miller/betts_miller.py
```diff
@@ -0,0 +1,366 @@
...
+        phalf: half-level pressure [Pa], (kx+1,).
+        dt: timestep [s].
+        params: static :class:`BettsMillerParameters`.
+
+    Returns:
+        (tdel, qdel, precip): temperature & humidity increments over ``dt``
+        [(kx,), (kx,)] and column precipitation [kg/m^2], scalar.
+
+    """
+    hlv, cp, grav = c.alhc, c.cpd, c.grav
```
duncanwp: Don't rename variables

jcm/physics/convection/betts_miller/params.py
```diff
@@ -0,0 +1,72 @@
...
+    #: Scale the reference humidity (and temperature) profiles so the
+    #: column-integrated precip is exactly zero (Isca ``do_changeqref``).
+    CHANGEQREF = "changeqref"
+    #: No shallow scheme: zero the tendencies when the deep adjustment would dry
+    #: the column (Isca with all shallow flags false).
+    NONE = "none"
+
+
+@dataclass(frozen=True)
+class BettsMillerParameters:
```
duncanwp, line 48: No, these need to be pytrees like all the other Physics parameters so we can calculate gradients with respect to them

jcm/physics/convection/saturation.py
duncanwp, line 1: Good. Let's add some tests around this though (or move the relevant Tiedke/Betts tests)

Issue #498, opened 2026-06-05
Prescribed CEDS anthropogenic aerosol emissions (grid-flexible, super-sectored)


duncanwp:
Follow-up to the aerosol harness (#461, PR #497). Replace the harness's order-of-magnitude placeholder anthropogenic sources with prescribed **CEDS** (JGCRI) emissions, gridded **on demand** to whatever dycore/resolution is in use, with HAMMOZ-style per-sector emission characteristics.

Detailed implementation plan: `.claude/aerosol_emissions_plan.md`.

### Source
- Anthropogenic: **CEDS** — aggregate Zenodo DOI `10.5281/zenodo.15059443`; canonical **0.5° gridded sectored** product (input4MIPs/ESGF). Grid from the *gridded sectored* product (country→grid needs spatial proxies = reimplementing CEDS gridding).
- Open biomass burning: GFED/van Marle (separate dataset, Phase E).

### Hosting & size
- Mirror a **subset** as chunked **zarr** on an object store (R2 / HF datasets) for transparent partial reads via `xr.open_zarr`; a zipped netCDF snapshot on **Zenodo** per release for a citeable DOI; in-repo `registry.txt` (sha256) for version tracking.
- Keep the mirror ~100s MB (not 10s GB) by **dimensional** reduction: aerosol-relevant species only; **3 super-sectors** (not 8); **annual fields + 12-month seasonal climatology**; float32 + zlib (+ optional bit-rounding); 0.5° (or 1°) ceiling. Users pull only the species/years they need (tens of MB), cached thereafter.

### Super-sector mapping (grounded in HAMMOZ source)
HAMMOZ varies emissions by **injection type** and **source size**, encoded as a small static characteristics table:

| Super-sector | CEDS sectors | Injection | Size |
|---|---|---|---|
| `surface_combustion` | TRA, RCO, AGR, WST, SLV | surface (lowest level) | fossil-fuel 0.03 µm (Aitken) |
| `elevated_industrial` | ENE, IND | ~50 m (2nd level) | fossil-fuel 0.03 µm |
| `shipping` | SHP | marine surface | fossil-fuel 0.03 µm |
| `biomass_burning` (Phase E) | GFED | fire profile | biomass 0.075 µm |

[...]

Issue #491, opened 2026-06-02
Aerosol: sectional / bulk microphysics support in the harness


duncanwp:
The `ModalAerosolSpec` contract carries a `family` field (`modal|sectional|bulk`) but only `modal` is implemented. Add sectional (SALSA-style bins) and/or bulk population builders + emissions distributors so the harness is microphysics-family-agnostic, as HAMMOZ is.

Tracked as out-of-scope for the initial aerosol harness (#461). See the harness plan in that issue (comment) / `.claude/aerosol_harness_plan.md`.

PR #513, opened 2026-06-11
feat(aerosol): prescribed CEDS/biomass emissions — pipeline, wiring & CAM-faithful path (#498)

duncanwp:
Implements #498: prescribed CEDS anthropogenic + open biomass-burning aerosol emissions for JAM, grid-flexible and differentiable. Builds on the merged gas-coupling (#496) and harness (#497).

## What's here

**Two complementary prescribed-emission paths** — both differentiable w.r.t. the emission `ForcingData` fields:

1. **Bulk / differentiable** (`AnthropogenicEmissions`, `jam_anthropogenic=True`) — bulk per-super-sector SO₂/BC/OC → **in-model** HAMMOZ speciation (2.5 % primary SO₄, Aitken/accum split, OM:OC=1.4) + a **smooth Gaussian vertical profile** so injection height / SO₄-fraction are *calibratable by gradient* (a hard level pick has none). Four super-sectors: surface / elevated-industrial / shipping / **biomass_burning** (deep FIRE profile).
2. **CAM6-faithful pre-speciated** (`PreSpeciatedEmissions`, `jam_prescribed_speciated=True`) — reads already-speciated per-tracer fields and injects them directly (surface → bottom layer; 3-D volume → per level), mirroring CAM's `mo_srf_emissions` / `mo_extfrc`. No in-model speciation, but still differentiable w.r.t. the emission fields. For bit-faithful reproduction of CESM emissions / validation.

**Data pipeline** (`jcm/data/emissions/`):
- `regrid.py` — light, first-order **conservative** (area-weighted) remap from any lat/lon **or** unstructured `ncol` source onto the model grid (**0.07 % global-mass error** ne30→T31).
- `prepare.py` — mapping-driven source→contract assembly: `cesm_cmip_anthro` / `cesm_bb4cmip7` (bulk) and `cesm_mam4_speciated` / `prepare_speciated_emissions` (pre-speciated) adapters for the CESM CMIP7 files.
- `downloader.py` — host-agnostic fetch (local path / arbitrary URL), **no ESGF coupling**.

**Wiring** — `ForcingData` carries emissions as dict-valued fields (`anthropogenic_emissions`, `prescribed_aerosol_emissions`) with `read_anthropogenic_emissions` / `read_prescribed_aerosol_emissions` loaders; `select(date)` slices the per-channel `TimeSeries` automatically. Opt-in via `echam_physics(jam_anthropogenic=…, jam_prescribed_speciated=…)`.

## Scope decision (data investigation)
The raw 0.5° gridded CEDS is **ESGF-only** (Zenodo carries only CSV totals; only the CESM ne30 MAM4-processed product is on disk). So the original ESGF-fetch pipeline and mirror-upload were **dropped** in favour of the general regridder + documented emissions-file contract + host-agnostic downloader, assuming the user supplies a conformant model-grid file. Self-hosting a compressed 0.5° mirror is tracked in #515.

## Validation
The pre-speciated adapter on the real ne30 files reproduces CESM's global sulfur budget (**~46.6 Tg S/yr, 2014**) and **recovers the HAMMOZ 2.5 % primary-sulfate split to 3 decimals** — a clean cross-check of the mode mapping, number encoding, and elevated-integration.

## Test plan
- [x] Injection profile, bulk speciation (S-conserving), BC/OC→primary-carbon, grad finiteness (height/thickness/SO₄-fraction).
- [x] Biomass `FIRE` profile deeper than surface; grad through biomass injection height.
- [x] Pre-speciated term: surface + 3-D mass conservation, multi-tracer, `∂(mmr)/∂(emission field)` finite.
- [x] Conservative regridder incl. a real ne30 reference (<0.5 % mass error).
- [x] Contract round-trips (both loaders + `select` slicing, 2-D & 3-D).
- [x] End-to-end T21 JAM runs for **both** paths raise the expected tracer burden.
- [x] Guarded budget tests (2.5 % recovery + global S). Full fast suite green; lint clean.
- [x] Emissions-file contract documented in `docs/design/jam.md`.

## Follow-up
- #515 — self-host a compressed 0.5° CEDS mirror (feed the bulk path without ESGF).

### DIFF
--- a/jcm/config/physics/echam-jam.yaml
+++ b/jcm/config/physics/echam-jam.yaml
@@ -14,19 +14,20 @@
 #
 # Both emission terms are enabled but **inert until a matching file is given**,
 # so a plain ``physics=echam-jam`` run is the zero-emission JAM baseline.
+#
+# Defaults use the real MAM4-JAX core, which is an optional GPL extra:
+#     pip install jcm[mam4]
+# Without it, override the core: ``physics.jam_microphysics=placeholder``.
 
 builder: echam_physics
 
 aerosol_module: jam
 cloud_scheme: 2m            # ARG activated_cdnc feeds the 2-moment scheme
-radiation_scheme: grey      # set radiation_scheme=rrtmgp for RRTMGP
-jam_microphysics: placeholder
-jam_arg_variant: arg2000    # or ghosh2025
+radiation_scheme: rrtmgp    # set radiation_scheme=grey for the cheap grey scheme
+jam_microphysics: mam4_jax  # the MAM4-JAX core (#490); "placeholder" for κ-Köhler
+jam_arg_variant: ghosh2025  # or arg2000
 jam_aqueous_scheme: full
 
 # Prescribed-emission terms (driven by forcing.emissions_file):
 jam_anthropogenic: true        # bulk SO2/BC/OC → in-model speciation + injection
 jam_prescribed_speciated: true # already-speciated per-tracer fields (CAM-faithful)
-
-# Consumed by Model (not the factory); null = auto-detect from device HBM.
-radiation_chunk_size: null
--- a/jcm/config/physics/echam-jam.yaml
+++ b/jcm/config/physics/echam-jam.yaml
@@ -1,0 +1,32 @@
+# ECHAM physics with the JAM prognostic aerosol module.
+#
+# Unlike the term-list presets (``echam.yaml`` etc.), this is **factory-built**:
+# ``physics.builder`` tells ``runners.build_physics`` to call
+# ``echam_physics(**these_kwargs)``, which encodes the JAM aerosol chain's
+# validated ordering (the chain is split around the cloud term, so a flat term
+# list can't express it). Only scalar flags below are forwarded; per-scheme
+# ``Parameters`` keep their ``.default()``.
+#
+# End-to-end usage (supply emissions on the model grid; auto-routes by content):
+#
+#     python -m jcm.main physics=echam-jam grid=echam_t42_l8_sigma \
+#         forcing.emissions_file=/path/to/emissions_on_model_grid.nc
+#
+# Both emission terms are enabled but **inert until a matching file is given**,
+# so a plain ``physics=echam-jam`` run is the zero-emission JAM baseline.
+
+builder: echam_physics
+
+aerosol_module: jam
+cloud_scheme: 2m            # ARG activated_cdnc feeds the 2-moment scheme
+radiation_scheme: grey      # set radiation_scheme=rrtmgp for RRTMGP
+jam_microphysics: placeholder
+jam_arg_variant: arg2000    # or ghosh2025
+jam_aqueous_scheme: full
+
+# Prescribed-emission terms (driven by forcing.emissions_file):
+jam_anthropogenic: true        # bulk SO2/BC/OC → in-model speciation + injection
+jam_prescribed_speciated: true # already-speciated per-tracer fields (CAM-faithful)
+
+# Consumed by Model (not the factory); null = auto-detect from device HBM.
+radiation_chunk_size: null
--- a/jcm/data/emissions/regrid.py
+++ b/jcm/data/emissions/regrid.py
@@ -24,6 +24,15 @@ Design (kept deliberately light — no ESMF/xesmf dependency):
 The source grid is described purely by per-cell ``(lon, lat, area)`` triples, so
 structured and unstructured sources go through the same path — only the adapter
 that produces those triples differs.
+
+Note — this is intentionally *not* the same remapper as
+``jcm.data.bc.interpolate`` (``upsample_forcings_ds``): that path **bilinearly
+interpolates** smooth structured lat/lon boundary fields (SST, soil moisture, …)
+to a higher spectral resolution and is *not* mass-conserving, whereas emissions
+are **flux** fields where conserving the column-integrated source is the property
+that matters, and their sources are often unstructured (``ncol``). Different
+problems (non-conservative bilinear upsampling vs conservative,
+unstructured-capable flux remap), hence kept separate.
 """
 
 from __future__ import annotations
--- a/jcm/data/emissions/regrid.py
+++ b/jcm/data/emissions/regrid.py
@@ -1,0 +1,152 @@
+"""General, mass-conserving regridding of emission fields onto the model grid.
+
+Emissions reach jax-gcm on whatever grid their source happens to use — a regular
+0.5° lat/lon CEDS product, or an unstructured ``ncol`` mesh like the CESM ne30
+files in ``inputdata/atm/cam/chem/emis/cmip7``. This module remaps any such
+source onto the model's spectral Gaussian nodal grid **conservatively**, so the
+global emitted mass is (to first order) preserved — the property that matters for
+a flux boundary condition.
+
+Design (kept deliberately light — no ESMF/xesmf dependency):
+
+* The regridder is **first-order conservative / area-weighted**. Each source cell
+  is assigned, by nearest cell centre, to exactly one target cell; the target
+  value is the source-area-weighted mean of the cells that land in it
+  (``Σ fₛ Aₛ / Σ Aₛ``). For *coarsening* (fine source → coarse model grid, the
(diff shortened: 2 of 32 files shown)


jcm/config/physics/echam-jam.yaml
```diff
@@ -0,0 +1,32 @@
...
+jam_microphysics: placeholder
+jam_arg_variant: arg2000    # or ghosh2025
+jam_aqueous_scheme: full
+
+# Prescribed-emission terms (driven by forcing.emissions_file):
+jam_anthropogenic: true        # bulk SO2/BC/OC → in-model speciation + injection
+jam_prescribed_speciated: true # already-speciated per-tracer fields (CAM-faithful)
+
+# Consumed by Model (not the factory); null = auto-detect from device HBM.
+radiation_chunk_size: null
```
duncanwp: This is redundant now, please remove

jcm/config/physics/echam-jam.yaml
```diff
@@ -0,0 +1,32 @@
...
+#         forcing.emissions_file=/path/to/emissions_on_model_grid.nc
+#
+# Both emission terms are enabled but **inert until a matching file is given**,
+# so a plain ``physics=echam-jam`` run is the zero-emission JAM baseline.
+
+builder: echam_physics
+
+aerosol_module: jam
+cloud_scheme: 2m            # ARG activated_cdnc feeds the 2-moment scheme
+radiation_scheme: grey      # set radiation_scheme=rrtmgp for RRTMGP
```
duncanwp: make this rrtmgp by default

jcm/config/physics/echam-jam.yaml
```diff
@@ -0,0 +1,32 @@
...
+jam_microphysics: placeholder
```
duncanwp: Make this mam4 by default

jcm/config/physics/echam-jam.yaml
```diff
@@ -0,0 +1,32 @@
...
+jam_arg_variant: arg2000    # or ghosh2025
```
duncanwp: ghosh2025 by default

jcm/data/emissions/regrid.py
duncanwp, line 1: Don't we have one of these in the forcing data folder? Try and consolidate

jcm/physics/aerosol/jam/emissions/anthropogenic.py
```diff
@@ -0,0 +1,182 @@
...
+            )
+            so2 = p.scale * self._flux(forcing, f"emis_{sector}_so2", ncols)
+            bc = p.scale * self._flux(forcing, f"emis_{sector}_bc", ncols)
+            oc = p.scale * self._flux(forcing, f"emis_{sector}_oc", ncols)
+
+            # SO2 → primary SO4 (Aitken/accum) + g_so2 gas (S-conserving).
+            frac = p.so4_primary_fraction[i]
+            so4_mass = frac * so2 * SO2_TO_SO4_MASS
+            ait, acc = SO4_MODES
+            add_aerosol("so4", ait, so4_mass * SO4_AITKEN_FRACTION, weights)
```
duncanwp: Wait, I *think* everything up to here is agnostic to the actual aerosol population representation - which is what we want, but this hardcodes the aitken mode. Let's try and keep this as general as possible to make the sectional implementation easier

duncanwp: Maybe the Population itself should decide?

jcm/physics/aerosol/jam/emissions/sectors.py
```diff
@@ -0,0 +1,91 @@
...
+    # pyroconvection lofts the tail far higher), so these are deliberately just
+    # defaults — ``injection_height``/``injection_thickness`` are differentiable
+    # and meant to be calibrated.
+    "biomass_burning": SectorDefaults(injection_height=1000.0,
+                                      injection_thickness=1500.0),
+}
+
+# --- HAMMOZ species-handling constants (differentiable defaults on the term) --
+SO4_PRIMARY_FRACTION = 0.025     # fraction of SO2 sulfur → primary SO4 (zfacso2)
+SO4_AITKEN_FRACTION = 0.5        # primary-SO4 split into Aitken vs accum
```
duncanwp: Yeah, the modal assumption crops up here too - let's try and avoid it. How does HAMMOZ deal with this?

jcm/forcing.py
```diff
@@ -237,6 +237,24 @@ class ForcingData:
     dms_seawater: Any = None   # seawater DMS concentration (DmsEmissions)
     dust_source: Any = None    # dust source / erodibility 0–1 (DustEmissions)
 
+    # Prescribed anthropogenic aerosol emissions (#498). A single mapping
```
duncanwp, line 240: Can these not both be the same field? And a (possibly already existing) boolean flag just decides how to distribute them?

jcm/runners.py
```diff
@@ -598,6 +645,71 @@ def _attach_ozone(forcing, forcing_cfg, coords):
...
+    import xarray as xr
+
+    from jcm.forcing import (
+        default_forcing,
+        read_anthropogenic_emissions,
+        read_prescribed_aerosol_emissions,
+    )
+
+    ds = xr.open_dataset(path)
+    anthro = read_anthropogenic_emissions(ds)
```
duncanwp, line 685: These might need to deal with multiple files (e.g. separate for biomass burning and the rest.) I think open_mfdataset could deal with this transparently for us

duncanwp:
Rebased onto `dev` (picked up single-vmap RRTMGP #517 + ice nucleation #514) and addressed the review. Summary of resolutions (force-pushed as `3ef1805`):

**Population-owned emission distribution (the modal-hardcoding comments).** `ModalAerosolSpec` now has `classes_for(species)` and a `primary_emission` policy table queried via `primary_split(species)`. `AnthropogenicEmissions`, `DustEmissions` and `SeaSaltEmissions` ask the spec which classes carry a species instead of naming `ait/acc/cor/pcm`; `SO4_MODES`/`SO4_AITKEN_FRACTION`/`CARBON_MODE` moved out of `sectors.py` into the MAM4 spec (`_MAM4_PRIMARY_EMISSION`). Implied number now uses dev's `mode.number_factor` (and `particle_mean_mass` delegates to it, so the log-normal moment lives in one place). A sectional population (#491) drops in through the same calls. *How HAMMOZ does it:* same centralized table in its aerosol module — primary SO₄ split soluble-Aitken/accumulation by a fixed mass fraction, primary BC/OC → insoluble Aitken; we mirror that placement on the spec.

**Sea salt — fully generalized.** The Gong source is now partitioned across the spec's `ss`-carrying classes by each class's `[dgnum_lo, dgnum_hi]` (every bin assigned to its containing/nearest class → **mass-conserving**), replacing the hardcoded HAMMOZ-M7 diameter bands. ⚠️ This is a deliberate **physics change**: the accum/coarse boundary now follows the MAM4 mode edges rather than the fixed 1 µm cut, so total sea-salt emission is conserved but redistributed acc↔coarse. Flagging for your eyes.

**`echam-jam` preset defaults:** `radiation_scheme: rrtmgp`, `jam_microphysics: mam4_jax`, `jam_arg_variant: ghosh2025`, and dropped `radiation_chunk_size` (removed upstream by #517). Note `mam4_jax` is the optional `jcm[mam4]` extra — documented in the preset; the factory test uses `placeholder`/`grey` overrides so the base CI image still exercises the wiring.

**`emissions_file`** now accepts a list → merged via `open_mfdataset` (e.g. biomass-burning + anthropogenic files with disjoint channels).

**Codex P2 (injection underflow):** the all-underflow fallback now routes to the *nearest* layer (top if above-column, bottom if below-surface) instead of always the surface, so a calibrated high-altitude injection isn't silently turned into a surface source.

**Two emission fields:** kept separate per your call (distinct schemas/terms); the single auto-routing CLI `emissions_file` is the unification point.

**Regrid consolidation:** documented why it stays separate from `jcm.data.bc.interpolate` — that's non-conservative bilinear upsampling of smooth structured fields; emissions need conservative, unstructured-capable *flux* remapping. Different problem, so not merged.

Fast suites green; running a T21 jam-mam4 integration with the CESM emissions to sanity-check loadings before this is review-ready.

PR #518, opened 2026-06-13
perf(aerosol): MAM4 condensation speedups — relaxed tolerances + gate + operator-split substep backend

duncanwp:
## Summary

The MAM4-JAX microphysics core — a per-cell `diffrax Kvaerno5` *implicit* ODE solve, vmapped over all cells — is essentially the entire cost of a JAM aerosol step (~1426 ms vs a 34 ms non-JAM baseline at T63L47/A100). This PR makes the condensation solve **selectable** and defaults it to a fast operator-split backend, plus relaxes tolerances and gates cold-start pathologies.

### Condensation backends (`condensation_backend` on `Mam4JaxMicrophysics`)

- **`"substep"` (new default)** — operator-split: analytic (exact) H2SO4 + `n_substeps` fixed substeps that integrate the frozen-`g_star` linear SOA sub-ODE with its **exact closed form** per substep. No adaptive while-loop ⇒ no `max_steps` fragility / worst-cell gating. ~19× faster than the relaxed-tolerance diffrax path (~55× over the tight upstream default) at ~0.3%/step vs the tight reference.
- **`"astem"`** — the Fortran-faithful adaptive scheme (upstream semi-implicit step1/step2 SOA with adaptive `dtcur = alpha/tmpa` via a per-cell `while_loop`, + analytic H2SO4). Matches the CAM/E3SM discretization. ~14× faster than relaxed diffrax.
- **`"diffrax"`** — the original adaptive Kvaerno5 (kept for reference/validation), governed by the relaxed `rtol=1e-6 / atol=1e-15` tolerances + the cold-start gate below.

Requires a mam4_jax with `configure_condensation` (**reflective-org/MAM4-JAX#59**); if absent, the wrapper **falls back to `"diffrax"` with a warning** so a stale install still runs.

### Tolerances + cold-start gate (the `"diffrax"` path)

- Relax to `rtol=1e-6 / atol=1e-15` via `solvers.configure()` (**reflective-org/MAM4-JAX#58**) — ~2.8× on the core for ~0.13%/step.
- Sanitise inputs to finite non-negative mixing ratios; run `throw=False`; fall back to a zero tendency for non-finite cells. The full step now runs from a cold start (previously aborted at `max_steps` / timed out). A `jax.debug.callback`→`logger.warning` (gated by `lax.cond`) surfaces non-convergence counts without per-step host sync.

## Multi-day validation (full ECHAM + JAM-MAM4 T21 aquaplanet, 3 sim-days, cold start)

| backend | wall (incl ~200 s compile) | end-to-end speedup | 3-day global aerosol burden |
|---|---|---|---|
| diffrax | 972 s | 1× | 5.56e12 |
| **substep** | 230 s | **4.2×** | 3.26e13 |
| astem | 244 s | 4.0× | 3.26e13 |

- **substep and astem agree to 0.18%** on 3-day global burden — the fast default is validated against the faithful CAM scheme in a real integration.
- **diffrax is the outlier** (~6× lower burden): from a cold start the gate suppresses aerosol in cells that hit `max_steps`. It is both the slowest and the least reliable cold-start path — further reason it is no longer the default.
- astem's worst-cell gating is mild on a real heterogeneous state (only ~6% over substep).

## Why not float32

float32 gives zero speed/memory benefit for the core on the A100 — the implicit solve is latency-bound, not fp64-throughput-bound. De-prioritised.

## Tests

- Parametrised backend run-finite check (substep + astem) through the full wrapper; default-backend assertion (`"substep"`); input-sanitisation and non-convergence-gate/log tests. 11 wrapper tests pass.

## Remaining before full production confidence

The 3-day run is cold-start spin-up, not equilibrium. The 0.2% substep–astem agreement is encouraging; confirm over a longer / spun-up run. All three backends are config-selectable so reverting is a one-liner.

### DIFF
--- a/jcm/physics/aerosol/jam/microphysics/mam4_jax.py
+++ b/jcm/physics/aerosol/jam/microphysics/mam4_jax.py
@@ -146,7 +146,7 @@ class Mam4JaxMicrophysics(ModalMicrophysicsTerm):
         rtol: float = 1e-6,
         atol: float = 1e-15,
         max_steps: int = 4096,
-        condensation_backend: str = "diffrax",
+        condensation_backend: str = "substep",
         n_substeps: int = 4,
     ):
         """Import the core, enable float64, configure the solver, precompute maps.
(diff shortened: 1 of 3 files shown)


jcm/physics/aerosol/jam/microphysics/mam4_jax.py
```diff
@@ -89,19 +91,19 @@
...
+    Precision is **not** forced here — it is decided per-instance in
+    ``Mam4JaxMicrophysics.__init__`` (see ``enable_x64`` there). The ``diffrax``
+    condensation backend needs float64; the ``substep`` / ``astem`` backends are
+    float32-safe (the coag ``qv12`` coefficient was reformulated upstream to
+    avoid a float32 underflow), so they can run the whole coupled model in
+    float32. Importing ``mam4_jax`` still enables float64 *by default* (unless
+    ``MAM4_JAX_ENABLE_X64=0``); the wrapper overrides it as needed.
     """
     global _CORE
-    jax.config.update("jax_enable_x64", True)
```
duncanwp, line 102: What exactly is this _CORE thing doing now if the config update happens implictiyl in the import? Can't we just import it here and it'll either work or not? (better to ask for forgiveness)

jcm/physics/aerosol/jam/microphysics/mam4_jax.py
```diff
@@ -127,12 +129,73 @@ class Mam4JaxMicrophysics(ModalMicrophysicsTerm):
...
+        60-tracer spectral transport ~halve their traffic). ``None`` (default)
+        reads the ``MAM4_JAX_ENABLE_X64`` env var (default ``"1"`` → float64,
+        the safe default); ``True`` / ``False`` override it. Applied here, at
+        construction, so the dycore state built afterwards inherits it.
+        """
         if spec is not None:
             self.spec = spec
         _, _, _, data = _core()
 
+        if condensation_backend not in ("substep", "astem"):
```
duncanwp: Again, drop these protections - let the library worry about it's backends and raise an error if it wants

Issue #521, opened 2026-06-14
T63L47 jam aerosol NaN: modal mass/number inconsistency from spectral ringing of CEDS emissions


duncanwp:
## Summary

The T63L47 `echam-jam` + prescribed CEDS anthropogenic emissions stack NaNs the
aerosol tracers (and, downstream via the aerosol–radiation coupling, the whole
state) after ~1.5–2 simulated days. The same configuration is **finite at T42**.
Root cause is **modal mass/number inconsistency produced by spectral ringing of
the sharp emission sources**, which drives the diagnosed modal particle size far
outside its `dgnumlo`/`dgnumhi` bounds and NaNs the MAM4 size/condensation code.

This is a transport/microphysics-robustness problem, *not* a bug in the
emissions pipeline (#498), which is merged and correct.

## Evidence

5-day GPU run, T63L47, `physics=echam-jam`, `grid=echam_t63_l47_hybrid`, real
2014 CEDS+BB emissions (job 4569973). Day-0/day-1 snapshots are finite; day-2 is
all-NaN. Inspecting the last finite snapshot (day 1):

- Aerosol mass grows ~2×/day — smooth source accumulation, **not** an
  exponential transport blow-up.
- The dynamics tracer state carries ~10–12 % **negative** points (Gibbs ringing
  of the sharp industrial/shipping sources).
- Accumulation-mode **mass-per-particle** (Σ mode mass / mode number) spans
  **−9.4e-16 … +4.7e-13 kg**, vs. a healthy MAM4 accumulation range of
  ~1e-19…1e-17 kg. The *median* column is healthy (1.5e-18 kg); it is the ~1 %
  of ringing-corrupted columns that blow up:
  - 2770 / 866304 columns: number > 0 but mass ≤ 0 → diagnosed size → 0
  - 6352 / 866304 columns: mass > 0 but number ≤ 0 → diagnosed size → ∞

The resolution dependence (finite T42 → NaN T63) is consistent with ringing
amplitude growing with spectral truncation.

## Why the obvious guards don't fix it

- A per-tracer **positivity clamp** (now in place at the dynamics→physics
  boundary, mass-conserving — see PR linked below) removes negatives but clips
  each mode's mass and number *independently*, so it cannot restore their
  **ratio** (∝ particle size). It is a guard, not a cure.
- A **correction-tendency feedback** (clip in grid space, feed `(clipped −
  original)/dt` back as a spectral tendency) was tried and **reverted**: the
[...]

PR #522, opened 2026-06-14
fix(aerosol): smoke-test fixes for prescribed-emissions runs (#498)

duncanwp:
Four independent fixes surfaced while running the T63L47 `echam-jam` +
prescribed CEDS emissions stack on GPU (NCAR Casper A100). All four are
needed to get the science config to *run*; the residual aerosol instability
they expose is tracked separately in #521.

## Fixes

- **Calendar-aware emissions time axis** (`forcing.py`). CEDS/biomass files
  carry a cftime (365_day/noleap) calendar; the old pandas-only path raised on
  the non-standard calendar. Object-dtype (cftime) time coords now go through
  `cftime.date2num`; datetime64 keeps the fast pandas path.

- **`_jam_state` per-mode output** (`model.py` + `utils.py`). The per-mode JAM
  aerosol state (`_jam_state.*`, shaped `(mode, level, lon, lat)`) now
  serializes with a named `mode` coordinate sourced from the aerosol population
  spec, plus matching `(mode, level, …)` shape→dims entries in
  `data_to_xarray`. Previously the unmapped mode axis failed the shape lookup.

- **Forcing dtype cast** (`model.py`). `ForcingData` is built before the
  MAM4-JAX core enables `jax_enable_x64`, so its float32 arrays clashed with the
  float64 scan carry ("carry input/output types differ"). The selected forcing
  is now cast to the gridpoint-state float precision each step.

- **Mass-conserving positivity guard** (`model.py`). Spectral transport of the
  sharp, near-zero CEDS sources leaves Gibbs-ringing negatives; a negative mass
  or number is unphysical for *any* term (MAM4 size/condensation, ARG
  activation, radiation aerosol optics), so they are floored at the
  dynamics→physics boundary (clip + per-column rescale to conserve column mass).
  Documented as a guard, **not** a cure — it cannot restore modal mass/number
  consistency, which is the deeper T63+ instability tracked in #521.

## Testing

- `JAX_PLATFORMS=cpu pytest jcm/forcing_test.py` — 45 passed
- `JAX_PLATFORMS=cpu pytest jcm/model_test.py jcm/utils_test.py -m "not slow"` —
  65 passed, 1 skipped
- `ruff check` clean on all three files
- End-to-end: full T63L47 echam-jam + emissions stack runs 5 sim-days on GPU;
  dynamics stable, `_jam_state` and aerosol tracers serialize with the `mode`
  coordinate. (Aerosol tracers themselves still NaN at T63 — see #521.)

## Related

- Tracks residual instability: #521
- Upstream x64 `lax.cond` dtype fix: climate-analytics-lab/jax-rrtmgp#12

### DIFF
--- a/jcm/model.py
+++ b/jcm/model.py
@@ -26,6 +26,7 @@ import pandas as pd
 from functools import partial
 import logging
 
+from jcm.constants import physical_constants
 from jcm.date import DateData, parse_duration_days
 from jcm.forcing import ForcingData, default_forcing
 from jcm.physics_interface import (
@@ -39,6 +40,32 @@ from jcm.dycore.dinosaur.dycore import DinosaurDycore
 
 
 logger = logging.getLogger(__name__)
+
+
+def _mass_conserving_positivity(q, m):
+    """Clip ``q`` to non-negative while conserving its column-integrated mass.
+
+    Spectral transport of a sharp source (e.g. CEDS industrial/shipping
+    emissions) leaves Gibbs-ringing negatives in the tracer field; fed to the
+    physics (MAM4 size/condensation, ARG activation, …) a negative aerosol
+    mass/number produces NaNs. A naïve floor at zero would *add* mass, so instead
+    clip the negatives and rescale the surviving positive part of each column so
+    the column mass is unchanged ("hole-filling"):
+
+        q' = max(0, q) · max(M, 0) / M_clip,
+        M = Σ_k m_k q_k,   M_clip = Σ_k m_k max(0, q_k),
+
+    with ``m_k`` the per-layer air mass (∝ Δp). Non-negative by construction and
+    column-mass-conserving when ``M > 0``; a column whose mass is spuriously
+    net-negative is zeroed (the only, unavoidable, non-conservation). ``q`` and
+    ``m`` are ``(nlev, *spatial)``; the reduction is over the leading level axis.
+    """
+    q_clip = jnp.maximum(0.0, q)
+    col_mass = jnp.sum(m * q, axis=0)
+    col_mass_clip = jnp.sum(m * q_clip, axis=0)
+    scale = jnp.where(col_mass_clip > 0.0,
+                      jnp.maximum(col_mass, 0.0) / col_mass_clip, 0.0)
+    return q_clip * scale[jnp.newaxis, ...]
 
 
 class ModelPredictions:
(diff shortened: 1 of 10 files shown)


duncanwp:
@codex review again please

jcm/model.py
```diff
@@ -435,6 +524,36 @@ def step(state, physics_state):
...
             forcing_now = forcing.select(date, calendar=self.calendar)
             physics_state_grid = self.dycore.to_physics_state(state)
+            # Keep forcing in the model's working float precision. ``ForcingData``
+            # is often built before the MAM4-JAX core enables ``jax_enable_x64``
+            # (so its arrays are float32), whereas the x64 model state is float64.
+            # The diagnostics that read forcing (radiation solar geometry, surface
+            # fields) would then come out float32 against a float64 scan carry —
+            # a "carry input/output types differ" error. Cast float leaves of the
+            # selected forcing to the gridpoint-state dtype to stay consistent.
+            _fdtype = physics_state_grid.temperature.dtype
```
duncanwp: I don't like this happening inside the step. This should happen at the top when forcing is first setup in the model (before time slixing)

jcm/model.py
```diff
@@ -421,6 +475,41 @@ def _prepare_initial_dycore_state(self, physics_state: PhysicsState = None,
...
+        """Apply the mass-conserving positivity filter to all gridpoint tracers.
+
+        No-op when there are no tracers (e.g. dry dynamical-core configs). The
+        per-layer air-mass weight ``Δp`` is reconstructed from the vertical
+        coordinate (hybrid ``a``/``b`` or pure sigma) and the column surface
+        pressure, exactly as :class:`MoistAirColumnState` does, so the rescale
+        conserves the same column mass the model integrates.
+        """
+        tracers = physics_state_grid.tracers
+        if not tracers:
```
duncanwp: I count twenty lines of code here just to get dp. This is stored somewhere in the model, or should be easily derived from it. This should be dycore agnostic too - so why are we importing dinosaur? Whatever the coords are should provide their own half levels. But like I say  I think we should already have dp somewhere anyway

jcm/model.py
```diff
@@ -41,6 +42,32 @@
...
+    """
+    q_clip = jnp.maximum(0.0, q)
+    col_mass = jnp.sum(m * q, axis=0)
+    col_mass_clip = jnp.sum(m * q_clip, axis=0)
+    scale = jnp.where(col_mass_clip > 0.0,
+                      jnp.maximum(col_mass, 0.0) / col_mass_clip, 0.0)
+    return q_clip * scale[jnp.newaxis, ...]
+
+
 class ModelPredictions:
```
duncanwp: This is getting pretty big now - let's move it into its own module

jcm/model.py
```diff
@@ -41,6 +42,32 @@
 logger = logging.getLogger(__name__)
 
 
+def _mass_conserving_positivity(q, m):
```
duncanwp: There are a number of filters in the model now - should these have their own module?

jcm/model.py
```diff
@@ -435,6 +524,36 @@ def step(state, physics_state):
...
+            #
+            # This is a guard, not a cure for the deeper instability: it cannot
+            # restore *modal mass/number consistency*. Independent ringing of a
+            # mode's mass and number fields drives their ratio (∝ particle size)
+            # orders of magnitude out of the dgnumlo/dgnumhi bounds, which NaNs
+            # the microphysics at T63+ even with non-negative inputs. The root-
+            # cause fix is dynamical (tracer hyperdiffusion / a sign- and ratio-
+            # preserving tracer transport) and/or a calcsize robustness clamp in
+            # mam4-jax — tracked separately; see issue.
+            physics_state_grid = self._filter_tracer_positivity(physics_state_grid)
```
duncanwp: GIven different dycores won't have this problem I think this filter should be treated like the others and specified at Model construction so users can implement them or not

PR #536, opened 2026-06-27
fix(clouds): suppress spurious stratospheric Sundqvist cloud (ECHAM jks cutoff)

duncanwp:
## Summary

The Sundqvist diagnostic cloud-fraction closure was evaluated at every level, filling the cold model stratosphere with **spurious cloud**. At ~180 K the ice saturation specific humidity collapses to ~1.5×10⁻³ g/kg, so the grid-mean `q/qsat` reaches **~80× (8000 % "RH")** and cloud fraction saturates near **0.88 between ~17 and 110 hPa**.

Found while building the annual-mean cloud-cover map for the T63L47 ECHAM-RRTMGP-2M climatology: naïve total cloud cover came out at **96 %** (vs a physical ~65 %), traced to this stratospheric layer.

## ECHAM reference

`mo_cover.f90` zeros cloud cover for every level above `jks`:
```fortran
DO jk = 1,jks-1
   paclc(jl,jk) = 0.0_wp     ! no cloud in the stratosphere
DO jk = jks,klev             ! cloud only from jks down to the surface
```

## Fix

Mirror ECHAM's `jks` with a portable pressure threshold: new `CloudParameters.cloud_top_pressure_pa` (default **100 hPa**) forces cloud fraction to zero wherever the full-level pressure is below it. Set to `0` to disable.

The deeper cause is the model's cold + moist stratosphere bias (ECHAM's stratosphere is warmer via ozone SW heating, so `qsat` doesn't collapse) — the cutoff is ECHAM's own guard and a clean diagnostic fix.

## Scope / what this does NOT fix

This is the cloud-**cover** fix. It does **not** change the separately-found **zero cloud radiative effect** (all-sky == clear-sky, albedo 0.115): the stratospheric cloud carries no condensate so it is radiatively inert. The CRE bug is a different root cause (`init=jw` wipes the cloud-water tracers before the run) and is tracked separately.

## Validation
- New `test_stratospheric_cloud_cutoff` (cloud forms in troposphere, zero above the cutoff; disabling restores it).
- 25 Sundqvist tests pass; ruff clean.
- ECHAM T21L16 fingerprint **unchanged** (the cutoff doesn't bite in the 1-day reference — the cold-bias supersaturation builds over many days).

### DIFF
--- a/jcm/physics/clouds/sundqvist.py
+++ b/jcm/physics/clouds/sundqvist.py
@@ -51,13 +51,25 @@ class CloudParameters:
     t_mix_min: float     # Lower bound of mixed phase (K)
     t_mix_max: float     # Upper bound of mixed phase (K)
 
+    # Cloud-top pressure cutoff (ECHAM ``jks`` analogue). ECHAM
+    # ``mo_cover.f90`` zeros cloud cover for all levels above ``jks``
+    # (``DO jk=1,jks-1: paclc=0``) so no cloud forms in the stratosphere.
+    # We express that level cutoff as a pressure threshold (portable across
+    # hybrid grids): cloud fraction is forced to zero wherever the full-level
+    # pressure is below ``cloud_top_pressure_pa``. Without it the RH-based
+    # Sundqvist closure fills the cold (here ~180 K, qsat→0) stratosphere with
+    # spurious cloud — the q/qsat ratio reaches ~80× — saturating cf there.
+    # Set to 0 to disable the cutoff.
+    cloud_top_pressure_pa: float
+
     @classmethod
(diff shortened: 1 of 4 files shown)


duncanwp:
Addressed the Codex P2: `shallow_cloud_scheme` now gates the condensation/evaporation tendencies (and the within-step condensate update) to zero above `cloud_top_pressure_pa`, so the standalone scheme no longer creates hidden stratospheric cloud above the masked levels. Added `test_standalone_scheme_gates_condensation_above_cutoff`. The composable `SundqvistCloudFraction` term was already unaffected (it emits zero qc/qi tendency; condensation is the microphysics' job).

PR #539, opened 2026-06-29
fix(clouds): 2M stratiform cloud via grid-mean Sundqvist condensation + sequential convection→cloud coupling

duncanwp:
## Summary

Gives the 2-moment (Lohmann) cloud scheme **radiatively-active stratiform cloud**, which it previously lacked, by adding the grid-mean implicit-Newton Sundqvist condensation/evaporation the 1M column-sweep already uses, fed by a **sequential convection→cloud coupling** so the cloud step condenses against the post-convective (moistened/cooled) column.

This is the remaining un-PR'd half of the moist-stability/CRE work; the init fixes landed on #536 and the RRTMGP in-cloud cap on #538. Validated together in stable 10-day moist T63L47 runs (rh=0.5) across all four configs (grey/rrtmgp × 1m/2m).

## Commits

1. **`feat(clouds): sequential convection→cloud coupling via thermo_run`** — a running thermodynamic state threaded through the per-step diagnostics: `moist_air_state` seeds `thermo_run = {temperature, specific_humidity}`, the Tiedtke term advances it by its own tendency, and the 2M scheme reads it (falling back to the step-start state when unseeded, preserving additive-splitting behaviour). Makes the convection→cloud handoff sequential (operator splitting), as in ECHAM. (Formerly the `wip` commit; the coupling is functional and validated — label dropped.)
2. **`fix(clouds): 2M grid-mean Sundqvist condensation + disable harmful clear-cell evap`** — condenses vapour→qc/qi where the post-convection grid box is supersaturated and evaporates qc/qi where sub-saturated, via the warming-feedback-damped Newton step. Without it the 2M formed almost no stratiform cloud (LWP ~50× smaller than 1M) and detrained condensate accumulated unbounded or was destroyed by the clear-cell-evaporation bolt-on (Part A, now disabled by default behind `JCM_CLEAR_SKY_EVAP`).
3. **`fix(clouds): ECHAM cdnc_min floor on 2M warm-microphysics droplet number`** — robustness floor on the droplet number.

## Validation

- `lohmann_2m_test.py` + `sundqvist_test.py`: 71 passed. `tiedtke_nordeng/`: 116 passed.
- 10-day moist T63L47 spin-ups stable across all four configs: grey+2M decorrelation 99%→20%, CRE_SW ≈ +12.5 (was +0/exploding); rrtmgp+2M stable with active CRE.

## Note for review

The disabled Part A clear-cell-evaporation code remains in `lohmann_2m.py` gated behind `JCM_CLEAR_SKY_EVAP` (off) as a documented A/B investigation knob — happy to strip it entirely if you'd prefer it gone rather than gated.

### DIFF
--- a/jcm/physics/clouds/lohmann_2m.py
+++ b/jcm/physics/clouds/lohmann_2m.py
@@ -58,9 +58,10 @@ from .lohmann_2m_params import (
     fact_coll_eff, fact_tke
 )
 
-from .cloud_utils import (eff_ice_crystal_radius, minimum_CDNC,
-                          consistency_number_to_mass, gridbox_frac_falling_hydrometeor, threshold_vert_vel, 
-                          breadth_factor
+from .cloud_utils import (
+    eff_ice_crystal_radius, minimum_CDNC,
+    consistency_number_to_mass, gridbox_frac_falling_hydrometeor,
+    threshold_vert_vel, breadth_factor
 )
 
 # @tree_math.struct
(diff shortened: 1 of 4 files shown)


duncanwp:
Rebased on `dev` (now includes #536 + #538) and addressed both review comments. Restructured into two clean commits so the review changes are folded in (no introduce-then-remove churn):

- `feat(clouds): sequential convection→cloud coupling via thermo_run`
- `feat(clouds): 2M grid-mean Sundqvist stratiform condensation + shared cdnc minimum`

**1. cdnc minimum → calibratable & consistent.** Agreed it shouldn't be a separate constant/parameter. The warm-rain droplet floor now reuses the scheme's existing `minimum_CDNC` — the *same* minimum the in-cloud path already uses (`lohmann_2m.py:2676`) — instead of the `_CDNC_MIN_PER_M3` module constant. So there's now a single calibratable cdnc minimum (`cdnc_min_fixed` / `ldyn_cdnc_min` / `cdnc_min_upper`/`_lower`) used everywhere. I set the `cdnc_min_fixed` default to **40 cm⁻³** (ECHAM's warm-microphysics value, and what the stable 10-day runs used) so the consolidation preserves the cloud-water reservoir the floor protects. Note this also raises the in-cloud minimum from 10→40 cm⁻³, since they're now one quantity.

**2. clear-sky-evaporation switch → removed.** You're right that we always want sub-saturated evaporation — and we always *do*: the grid-mean Sundqvist condensation/evaporation step in the 2M `__call__` evaporates qc/qi wherever the grid box is sub-saturated, every step. The `JCM_CLEAR_SKY_EVAP`-gated code was a *separate, earlier* in-scheme clear-cell evaporation that double-counted with that step and destroyed ~99% of the cloud water in an A/B test, which is why it was gated off. Rather than ship a flag, I removed that redundant path entirely (computation + application + the env switch + the now-unused `import os`).

Validation: `lohmann_2m` + `sundqvist` unit tests pass (73), ruff clean, lint green; build running. The 4-config 10-day moist spin-ups that exercised this remain stable.

PR #557, opened 2026-07-04
feat(forcing): file hooks for JAM natural emissions + prescribed oxidants

duncanwp:
Wires the three missing loader hooks so Hydra-launched `echam-jam` runs stop being silently zero-emission: `forcing.dms_file` (HAMMOZ seawater DMS, nanomol/l → kg/m³), `forcing.dust_file` (pot_source erodibility, clipped [0,1]), `forcing.oxidants_file` (MACC OH/O₃/H₂O₂/NO₃ VMRs on the model's hybrid levels, coefficient-checked; `PrescribedOxidants` prefers them over the analytic interim). Ozone-loader conventions throughout: null no-op, grid validation + lat flip, WRAP_YEAR monthly TimeSeries, loud errors on mismatch; the HAMMOZ `_FillValue=0`→NaN land cells are mapped to 0. GPU smoke on T63L47 with the real HAMMOZ + regridded CEDS files: DMS ≈29 Tg/yr (climatological ballpark), dust active at ~21 Tg/yr with default knobs. 18 new tests; jam+forcing+runners suites 296 green.

Unblocks the 2M+JAM climatological run (campaign deliverable #3).


duncanwp:
Addressed in 60682c4. `read_dust_source` now detects a static `(lat, lon)` map (no `time` dim) and returns a bare `(lon, lat)` array — `ForcingData.select` passes non-`TimeSeries` leaves through untouched, so `DustEmissions` sees the same 2-D field every step. The monthly `(time, lat, lon)` climatology path (e.g. the `dust_potential_sources_T63.nc` we run with, which does carry a 12-month axis) is unchanged. Clip / orient / missing-marker (`-1`→0) semantics are identical on both paths; regression pin `test_dust_reader_accepts_static_lat_lon_map` added.

PR #559, opened 2026-07-05
fix(#558): harden ECHAM physics against multi-step reverse-mode gradient NaNs

### LINKED ISSUE #558: ECHAM: reverse-mode gradient NaNs beyond one model step (forward and finite differences both finite)
Found while building JEM-Cal's Phase 2a gradient gate against **dev** (bce05cc), where multi-step reverse-mode through the ECHAM stack returns NaN while the forward pass and a central finite difference are both finite. Minimal characterization below; this currently blocks gradient calibration at ECHAM complexity.

**Configuration**: T21L47 (`get_coords(get_echam_levels(47), spectral_truncation=21)`), `echam_physics(radiation_scheme="grey")` at scheme defaults, balanced isothermal start (`inject_balanced_isothermal_profile`), `default_forcing` aquaplanet SSTs, `time_step=15.0` (minutes), `output_averages=True`, x64, CPU, jax 0.10.2. Objective: `mean(preds.dynamics.temperature)`, differentiated with respect to `RadiationParameters.solar_constant`.

| steps | checkpoint_terms | d meanT / d S0 |
|---|---|---|
| 1 | True | 2.6053230612888e-06 |
| 1 | False | 2.6053230612888e-06 (bit-identical) |
| 3 | True | **nan** |
| 3 | False | **nan** |
| 3 | central FD (eps = 5 W/m2) | 5.2104638712e-06 |

Forward is finite in every configuration (mean T 287.91 K after 5 steps). The same holds for a convection parameter (`entrpen`): AD gives exactly 0.0 over a short dry start, which is physically right (no convection yet), so the defect is specific to the *chained* backward, not to one scheme's VJP in isolation.

Reading of the evidence: the NaN enters the backward pass only when more than one model step is chained, and independently of term-level remat. That points at either the cross-step physics carry (radiation cache, prior-step TKE) or the classic unselected-branch cotangent poisoning (`where`/`sqrt`-at-zero guards whose NaN cotangent only propagates once a later step consumes the output), rather than at `jax.checkpoint` or any single term.

Reproduction (venv with dev-editable jcm, `JAX_PLATFORMS=cpu`):

```python
import dataclasses
import jax, jax.numpy as jnp
jax.config.update("jax_enable_x64", True)

from jcm.forcing import default_forcing
from jcm.model import Model
from jcm.physics.echam.echam_levels import get_echam_levels
from jcm.physics.echam.echam_terms import echam_physics
from jcm.physics.radiation.radiation_types import RadiationParameters
from jcm.runners import inject_balanced_isothermal_profile
from jcm.utils import get_coords

coords = get_coords(get_echam_levels(47), spectral_truncation=21)
forcing = default_forcing(coords.horizontal)
rad0 = RadiationParameters.default()

def objective(s0, steps=3):
    rad = dataclasses.replace(rad0, solar_constant=s0)
    physics = echam_physics(radiation=rad, radiation_scheme="grey", checkpoint_terms=False)
    model = Model(coords=coords, physics=physics, time_step=15.0)
    inject_balanced_isothermal_profile(model)
    days = 15.0 * steps / 1440.0
    preds = model.resume(forcing=forcing, save_interval=days, total_time=days, output_averages=True)
    return jnp.mean(preds.dynamics.temperature)

print(jax.grad(objective)(jnp.asarray(1361.0)))  # nan; finite with steps=1
```

JEM-Cal's ECHAM FD-vs-AD test (`examples/test_jcm_echam.py::test_ad_matches_finite_difference_through_echam`) is `xfail(strict=True)` on exactly this configuration, so it will flag (XPASS) as soon as a fix lands.

duncanwp:
Closes #558.

## What this fixes
Multi-step reverse-mode differentiation through the ECHAM column physics returned **NaN** while the forward pass and a central finite difference were both finite. All four configurations calibration relies on now backprop cleanly through two model steps:

| config | 2-step d(meanT)/d(S0) |
|---|---|
| 1M / MACv2-SP | 5.2108e-6 |
| 2M / MACv2-SP | 5.2108e-6 |
| 1M + JAM | 5.2108e-6 |
| 2M + JAM | 5.2108e-6 |

matching the issue's central FD (5.2105e-6).

## Root cause (not the get_empty_data probe)
The issue hypothesised the structural probe in get_empty_data. I implemented that fix (jax.eval_shape) and it did **not** help — disproving it. The real cause is a *class* of degenerate-state cotangent poisons: sqrt(0), x**p at x==0 for fractional p, and a/0, sitting in where-masked branches. The forward masks the bad value, but the masked branch's derivative is inf and 0*inf = nan poisons the gradient once a second step chains through it. Triggered by states that are entirely normal in practice — an aquaplanet (zero sub-grid orography → SSO denominators are 0), a balanced isothermal start (zero wind → sqrt(u²+v²) has infinite derivative), and clear / ice-free cells (cloud fraction or condensate 0 under a fractional power).

## Terms hardened (forward-preserving guards)
SSO Lott-Miller (orography denominators + zero-wind sqrts), 1M microphysics, TTE-TKE vdiff + surface layer, ECHAM surface (turbulent fluxes / surface physics / ocean), Tiedtke convection (updraft + precip evap), lohmann_2m 2-moment microphysics, cloud_utils.eff_ice_crystal_radius, SPA activation, and JAM sea-salt/DMS emissions.

## Localization method
Manual N-step loop (bypassing the outer lax.scan) + lax.scan monkeypatched to a Python unroll, so jax.debug_nans names the exact op inside physics scans rather than pointing at the scan primitive.

## Validation
- New jcm/physics/echam/gradient_finiteness_test.py (slow): finite gradient for all four configs + FD match on 2M+JAM.
- Forward physics unchanged: **652** fast tests across clouds/convection/vertical_diffusion/surface/gravity_waves/aerosol pass; ruff clean.

### DIFF
--- a/jcm/physics/aerosol/jam/emissions/dms.py
+++ b/jcm/physics/aerosol/jam/emissions/dms.py
@@ -89,7 +89,7 @@ class DmsEmissions(PhysicsTerm):
         dz = diagnostics["layer_thickness"]
         nlev, ncols = state.temperature.shape
 
-        u10 = jnp.sqrt(state.u_wind[-1] ** 2 + state.v_wind[-1] ** 2)
+        u10 = jnp.sqrt(jnp.maximum(state.u_wind[-1] ** 2 + state.v_wind[-1] ** 2, 1.0e-30))
         sst = self._forcing_field(
             forcing, "sea_surface_temperature", ncols, 288.0
         )
--- a/jcm/physics/aerosol/jam/emissions/seasalt.py
+++ b/jcm/physics/aerosol/jam/emissions/seasalt.py
@@ -154,7 +154,7 @@ class SeaSaltEmissions(PhysicsTerm):
         dz = diagnostics["layer_thickness"]
         nlev, ncols = state.temperature.shape
 
-        u10 = jnp.sqrt(state.u_wind[-1] ** 2 + state.v_wind[-1] ** 2)
+        u10 = jnp.sqrt(jnp.maximum(state.u_wind[-1] ** 2 + state.v_wind[-1] ** 2, 1.0e-30))
         seafrac = self._open_water_fraction(forcing, terrain, ncols)
         wind = p.scale * u10 ** p.wind_exponent * seafrac   # (ncols,)
 
--- a/jcm/physics/aerosol/spa.py
+++ b/jcm/physics/aerosol/spa.py
@@ -58,7 +58,14 @@ def spa_activated_cdnc(Nccn: jnp.ndarray, cloud_fraction: jnp.ndarray,
     # result is then already in m^-3 (no further unit conversion).
     nccn_m3 = jnp.maximum(Nccn, 0.0) * _CM3_TO_M3
     arg = jnp.maximum(nccn_m3 * cloud_fraction, 0.0)
-    nc_min_m3 = prefactor * arg ** exponent
+    # Double-where guard: ``arg`` is 0 in a clear cell (cloud_fraction == 0,
+    # the common case) and the fractional ``arg ** exponent`` has an infinite
+    # derivative at 0, poisoning the reverse pass while the forward is 0
+    # (issue #558). Keep the forward exactly 0 there; differentiate the power
+    # only on a strictly-positive floored argument.
+    nc_min_m3 = prefactor * jnp.where(
+        arg > 0.0, jnp.maximum(arg, 1.0e-30) ** exponent, 0.0,
+    )
     # Physical bound: the (grid-mean) droplet floor cannot exceed the cloudy CCN
     # that actually entered the activation fit — i.e. ``arg = Nccn·cloud_fraction``,
     # NOT the full-column ``Nccn``. The power-law only overshoots at very low
--- a/jcm/physics/clouds/cloud_utils.py
+++ b/jcm/physics/clouds/cloud_utils.py
@@ -35,9 +35,13 @@ def eff_ice_crystal_radius(
     # #558). Keep the forward exactly 0 where there is no ice; differentiate
     # the power only on the strictly-positive floored base.
     base = pxice / jnp.maximum(params.fact_PK * jnp.maximum(picnc, eps), eps)
+    # The ``where`` already returns 0 for ice-free cells; the base floor only
+    # needs to keep the differentiated branch strictly positive, so it must be
+    # NEGLIGIBLE (not ``params.eps`` ≈ 1e-7, which would inflate the effective
+    # radius of small-but-nonzero ice and perturb the physics).
     return 0.5e4 * jnp.where(
         pxice > 0.0,
-        jnp.maximum(base, eps) ** (1.0 / params.pow_PK),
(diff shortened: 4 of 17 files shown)


duncanwp:
Fixed in the latest commit. The module-level `jax_enable_x64` flip was leaking into sibling tests during collection (the `build (3.11)` failure was `speedy_humidity`/`speedy_shortwave` float32 assertions breaking under the leaked x64). Replaced it with an `autouse` fixture that enables x64 per-test and restores the prior value in `finally` — same pattern as the MAM4-JAX tests you referenced. Verified those two suites now pass with this module collected under `-m "not slow"`.

duncanwp:
**Correction on composed_budget — it was my regression, now fixed (40e5e7f).**

My earlier 'pre-existing flake' call was wrong: it was masked by my local jax being 0.9.2, below the repo requirement (`jax>=0.10.0`; CI uses 0.10.2). On jax 0.10.2, composed_budget PASSES on clean dev but FAILED on this branch — so my guards were the cause.

Root cause: two #558 guards floored a legitimately-small value with a too-large floor, changing the forward physics:
- `echam_1m` ice sedimentation: `maximum(rho*qi, config.epsilon)**0.16` — `config.epsilon` (1e-12) inflates the fall speed of tiny ice by orders of magnitude.
- `cloud_utils.eff_ice_crystal_radius`: `maximum(base, params.eps)` — `params.eps` (float32 machine eps ≈1e-7) inflates small-ice effective radius.

Both sit inside a `where` that already returns 0 for the ice-free case, so the inner floor only needs to keep the differentiated branch strictly positive — it must be negligible (1e-30), not epsilon. Fixed.

Verified on jax 0.10.2: composed_budget passes; 113 cloud fast tests pass; 1M and 2M+JAM gradients still finite and FD-matched (5.2108e-6).

duncanwp:
@codex review

duncanwp:
Thanks — both addressed in cdb6cfa:

**"Seed the physics carry before direct stepping" (valid, fixed).** Confirmed the mechanism: `inject_balanced_isothermal_profile` only sets `_final_dycore_state`, so `_final_physics_state` was still `None` from the constructor, and stepping with `None` synthesises a zero carry — the TTE-TKE term fell back to `VerticalDiffusionData.zeros(...)` (TKE=0) instead of its seeded 0.01 floor, a state production never runs. The helper now seeds the carry via `_build_initial_physics_carry()`, exactly as `Model.run`/`resume` do when it's `None`. Both configs stay finite and now agree at 5.2104e-6 (matching the float64 central FD 5.2105e-6), replacing the non-production zero-TKE value. The #558 triggers (SSO zero-orography denominators, zero-wind `sqrt(u²+v²)`, clear/ice-free fractional powers) are all carry-independent, so guard coverage is unchanged.

**"Restore x64 after the slow gradient tests" (stale).** That flagged the earlier revision that flipped the process-global `jax_enable_x64` at import time. The current module never touches that flag — the `0*inf=nan` poison is dtype-agnostic and both non-JAM configs are float32-finite, so the test runs in-process at the session default. No import-time state leaks to sibling tests.

PR #560, opened 2026-07-07
fix(#558): safe-denominator guard in radiation optical-property combination

duncanwp:
## Summary

Fixes the residual #558 cloud-parameter backward NaN reported on the issue: differentiating a **cloud** parameter (`crt`) through a production-length grey-radiation rollout NaN'd (finite at ≤8 steps, NaN at ≥12) even after the earlier differentiability-floor hardening. The forward is finite at every step — a pure reverse-mode artifact.

## Root cause

Two optical-property combinations weighted single-scatter albedo / asymmetry with a **divide-by-the-`where`-condition**:

```python
combined_ssa = jnp.where(total_tau > 0, total_scattering / total_tau, 0.0)
```

`jnp.where` evaluates the true branch in *every* cell, so a clear, aerosol-free layer (`total_tau == 0`, or `total_scattering == 0` for the asymmetry weight — the common case) computes `x/0`: finite-masked in the forward, but derivative `∝ 1/0 = inf`, and `where`'s VJP forms `0 (mask) · inf = nan`. This poisons the gradient of any upstream cloud parameter (`crt` → cloud fraction → whether a layer's τ hits 0), and only accumulates enough sensitivity to surface past ~10 steps — exactly the reported characterization.

Four sites, all in the grey-radiation path:
- `grey_two_stream/radiation_scheme.py` · `combine_optical_properties` (ssa, g)
- `cloud_optics.py` · `cloud_sw_optics` (ssa, g)

## Fix

The safe-denominator double-`where`: floor the denominator to 1 in the masked cells so the differentiated division never sees a zero; the outer `where` still selects the clear-sky value, so the forward is **bit-identical**.

## Validation

- `d(meanT)/d(crt)` through 12- and 24-step rollouts is now **finite** (x64: `-2.0e-81` / `1.2e-82`; was NaN at 12).
- Forward unchanged: **70** grey two-stream + cloud-optics tests pass.
- Regression test `test_cloud_parameter_gradient_finite_through_long_rollout` added. The gradient is ~1e-81 with no float32 representation (intermediate cotangents overflow float32 regardless of the fixed poison), so it runs float64 in an **isolated subprocess** (no `jax_enable_x64` leak into the in-process float32 tests, no xdist-cache corruption) via `lax.scan` (a 12-step unroll of the full column exhausts CI's LLVM section memory).

This resolves the x64 backward used by JEM-Cal calibration (its `test_cloud_parameter_gradient_finite_through_long_rollout` gate should XPASS).

## Out of scope

A separate **float32-only** degeneracy in this same long cloud rollout remains (dynamic range, not a #558 poison — the gradient is unrepresentable in float32). Calibration runs in float64, so it's unaffected; flagged for follow-up.

Closes part of #558.

### DIFF
--- a/jcm/physics/echam/gradient_finiteness_test.py
+++ b/jcm/physics/echam/gradient_finiteness_test.py
@@ -121,3 +121,104 @@ def test_two_step_gradient_is_finite_and_correct(cloud_scheme, aerosol_module):
         f"{cloud_scheme}/{aerosol_module}: gradient {float(grad):.4e} is far "
         f"from the validated {_EXPECTED_GRAD:.4e}."
     )
+
+
+# --- Cloud-parameter gradient through a long rollout (#558 residual) ----------
+#
+# A distinct #558 poison surfaced only when differentiating a *cloud* parameter
+# (not the solar constant) through a longer rollout: finite at <=8 steps, NaN at
+# >=12. Root cause: ``combine_optical_properties`` (grey radiation) and
+# ``cloud_sw_optics`` (cloud_optics) combined single-scatter-albedo / asymmetry
+# with a bare ``jnp.where(tau>0, scattering/tau, 0)`` — the true branch divides
+# by the *same* quantity the mask tests, so a clear/aerosol-free layer computes
+# ``x/0`` whose derivative is ``inf`` and ``where``'s VJP forms ``0*inf = nan``.
+# Backward-only (forward finite), cloud-parameter-driven (crt -> cloud fraction
+# -> whether a layer's tau hits 0), rollout-dependent (a layer crosses into
+# clear ~10 steps in). Fixed with the safe-denominator double-``where`` in both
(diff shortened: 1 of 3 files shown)


duncanwp:
Good catch on the `sys.path` gotcha — but that subprocess test is already gone as of beb01ae (your comment is pinned to e3877c9). The x64 full-grid rollout it ran timed out on CI's 2-vCPU runner anyway, so I replaced it with two fast in-process function-level guards (`test_combine_optical_properties_gradient_finite_in_clear_layer`, `test_cloud_optics_gradient_finite_in_clear_layer`) that `jax.grad` the two fixed functions directly with a clear (zero-denominator) layer. No subprocess, no `__main__`, no x64 toggle, <10s each — and verified to have teeth (the pre-fix bare `where(x>0, .../x, 0)` NaNs on those inputs, the fix is finite).

PR #561, opened 2026-07-10
Differentiability of the RRTMGP + 2M configuration: r_eff NaN fix and backward-memory chunking

duncanwp:
Two fixes that, together with climate-analytics-lab/jax-rrtmgp#15 and the stacked climate-analytics-lab/jax-rrtmgp#16, make `echam_physics(radiation_scheme="rrtmgp", cloud_scheme="2m", aerosol_module="macv2sp")` reverse-mode differentiable at T63L47. Further commits will follow on this PR for the remaining physics NaNs (TTE-TKE vdiff under JW + real terrain; Tiedtke `entrpen` when convection is active).

### 1. `fix(2m)`: guard the liquid effective radius on liquid, not on temperature

A gradient through RRTMGP+2M returned NaN from the second rollout step onward while the forward stayed finite. `liquid_cloud_flag` is `temperature > tmelt` (a purely thermal test), so in every warm cloud-free cell the double-where guard on the cube root in the liquid effective radius put a **zero on the differentiated branch**, where `d(x^(1/3))/dx = inf`; the incoming cotangent is exactly zero there, giving `0 * inf = NaN`. Two steps because radiation runs before the microphysics in the term order, so it consumes the radii 2M wrote into the cloud carry on the *previous* step; at step 1 nothing consumes `r_eff` and its VJP is never evaluated.

Guarding on `liquid_cloud_flag AND cloud_liquid_in_cloud > 0` leaves the forward bit-identical (T21L47 meanT `287.9279052522753` before and after) and also removes a genuine forward NaN when spectral ringing drives the condensate negative.

Verified: AD vs central FD to 1.6e-4 at T21L47 and T63L47 (the FD truncation floor); in a cloudy state (JW rh=0.9) the cloud/microphysics parameters get nonzero FD-verified gradients (`clouds.crt` -1.80e-5 vs -1.79e-5, `ccraut` -1.11e-5 vs -1.12e-5).

### 2. `perf(rrtmgp)`: optional column chunking for the backward

The forward's per-column working set is small, but the reverse pass holds every column's per-g-point profiles live at once. Measured `jax.grad` peaks (2 steps, x64): **19.3 GiB (T21) / 77.4 GiB (T31) / 309.5 GiB (T63)** — the target resolution does not fit an 80 GB device, and XLA's rematerialization pass cannot reduce it. `JCM_RRTMGP_COL_CHUNKS=n` maps a `jax.checkpoint`-ed vmap over n column blocks: T63 drops to **23.75 GiB** at n=8 (T21: 4.96 GiB at n=4) for the cost of one extra radiation forward in the backward pass. Unset reproduces the original single vmap exactly. Gradient value unchanged (T21 chunked vs unchunked agree to float reassociation, ~1e-7 relative).

https://claude.ai/code/session_01Aqshaa5nDrxcDWgg1FYw8Z

### DIFF
--- a/jcm/physics/aerosol/macv2_sp_params.py
+++ b/jcm/physics/aerosol/macv2_sp_params.py
@@ -59,10 +59,21 @@ class AerosolParameters:
     # differentiable through `jax.grad` for sensitivity / calibration.
     spa_prefactor: jnp.ndarray    # default 2000.0 (Lin 2025, fit to E3SMv3)
     spa_exponent: jnp.ndarray     # default 0.55  (sublinear; observational band 0.3–0.8)
+    # Half-width [m^-3] of the smooth transition that replaces the hard
+    # ``min(Nc_min, arg)`` physical cap in ``spa_activated_cdnc``. The hard min
+    # makes the loss piecewise in ``spa_prefactor`` (the cap boundary sits in
+    # the middle of the cloudy-cell distribution, so cells flip branches as the
+    # prefactor moves and the exact local gradient can even have the wrong sign
+    # relative to the large-scale response — measured through a 2-step T21L47
+    # rollout). Default 1e6 m^-3 (1 cm^-3), a few percent of typical activated
+    # CDNC, so the forward changes by at most half that at the cap corner.
+    # Set to 0.0 to recover the hard cap exactly.
+    spa_cap_smoothing: jnp.ndarray
 
     @classmethod
     def default(cls, background_aod=0.02,
-                spa_prefactor=2000.0, spa_exponent=0.55) -> 'AerosolParameters':
+                spa_prefactor=2000.0, spa_exponent=0.55,
+                spa_cap_smoothing=1.0e6) -> 'AerosolParameters':
         """Create the MACv2-SP v1 reference parameters.
 
         Static plume geometry/optics transcribed verbatim from
@@ -163,11 +174,13 @@ class AerosolParameters:
             background_aod=jnp.array(background_aod),
             spa_prefactor=jnp.array(spa_prefactor),
             spa_exponent=jnp.array(spa_exponent),
+            spa_cap_smoothing=jnp.array(spa_cap_smoothing),
         )
 
     @classmethod
     def from_dataset(cls, ds, background_aod=0.02,
-                     spa_prefactor=2000.0, spa_exponent=0.55) -> 'AerosolParameters':
+                     spa_prefactor=2000.0, spa_exponent=0.55,
+                     spa_cap_smoothing=1.0e6) -> 'AerosolParameters':
         """Build parameters from an opened ``MACv2.0-SP_v1.nc`` dataset.
 
         Owns the (plume, feature) -> (feature, plume) transposes and the
@@ -200,6 +213,7 @@ class AerosolParameters:
             background_aod=jnp.array(background_aod),
             spa_prefactor=jnp.array(spa_prefactor),
             spa_exponent=jnp.array(spa_exponent),
+            spa_cap_smoothing=jnp.array(spa_cap_smoothing),
         )
 
     def isnan(self):
--- a/jcm/physics/aerosol/macv2_sp_test.py
+++ b/jcm/physics/aerosol/macv2_sp_test.py
@@ -381,7 +381,7 @@ class TestJAXCompatibility:
                 'beta_b', 'aod_fmbg', 'asy550', 'ssa550', 'angstrom',
                 'sig_lon_E', 'sig_lon_W', 'sig_lat_E', 'sig_lat_W', 'theta',
                 'ftr_weight', 'background_aod', 'spa_prefactor',
-                'spa_exponent')}, 'aod_spmx': aod_spmx})
+                'spa_exponent', 'spa_cap_smoothing')}, 'aod_spmx': aod_spmx})
             out = get_simple_aerosol(height, dz, oro, lats, lons, data, p,
                                      f, jnp.asarray([550.0]))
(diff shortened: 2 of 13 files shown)


duncanwp:
Pushed the promised follow-up fixes. The remaining physics NaNs turned out to be **one bug with three faces**:

**`fix(clouds)`: squared-underflow in the condensate-partition divisions.** Every convection-parameter gradient NaN'd when convection was active, and `d(mean T)/d(solar_constant)` NaN'd at T63L47 under JW + real terrain (which had looked like a vertical-diffusion defect from term-level traces — the NaN back-propagates through the shared post-physics thermo view that vdiff and convection both feed). The creator is the liquid/ice partition in the saturation adjustment: the divisions are double-where guarded, but with `total_cloud > 0`, and spectral-ringing tails (qi ~ 1e-287 in real JW columns) land on the *selected* branch where the division VJP computes `-g·x/(y·y)` and `y²` underflows to exactly 0 → `0/0 = NaN` with a finite forward. Measured bad window on captured model columns: `total_cloud ∈ [~1e-300, ~1e-154]` — 1393 of 2048 columns. Guard raised to the `d_epsilon` differentiability floor; twin site fixed in `sundqvist.condensation_evaporation`.

Verified in the live model:
| case | before | after |
|---|---|---|
| `entrpen`, T21 grey+1M, 1 step | NaN | 1.0638e-4 (FD agrees to 1.1e-5) |
| `solar_constant`, T63 JW real terrain | NaN | 5.667e-4 (matches the stop-gradient ablation value) |
| `entrpen`, T21 RRTMGP+2M, 2 steps | NaN | -1.73e-4 (FD -2.13e-4 at eps=1e-3; same sign/magnitude, truncation through the sigmoid trigger gates) |

**`fix(1m)`: stable expm1 form for the ice-sedimentation influx.** The factored `zal2·(1−zal1)` with an epsilon floor destroys a removable 0/0 cancellation: `d(zal2)/d(zxiflux) = 1/max(ρv, eps)` reaches 1e12 per level on the scan carry (an unbounded backward amplifier), and the forward silently dropped incoming flux in layers below the floor (a small mass leak). The `cvtfall` test fixture is updated to a contiguous mixed-phase deck — the old disjoint one only passed via the amplifier artifact (the riming path it names was never active) — and now cross-checks AD against FD (agrees to 9 digits).

**`docs(rrtmgp)`:** measured chunking guidance. The overhead when set is not one flat extra forward: `lax.map` serializes the blocks, so gradient wall time grows ~linearly with n (T21: 7.5/14/27/51 s for n=0/2/4/8; T63: 58.7 s at n=8, 104.3 s at n=16). Default (unset) is the identical single-vmap path. Use the smallest n that fits: n=8 at T63L47 → 23.7 GiB peak (unchunked needs 171 GiB even with jax-rrtmgp#16's max-width scan).

duncanwp:
`bc391e3` adds the SPA branch smoothing requested for calibratability of the aerosol activation fit.

**Why:** `spa_prefactor` reaches the physics only through two hard branches — the `min(Nc_min, arg)` physical cap in `spa_activated_cdnc` and the `max(activated_cdnc − cdnc, 0)` nucleation increment. The cap boundary crosses the middle of the cloudy-cell distribution, so the exact local gradient was verifiably correct (FD → AD as ε → 0) yet **wrong-signed and 20× smaller** than the ε=1e-3 secant: an optimizer walks the wrong way. 

**What:** both branches become hyperbolic smooth min/max with config-controlled widths (`AerosolParameters.spa_cap_smoothing`, `CloudParams2M.activation_smoothing`), default 1 cm⁻³ (a few % of typical activated CDNC; forward change ≤ 0.15 cm⁻³). Width 0 recovers the hard scheme exactly, double-where-guarded so the unselected branch never differentiates `sqrt` at zero.

**Acceptance measurement** (2-step T21L47 JW rh=0.9, RRTMGP+2M): AD = +1.4211e-8 with central FD giving the identical value at every ε from 1e-2 to 1e-5 (ratio 1.000 ± 5e-6), and the loss over prefactor 1716→2313 is monotone and convex with the local slope matching the ±0.02 secant to 1%. One honest caveat: the hard version's apparent negative macro trend lived entirely in branch-flip jumps at the ~1e-7 K scale of this short-window objective, so the smoothing reworks the response at exactly that scale — the width parameter makes that modeling choice explicit rather than hiding it in float-exact branch flips.

All 291 aerosol+2M tests pass, and the downstream RRTMGP+2M FD-vs-AD gates pass with the smoothing defaults on.

duncanwp:
`73c67ea` (warm/melt rain-source split + KK2000 CDNC floor) is now exercised end-to-end: JEM-Cal's Phase 2c calibration drives `ccraut`, `crt`, `spa_prefactor`, and `entrpen` against MSWEP precipitation and a warm-rain-fraction target through the full RRTMGP+2M+MACv2-SP stack at T31L47, at ~16 s per gradient step (details on climate-analytics-lab/JEM-Cal#2).

One measured caveat for anyone differentiating through the 2M scheme: the exact adjoint of the precipitation chain grows ~×10 per 12-minute step (AD = FD exactly at 2 steps; 1e5 apart at 5; ~1e+245 by 40 steps — huge-but-finite, not NaN — while FD stays O(1) throughout, i.e. the forward is fine and the cliffs are microscopically narrow). Per-function cotangent taps put the largest per-call gains in `update_precip_fluxes` (~×1300) and `precip_formation_cold` (~×800), compounding through the qr/qs/qnc/qni step carry. Gradient windows must respect this microphysics adjoint horizon (~30 min) until the cold-chain branch structure is smoothed — that's the natural follow-up to this PR's smoothing work.

PR #564, opened 2026-07-11
feat(dycore): pySES CAM-SE backend + CAM frontal GW port + Hydra integration (combines #568)

duncanwp:
Adds the pySES CAM-SE spectral-element dynamical core as a first-class, registered `DynamicalCore` backend (`"pyses_cam_se"`), completing the dycore-protocol skeleton with a second production backend alongside dinosaur. Built from the pyses developer's PG2 coupling prototype, implemented as library code with tests throughout.

## What's here

- **`jcm/dycore/pyses/`** — full protocol implementation over `pyses==0.1.3a2` (optional dep: `pip install jcm[pyses]`):
  - forced step via `run_dycore.advance_coupling_step` (`lump_all` coupling, RK3-5STAGE, tensor hyperviscosity, configurable `nu_top` Laplacian upper sponge, Zerroukat vertical remap); moisture enters as `dr/dt = (dq/dt)/(1−q)²` on the dry-mixing-ratio core; arbitrary `tracer_specs` ride as passive SE tracers;
  - **pg2 physics grid from pyses itself** (Hannah et al. 2021 `gll_to_fv`/`fv_to_gll`; FV→GLL→FV round trip exact to 1e-10 in tests) with DSS via `project_scalar`; physics runs float32 on a `(1, ncol)` layout so `ComposablePhysics(vectorize_columns=True)` works unchanged; dynamics stays float64;
  - full 47-level finite-top ECHAM hybrid (~1 Pa top), USSA-1976 resting IC over real interpolated orography, `build_terrain`/`build_forcing` from the bundled T63 data, `to_xarray` bin-average regrid to lat/lon with hybrid level coordinates.
- **Physics-side enablers** (no behaviour change for dinosaur runs):
  - `jcm.physics.coords_util.column_lat_lon`: solar-zenith/MACv2-SP coordinate caching now supports scattered-column grids (bit-identical legacy meshgrid on separable grids — tested);
  - float32 physics under `jax_enable_x64`: radiation + Tiedtke `lax.cond` branches pinned to carry/state dtypes, a working-dtype invariant at the `ComposablePhysics` seam, and `Model` casting its carry template to an optional `dycore.physics_dtype` — verified by a coupled `Model` smoke in the production f32/f64 split.
- **`tools/run_pyses_climatology.py`** — chunked long-run driver (netCDF + checkpoint per chunk, NaN abort gate, runtime + GPU-memory accounting) for the ne30L47 climatology campaign.

## Tests
`jcm/dycore` 43+ passed (protocol conformance, pg2 identities, USSA/L47/interp helpers, real-geography terrain, sponge config, f64 & f32 coupled Model smokes); broad fast suite **1348 passed**; ruff clean. Design record: `docs/source/design/pyses_cam_se_dycore.md` (includes ne30 open items: element- vs column-axis sharding, offline forcing regrid, sponge tuning).


tools/run_pyses_climatology.py
```diff
@@ -0,0 +1,177 @@
...
+    while day_done < args.days - 1e-6:
+        chunk = min(args.chunk_days, args.days - day_done)
+        t_chunk = time.time()
+        preds = model.resume(
+            forcing=forcing,
+            save_interval=args.save_interval,
+            total_time=chunk,
+            output_averages=True,
+        )
+        ds = preds.to_xarray()
```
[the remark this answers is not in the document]
duncanwp: Right diagnosis — this bit us live on the first driver smoke and is fixed in 0081fca (pushed before the climatology runs started): `ModelPredictions` now takes the dycore and delegates `to_xarray` to it whenever `coords.horizontal` has no `modal_axes`, so the generic dinosaur serializer only runs for modal grids. The ne30 runs' chunk netCDFs are being written through the delegated `PysesCamSEDycore.to_xarray` (bin-average regrid) — verified by opening the day10/day20 chunk files from the first campaign attempt.

duncanwp:
This PR now **combines #568** (the CAM frontal spectral GW port + the `physics_fields` dycore contract — merged into this branch at d00500b/afa2c85 with the pySES frontogenesis provider added on top) and additionally carries, per review:

- **Hydra integration** (ad6c382): `dycore` config group with the production-validated `dycore/pyses_ne30l47.yaml` (every stability setting annotated with its rationale from the ne30 campaign), `run/pyses_year.yaml`, and `runners` dispatch — a production year is now `python -m jcm.main dycore=pyses_ne30l47 physics=echam-rrtmgp-2m run=pyses_year`. The bespoke campaign driver and launcher are **deleted**; CLAUDE.md now mandates Hydra-first, no bespoke run scripts.
- `run_chunked` hardening from the campaign's lessons: health-gate **before** checkpoint write, `.prev` rotation, periodic checkpoint archives (`run.archive_ckpt_every`).
- JAM aerosol forcing files on the pySES column path fail loudly (NotImplementedError) instead of running silently aerosol-dark.

Reviewer note: `gw_scheme` defaults to `hines`; the frontal scheme ships dormant (correct + tested in isolation; its thin-lid dynamical interaction is an open research item documented in the design doc).

jcm/dycore/pyses/forcing.py
```diff
@@ -0,0 +1,123 @@
...
+_TIME_FIELDS = {
+    "sst": "sea_surface_temperature",
+    "icec": "sice_am",
+    "stl": "stl_am",
+    "soilw_am": "soilw_am",
+    "snowc": "snowc_am",
+}
+
+
+def build_forcing(forcing_file: str, dycore, *, validate: bool = True) -> ForcingData:
```
duncanwp, line 42: Why can't this just call the original build_forcing method then do the grid to points? Would save a whole module. Possible Refactoring target

jcm/initial_states/ussa1976.py
duncanwp, line 1: Why does this live in pyses? Let's create a module / folder of initial state helpers since we have a few now

jcm/physics/echam/echam_terms.py
```diff
@@ -103,6 +104,15 @@ def echam_physics(
...
         hines: Override for non-orographic GW ``HinesParameters``.
         sso: Override for sub-grid-scale orography ``SSOParameters``.
+        gw_scheme: Non-orographic gravity-wave scheme: ``"hines"`` (ECHAM's
+            Doppler-spread scheme, the default), ``"frontal"`` (CAM's
+            frontogenesis-triggered spectral scheme — requires a
+            frontogenesis provider, e.g.
+            ``DinosaurDycore(compute_frontogenesis=True)``), or ``"none"``.
+            The two schemes fill the same role (mid-atmosphere momentum
+            deposition from non-orographic waves), so they are exclusive
+            alternatives — running both would double-count the drag
```
duncanwp: This comment contradicts the implementation which allows a 'both' option

jcm/physics/physics_term.py
```diff
@@ -76,6 +76,13 @@ class PhysicsTerm(nnx.Module):
     category: ClassVar[str] = ""
     requires: ClassVar[tuple[str, ...]] = ()
     provides: ClassVar[tuple[str, ...]] = ()
+    # Diagnostic fields this term needs the DYCORE (or an upstream term)
+    # to supply each step — e.g. the frontogenesis function, a
+    # horizontal-gradient quantity only the dynamical core can compute
+    # faithfully. Model validates at construction that each named field
+    # is covered by DynamicalCore.physics_field_names() or an upstream
+    # term's ``provides``.
+    requires_dycore_fields: ClassVar[tuple[str, ...]] = ()
```
duncanwp, line 95: Please go through each term and check that the requires that it needs are actually listed - I suspect they're not. I guess almost all of them need e.g. specific_humidity, so maybe factor that out somehow

jcm/dycore/pyses/forcing.py
duncanwp: Agreed it's a good refactor target. Half is already true: the JAM leaves go through the shared jcm.forcing readers + one bilinear column sampler. The met half re-implements the monthly stacking because the spectral loader validates an exact Gaussian-grid match and its leaves come back grid-shaped — but a generic 'load with the standard reader, tree-map every leaf through _leaf_to_columns' would collapse most of the module. Since #575 rewrites this file substantially (JAM + ozone attachments), I'd do the refactor there after this merges rather than churn both. Tracking it in the #575 description.

jcm/initial_states/ussa1976.py
duncanwp: Done in 009fb89: new jcm/initial_states package; USSA-1976 moved there (it is dycore-agnostic) with imports retargeted. The JW/balanced-isothermal injectors stayed in runners since they mutate a built Model — happy to migrate their pure profile math too if you want them usable standalone.

jcm/physics/echam/echam_terms.py
duncanwp: Fixed in 009fb89 — the docstring now lists 'both' (Hines background + frontal deposition, with the retune caveat) instead of calling the schemes exclusive.

jcm/physics/physics_term.py
duncanwp: Audited every term with an AST pass in 009fb89: one apparent violation (Lohmann2M's tke read) turned out to be membership-guarded, so all terms are compliant — but the convention was implicit, so it's now documented on PhysicsTerm.requires and permanently enforced by jcm/physics/requires_audit_test.py (unconditional reads must be declared; optional ones via .get or an 'in diagnostics' guard). On specific_humidity etc.: prognostics arrive on PhysicsState, not diagnostics, so they never need declaring — also now stated in the convention.

Issue #581, opened 2026-08-04
AeroCom Phase 4: add the union of requested diagnostics as optional output


duncanwp:
We intend to submit to four AeroCom Phase 4 experiments (tracked in `jcm-experiments` #7-#10):
[AP4-CTRL](https://github.com/aerocom-community/experiments/blob/main/phase-4/AP4-CTRL/AP4-CTRL.md),
[aci-baseline](https://github.com/aerocom-community/experiments/blob/main/phase-4/aci-baseline/aci-baseline.md),
[MMPPE](https://github.com/aerocom-community/experiments/blob/main/phase-4/MMPPE/MMPPE.md),
[imo2020](https://github.com/aerocom-community/experiments/blob/main/phase-4/imo2020/imo2020.md).

Their diagnostic lists overlap heavily, so this issue is the **union**, checked against what a current `physics=echam-jam` run actually writes (186 variables). Everything below should be **optional output** — off by default, selectable per run, since most of it costs something.

(Note AP4-CTRL's own variable table is marked *"TBD; not yet current"*, so the ACI/MMPPE/IMO lists are the working superset.)

## A. Already emitted — only need CF/CMOR names

`clt`(total_cloud_cover) `rlut`/`rlutcs`/`rsut`/`rsutcs`/`rsdt`(radiation.toa_*) `rsds`(surface_sw_down) `hfls`/`hfss`(latent/sensible_heat_flux) `ps` `ts` `ta` `hus` `hur` `cl`(cloud_fraction) `cli`/`clw`(qi/qc) `pfull` `zg` `rho` `dzhalf`(layer_thickness) `cdnc`(droplet_number) `cdr`/`icr`(r_eff_liq/ice) `angstrm` `hdtcbl`(pbl_height) `lsm` `orog` and the aerosol mass/number tracers for `BC`/`OC`/`SO4`/`SS`/`DU`/`DMS`/`SO2`.

**Ask:** a CMOR name-mapping layer + AeroCom file-naming writer (`aerocom_<Model>_<Exp>_<Var>_<VertCoord>_<Period>_<Freq>.nc`, one variable per file). Cheap, unblocks a large fraction of the request.

## B. Derivable from existing state — should be native diagnostics

Column integrals and simple algebra, currently left to post-processing:
[...]

Issue #10, opened 2024-08-07
Implement a more complete radiation scheme


duncanwp:
For example [RRTMG](http://rtweb.aer.com/rrtm_frame.html). This is a fair amount of code, but it might help building tests etc. that it has already been wrapped in Python here: https://github.com/climlab/climlab-rrtmg

PR #582, opened 2026-08-04
feat(diagnostics): AeroCom phase-4 diagnostics, CMOR post-processing, and the CALIPSO/MODIS simulators

duncanwp:
Adds the derived diagnostics the four AeroCom phase-4 experiments request, plus a post-processor that writes submission-ready files. Tracked in climate-analytics-lab/jcm-experiments#7-#10; the diagnostic gap analysis is #581.

## Design: a Physics module, not a PhysicsData struct

`AerocomDiagnostics` is a single **diagnostic-only** `PhysicsTerm` (`category="diagnostics"`, following `wmo_tropopause` / `cosp_cloudsat`) placed **last** in the term list — after everything that can still modify the state, including the JAM post-cloud chain and the gravity-wave drag whose winds the pressure-level fields interpolate.

It emits a **plain dict, not a typed struct**. A `@tree_math.struct` exists to carry scheme state *between* terms (`CloudData`, `ConvectionData`); the AeroCom set is terminal — nothing downstream consumes it — changes with each protocol revision, and is selected per experiment. A ~100-field struct would force all-or-nothing evaluation and churn constantly. A dict is also what `output_averages` already accumulates.

Groups let a run pay only for what its experiment asks:

| group | contents |
|---|---|
| `cloud` | cloud-top sampled `ttop`/`cdr`/`icr`/`cdnc`/`lcc`/`icc`/`clt`, layer+column optical depths, condensate paths (`lwp`/`iwp`/`cllvi`/`clivi`) |
| `column` | `prw`, column droplet/ice number, `albedo` |
| `plev` | winds on pressure surfaces, `lts` |
| `aerosol` | per-species column burdens, `N70`/`N100`, `PM1`/`PM10` |

## Cloud-top sampling

A direct translation of the phase-2 Indirect3 Fortran reproduced in the aci-baseline protocol — all three overlap hypotheses (maximum / random / maximum-random), both thresholds, and the "uppermost layer is cloud-free" assumption — written as a `lax.scan` over the vertical so it stays differentiable. Outputs are **grid-box means, not divided by cloud cover**, per the protocol Q/A.

The physics-internal frame is TOA-first, which already matches the protocol's `k=1`-is-top convention, so no flip is needed. (Saved output is surface-first; that flip lives in the writer.)

## `tools/aerocom_cmor.py`

Maps jcm names to CMOR names and units (including the micron→metre effective-radius conversion), sums per-tracer burdens into per-species totals, and writes **one variable per file** as `aerocom[4]_<Model>_<Exp>_<Var>_<Vert>_<Period>_<Freq>.nc`. The no-underscore / ≤20-character `<ModelName>` rule is enforced up front rather than discovered at submission. Verified against a real 30-day T63L47 run: **35 variables mapped**; unmapped variables are reported so the coverage gap stays visible.

## Usage

```bash
python -m jcm.main physics=echam-jam-aerocom grid=echam_t63_l47_hybrid
python tools/aerocom_cmor.py out_day30.nc --model JCM-v2 \
    --experiment all_2000 --period 2010 --out submission/
```

## Tests (28 pass)

- **Protocol agreement**: an independent NumPy transcription of the published Fortran must match the vectorised `lax.scan` for every overlap hypothesis.
- Clear-sky, single-opaque-layer, and both-threshold edge cases.
- **Gradient finiteness** through the scan and the lognormal integrals.
- Pressure interpolation exact for a field linear in log(p); clamps rather than extrapolates below ground.
- **End-to-end**: the term is terminal in the ECHAM term list, and enabling it leaves the trajectory bit-for-bit unchanged (it is genuinely diagnostic-only).
- `requires_audit` and the ECHAM suite still pass (38 passed, 17 skipped).

## Deliberately not included

These cannot be computed post-hoc and need opt-in flags on their owning terms (#581): aerosol-free (`*noa`/`*_na`) radiation, per-species/per-mode/per-wavelength optics, microphysical process rates (`autoconv`/`accretn`/`wbf`), and emission fluxes (`emi_*`). Also outstanding: wiring the CALIPSO/MODIS simulators that already exist in jax-cosp, and a 3-hourly instantaneous output mode.


duncanwp:
## Codex review addressed (5 P1s) + measured cost

All five findings confirmed against the code and fixed in 5ab52e2, each with a named regression test.

| # | finding | fix |
|---|---|---|
| 1 | `geopotential` (m²/s²) written as `zg` (m) — every profile too large by g | map `height_full` (already metres); keep `geopotential` as a fallback with an explicit 1/g conversion |
| 2 | 2M never populates `CloudData.droplet_number`, so the advertised preset reported **zero CDNC/cdnum** | source the prognostic `qnc` tracer; fall back to the 1M volumetric field |
| 3 | `droplet_number` is m⁻³ not kg⁻¹, so dp/g folded in a spurious density factor | integrate the per-mass tracer with dp/g (exact for it) |
| 4 | `CloudData` has no `ice_number` → `icnum` silently dropped | derive from the `qni` tracer |
| 5 | cloud-borne `mc_*` mass excluded → species burdens undercount once aerosol activates | sum interstitial + cloud-borne + gas |

Thanks — #2 in particular would have produced a silently wrong submission for the headline ACI variable.

### A sixth bug the fixes exposed

Emitting `icnum` conditionally broke the run: **the diagnostics dict is part of the `lax.scan` carry**, so a key set that varies between the initial carry probe and the real step changes the pytree and the scan rejects it. Every group now emits a *fixed* key set, zero-filled where the configuration cannot supply a value, with `provides` declaring them all.

The same defect applied to the per-tracer burden keys, which were built from the live tracer dict (empty in the probe). Those are now fixed **per-species totals** summing interstitial + cloud-borne + gas — which is what AeroCom actually requests, and it reduces the CMOR writer to a rename. That crash was missed because the end-to-end test only exercised `cloud`/`column`/`plev`; there is now one that runs the `aerosol` group with JAM.

### Measured cost

T63L47 on an A100, warm compiled step, against the same run with `enable_aerocom=false`:

| groups | overhead |
|---|---|
| `cloud` | +3.3% |
| `column` | +3.3% |
| `plev` | +3.0% |
| `aerosol` | +3.2% |
| `cloud+column+plev` | +4.4% |
| `cloud+column+plev+aerosol` (the preset) | **+9.4%** |

Most of a single group's ~3% is the term's fixed per-step overhead, so groups cost far less together than their sum. `aerosol` is the largest true increment (~5% on top of the other three) from the per-mode lognormal integrals, and can be dropped to stay near 4%. Documented in both the term docstring and the preset.

No group triggers extra physics or an extra radiation call. The diagnostic that would — aerosol-free (`*noa`) radiation, which roughly **doubles** runtime since radiation is ~87% of the step — is deliberately excluded and gated separately in #583. Remaining coverage: #584 (per-species/mode optics), #585 (process rates + emission fluxes), #586 (3-hourly instantaneous output).

25 tests pass; lint clean.

duncanwp:
@codex review again please

duncanwp:
## Second subtask added to this PR: CALIPSO + MODIS simulators (7c28011)

I intended this as a separate stacked PR, but the commit had already been pushed to this branch and splitting it would have required a force-push that rewrites a branch with an open PR — not worth the risk for a review that is the same person either way. **Say the word and I'll split it**; otherwise this PR now covers both subtasks off #581. The two commits are cleanly separable if you prefer to review them one at a time.

### What the second commit adds

The CALIPSO lidar and MODIS imager simulators **that already exist in jax-cosp**, wired into the existing COSP term — 10 of the 11 CFMIP diagnostics aci-baseline and imo2020 request:

`cltcalipso` `cllcalipso` `clmcalipso` `clhcalipso` · `cltmodis` `clwmodis` `climodis` `tauwmodis` `tauimodis` `reffclwmodis` `reffclimodis` `lwpmodis` `iwpmodis`

Both run on the **same SCOPS realization** as the radar — COSP's own wiring, so the instruments see a consistent cloud field and the expensive subcolumn sampling is paid once. A test asserts the radar diagnostics are unchanged when they're enabled; it would fail if the realization were regenerated.

### Cost — above the 10% budget, hence off by default

Measured at T63L47 on an A100 vs no COSP:

| configuration | overhead |
|---|---|
| radar only (existing) | +4.5% |
| radar + MODIS | +12.8% |
| radar + CALIPSO | +20.1% |
| all three | **+20.4%** |

MODIS is nearly free once CALIPSO is paying for the lidar optics, so enable them together. CALIPSO is the expensive one (its own 532 nm optics per subcolumn plus regridding onto the 40×480 m grid). Intended for the ACI/IMO2020 runs, not general production — documented in the term, the factory, and the new `physics=echam-jam-aci` preset (~30% combined with the derived diagnostics).

### Two findings

**A convention trap, now pinned by a test.** Effective radius zero means *opposite* things to the two instruments: the radar reads `reff == 0` as "use PSD defaults", `lidar_optics` reads `radius <= 0` as "class absent". A run that leaves the radii unset reports **zero CALIPSO cloud cover** — this is how my first test run failed, and in production it would look exactly like a bug.

**MODIS `R_UNDEF` sentinels** for dark/cloud-free retrievals are mapped to zero so they can't poison the time average — legitimate because these are grid-box means not divided by cloud cover.

### Not included

`parasolRefl` — jax-cosp has the PARASOL optical-depth inputs but no reflectance simulator. Filing upstream; it's the only one of the eleven we can't produce.

43 tests pass across `jcm/physics/diagnostics/`; lint clean.

duncanwp:
## Third subtask added: process rates + emission fluxes (#585)

Both are values the schemes already compute; the work is publishing them.

**Process rates** (2-moment), column-integrated [kg m⁻² s⁻¹], following the existing `rain_formation_warm`/`rain_from_melt` convention: `autoconv`, `accretn`, `wbf`.

`precip_formation_warm` only returned the *combined* autoconversion+accretion increment, so the two warm-rain pathways are now returned separately with the same cloud-fraction / precip-cover weighting as the combined term. WBF is derived at its call site from the in-cloud liquid lost across `WBF_process`, leaving that routine untouched.

One convention worth noting: the new scalars go **before** the flux profiles in the return tuple. The profiles are last by design ("so existing `tend, rain, snow, *_` call sites keep working") and tests unpack them as `*_, rain_b, snow_b`; appending broke two of them, which is how I found out.

**Emission fluxes** `emi_<species>` for so2, so4, bc, oc, poa, soa, ss, du, moa, dms — via a shared helper that column-integrates each term's **own mass tendencies**. Deriving them from the tendencies rather than re-implementing five different flux formulas means the diagnostic cannot drift from the mass the term actually adds, and elevated injection (industrial, biomass burning) needs no special case since the integral spans the column. Fluxes accumulate across terms because several emit the same species. Uses `rho·dz` rather than `dp/g` (hydrostatically equivalent) purely because every emission term already reads both, so no new `requires` entry.

Both key sets are static for the scan-carry reason established earlier in this PR.

**Cost: none measurable.** No new computation — just reductions over fields already in registers. That keeps it well inside the ~10% default-on budget, unlike COSP (+20%) and the `*noa` radiation (~2×, #583) which stay gated.

**Tests:** 8 new for the flux helper (unit parsing, per-mode summation, cross-term accumulation, static key set, elevated injection, and an end-to-end check that `emi_ss` equals the mass `SeaSaltEmissions` actually injects). 117 cloud tests pass, 96 across diagnostics + emissions.

Remaining from #581: `*noa` radiation (#583), per-species/mode optics (#584), 3-hourly output (#586).

duncanwp:
## Fourth subtask added: per-species / per-mode / spectral aerosol optics (#584, 6c69850)

### The decision this needed, and why it isn't obvious

AeroCom asks for AOD split by component (`od550so4`, `od550bc`, `od550oa`, `od550dust`, `od550ss`, `od550aerh2o`) and absorption likewise. **JAM is internally mixed**: within a mode every species is volume-mixed into *one* effective refractive index before the Mie call. There is one particle. Extinction is a property of that mixed particle — there is no per-species extinction sitting in the calculation to read out, and a sulfate coating changing a soot core's absorption is the entire point of mixing internally.

Two honest options:

1. **Re-run Mie per species**, pretending each mode holds only that species. Well-defined (an *external* mixture), but it's a different optical model from the one radiation used, the parts **don't sum to the total**, and it costs ~7× the optics.
2. **Apportion** the mode's computed extinction. Parts sum to what radiation saw, negligible cost, but the split is a convention.

This does (2), labelled as an apportionment throughout — with **different weights** for the two quantities:

| quantity | weight |
|---|---|
| extinction | `V_s / ΣV` (volume fraction) |
| absorption | `V_s·k_s / Σ(V_s·k_s)` (k-weighted volume) |

The absorption weight is not ad-hoc: under the volume mixing rule `Σ V_s k_s` **is** `V_tot·k_eff`, so it is exactly the linear decomposition of the effective imaginary index the optics were computed from. A volume split would hand the absorption to whatever is most abundant instead of to soot, which is a tiny volume fraction and nearly all the absorption.

### Checked against ECHAM-HAM rather than assumed

`ham_rad_diag` (`mo_ham_rad.f90:1645-1972`) resolves the same tension the same way — it apportions, with the identical pair of weights. Their comment at `:1839-1841` states it outright: *"based on volume average for optical thickness, additionally weighted with ni for absorption"*; weights at `:1926-1936`. Water is its own component in both, and HAM likewise emits the split at **550 nm only**. So these fields are directly comparable to HAM's `TAU_COMP_*`/`ABS_COMP_*`.

Two deltas: HAM's Ångström pair is 550/865, so `ang550865aer` is published alongside AeroCom's 440-based `ang4487aer`; and HAM applies the volume apportionment even when its radiation ran Maxwell-Garnett/Bruggeman, making the split inconsistent with its own optics there. JAM has only the volume rule, so that can't arise.

**Caveat now documented**: coating/lensing enhancement is real in `abs_mode` and gets credited to the absorber, so `abs550_bc` is "BC *including* its coating enhancement", not "what bare BC would absorb". Don't difference it against an external-mixture calculation and call the residual an error.

### Wavelengths

Evaluated at the **observation** wavelengths — 355/440/550/670/865 nm — not at radiation band centres. `refractive_index_at` interpolates in log₁₀(λ), so `od550aer` is at 550 nm rather than "the band nearest 550 nm", which moves with the radiation config and is a single broad band under grey.

Adds: `od550aer` `abs550aer` `od355aer` `od440aer` `od670aer` `od865aer` `ssa440aer` `ang4487aer` `ang550865aer` `aerindex` `ec355aer` (3-D), plus `od550_<sp>`/`abs550_<sp>` for so4/poa/soa/moa/bc/du/ss/wat and `od550_mode_<m>`/`abs550_mode_<m>` for the 4 modes. 34 new output variables (239 → 273). `tools/aerocom_cmor.py` groups the three organics into `od550oa` — **summed**, not overwritten, which has its own test.

### Cost — essentially free, because of the gate

| config | s / sim day |
|---|---|
| `echam-jam-aerocom` | 240.7 |
| `echam-jam-aerocom-optics` | 241.1 |

**+0.2%, below the noise floor of a single-run comparison** — read it as "under ~1%", not as a precise 0.2%. That's the radiation gate working: the extra Mie sweep runs only on radiation-compute steps and adds 5 wavelengths against the shortwave's 14 bands of already-gated work. Off by default (`aerocom_optics: false`); new `echam-jam-aerocom-optics` preset turns it on. Requesting it without `aerosol_module=jam` now raises rather than silently writing a file missing the requested fields.

Everything except `ec355aer` is reduced to a column integral *inside* the term — keeping `n_species × n_wavelength × nlev` arrays in the scan carry would cost hundreds of MB at T63L47.

### Tests (26 new)

Closure is the load-bearing property: components sum to the total exactly for both extinction and absorption, including with strong absorbers stripped out (where the `Σ V_s k_s` denominator gets small). Beyond that, `test_angstrom_discriminates_fine_from_coarse` is a *physical* check — Ångström >1.5 for accumulation mode, <0.5 for coarse; a swapped wavelength pair would pass every closure test and fail this one. Plus scan-carry key stability and the radiation-gate `lax.cond` pytree.

Design note in `docs/source/design/aerosol_optics_diagnostics.md`.

---

## Also in this push: a reproducible benchmark harness (633eba9)

`tools/benchmark.py` + `jcm-run` / `jcm-benchmark` skills, since we keep doing this by hand. It exists because **jcm's `sim days/hr` line is cumulative including compile** — reading it as throughput is what produced the retracted jax-rrtmgp regression report. It parses per-chunk walls, discards the compile chunk, and refuses to report a rate until two consecutive chunks agree within 3%.

Two bugs it had, found by *running* it rather than reasoning about it, both fixed:

- It reported **0% GPU utilisation** for a run pegged at 82-98%. "Drop the first 20% of samples to skip compile" assumes compile is a small leading fraction — it was 71% of that run's wall clock. Now summarised over active samples, plus an `active_fraction` that exposes the compile/compute split (and flags a host-bound long run).
- The preset hard-coded an ozone path to a file that no longer exists, so it would have silently benchmarked radiation against the **analytic** surrogate — ~7.6× the tropospheric ozone column, i.e. the wrong workload. `forcing.ozone_file: auto` already resolves the packaged `t63/ozone.nc`; the correct answer was to pass no override.

Note `.gitignore` needs `.claude/*`, not `.claude/`, for the skills negation to be reachable — git won't descend into an excluded directory.

duncanwp:
@codex review this latest version

duncanwp:
## CI fix + re-review request

`requires_audit_test` was failing on this PR — `CloudsatCosp` read `pressure_half` from the diagnostics dict unconditionally without declaring it in `requires`. Introduced in 7c28011 (CALIPSO/MODIS), where the read was added for the layer-mass optical-depth integral and to pass through to the simulators.

Genuine unconditional dependency, so it's declared rather than routed through `.get`. Undeclared reads are how a term ends up silently ordered before whatever produces its input — which is precisely what that audit exists to prevent.

Also note the benchmark harness and the `jcm-run`/`jcm-benchmark` skills have been **moved out of this PR** to #588 (the skills PR), where they've been merged with the Derecho skill and the duplicated throughput/GPU logic factored into `tools/chunk_timing.py` / `tools/gpu_util.py`. They only landed here because #584's cost measurement needed them; nothing in the diagnostics depends on them. The measured #584 cost stands unchanged (241.1 vs 240.7 s per simulated day, i.e. under ~1%).

@codex review

duncanwp:
## Both P1s from the 02:39 review addressed (5e2d1e2) — CI now green

### P1 · Simulators fed stale condensate

Correct and consequential. `CloudData.qc`/`.qi` are the **step-start** values — microphysics returns its effect as a tendency under operator splitting, so the struct still holds pre-microphysics condensate when this diagnostic runs. The post-microphysics fields go to `thermo_run` (added in caf7ae3 for exactly this reason) and the simulators were not reading them, so MODIS optical depth and the CALIPSO `mr_lsliq`/`mr_lsice` path both disagreed with the saved tracer state whenever microphysics was active — i.e. essentially always.

Now resolved once, immediately after `clouds` is bound, so the radar and imager cannot diverge from each other either. Falls back to `CloudData` when `thermo_run` carries no condensate, keeping the choice structural (Python-level) and therefore stable across scan steps.

**The existing tests failed on this fix, correctly.** The fixture ran `MoistAirColumnState` over a zero state and then seeded `clouds` directly, leaving `thermo_run` cloud-free — so it was encoding the very assumption being fixed. The fixture now keeps the two consistent as the real model does, and a new test drives them deliberately *apart* (thermo_run emptied, CloudData still cloudy) to pin which one wins.

### P1 · COSP fields dropped by the CMOR writer

Correct — all 14 landed in `skipped`, so the advertised `echam-jam-aci` run produced the diagnostics and then no submission files for them. Added with CMOR names, units, vertical type, the 0-1 → percent scaling for cloud fractions, and CF standard names. Verified against `CloudsatCosp.provides`: **14 written, 0 skipped**.

### Also fixed: a CI failure I had missed

`requires_audit_test` was red on this PR — `CloudsatCosp` read `pressure_half` unconditionally without declaring it in `requires` (introduced in 7c28011). Genuine unconditional dependency, so declared rather than routed through `.get`; undeclared reads are how a term ends up ordered before whatever produces its input.

All checks now pass: both builds, lint, claude-review. 314 tests green locally across the affected suites.

---

Note the benchmark harness and the `jcm-run`/`jcm-benchmark` skills have moved out of this PR to #588, where they are merged with the Derecho skill and the duplicated throughput/GPU logic factored into `tools/chunk_timing.py` / `tools/gpu_util.py`. They only landed here because #584's cost measurement needed them. The measured cost stands unchanged: 241.1 vs 240.7 s per simulated day, under ~1%.

@codex review

duncanwp:
## Both P1s addressed — and the class swept this time

You were right on both, and they are the same defect as the condensate one from the last round. I fixed that instance and did not check for others; these are the others.

### The cause is structural, so the fix is too

Operator splitting means **every** term sees the step-start `state` and returns a tendency; the driver sums them and the dycore applies that sum once, afterwards. Any diagnostic reading `state.tracers` therefore reports a field one step behind the tracers saved at the same timestamp. `thermo_run` already solved this for T/q/qc/qi — there was simply no equivalent for tracers.

Rather than patch the two call sites, `ComposablePhysics` now publishes its running tendency accumulator as `_tendency_run` before each term. A term at position *i* sees the sum over terms [0, *i*); the diagnostics terms that consume it run **last**, so that is the full physics tendency. Costs nothing (a reference, not a computation), the tracer key set comes from `state.tracers` so the structure is fixed for the scan carry, and it is excluded from netCDF output.

- **P1 number tracers** — `aerocom_cdnc3d`, cloud-top `cdnc`, `cdnum`, `icnum` now post-physics. These were wrong on every step the 2M microphysics did anything.
- **P1 aerosol burdens** — `burden_*` now integrate post-physics mass, so they no longer miss the step's emissions, chemistry, sedimentation, deposition and scavenging.

Three tests pin it: the tendency is applied; absence of the view degrades to the raw tracer rather than to zero; and the result is floored at zero, since a sink can exceed a trace tracer's step-start burden within one step.

### The sweep I should have done two rounds ago

Every `diagnostics`-category term checked for the same assumption:

| site | status |
|---|---|
| `aerocom.py` number tracers, burdens | **fixed here** |
| `aerocom.py:438` condensate | already via `thermo_run` (caf7ae3) |
| `cosp_cloudsat.py` condensate | fixed in 0983bb1 |
| `cosp_cloudsat.py` remaining `clouds.qc` refs | `zeros_like` shape templates — legitimate |

No further occurrences. I have added "fix the class, not the instance" to the dev-workflow skill in #588, with these rounds as the worked example — four times in this review loop a finding was one `grep` away from having been caught the round before.

**1099 physics tests pass**, ruff clean.

@codex review

duncanwp:
## All three addressed — and my last "swept the class" claim was wrong

Two of these are the *same* staleness class for the third round running, so the honest note first: last round I said I had swept it. I had grepped for the two forms I'd already been shown (`state.tracers`, `clouds.qc`) rather than enumerating every prognostic read. **The `_tendency_run` accumulator I added in that very commit already carries `u_wind`/`v_wind`/`temperature`/`specific_humidity`, and I consumed only the tracers.** Both of your P1s were fixable by the mechanism I had just built.

### P1 · `aerocom_prw` used step-start humidity
### P1 · `u200`/`v200`/`u700`/`v700` interpolated step-start winds

Both now go through a generalised `_post_physics()`. It reads `_tendency_run` rather than `thermo_run` **deliberately**: `thermo_run` is advanced only by terms that opt in, whereas the accumulator captures every term's returned tendency automatically — including radiation and gravity-wave drag, which move the saved winds and temperature without ever touching `thermo_run`. Your wind example is exactly that case.

### The sweep, done properly this time

Enumerated **every** `state.<field>` read in every `diagnostics`-category term, rather than the patterns already reported:

| site | verdict |
|---|---|
| `aerocom.py` prw, plev winds | **fixed here** |
| `aerocom.py` number tracers, burdens | fixed last round |
| `aerocom.py:456` cloud-group temperature | `thermo_run` — deliberately left, so T/qc/qi in that group come from one self-consistent view |
| `aerocom.py` remaining | `zeros_like` shape templates |
| `cosp_cloudsat.py` T/q | already `thermo_run`; other reads are `.dtype`/`.shape` |

No further occurrences.

### P2 · PM cutoffs are aerodynamic, not geometric

This one is a real physics error rather than bookkeeping, and worth stating plainly: an impactor sizes particles by **settling velocity**, so PM1/PM10 are aerodynamic diameters. Comparing them against the geometric dry diameter over-counts dense modes — with `d_ae = d_g·√(ρ_p/ρ_0)`, dust (~2650 kg/m³) and sea salt (~2200) reach a given cut at ~0.6× the geometric diameter the code assumed. Both submitted PM fields were biased high wherever dust or sea salt contributed.

Now converted before the cut. **Slip correction deliberately omitted**, and commented as such: it matters below ~0.5 µm, where both cuts pass essentially all the mass anyway, so it cannot move PM1 or PM10.

---

**1099 physics tests pass**, ruff clean.

@codex review

duncanwp:
## Both P1s fixed — the emission-flux one is the most consequential yet

### P1 · `emi_*` accumulated across the entire run

Correct, and this is the worst defect this review cycle has surfaced. `accumulate_emission_fluxes` adds each term's contribution to whatever is already in the diagnostics dict — right across **terms**, since several emit the same species — but the model threads the previous step's dict back in as `prev_physics_data`, so it was accumulating across **timesteps** too. Every emission flux grew without bound with run length. Snapshots inflated, time-averages were meaningless, and the submitted `emi_*` fields would have been garbage.

Fixed with a `ResetEmissionFluxes` term that `jam_aerosol_physics` places at the head of the emissions block, so the ordering is **structural** rather than a convention each emitter has to honour. Demonstrated directly:

```
carried emi_so2 = 7.0  -> after reset = 0.0
after one emitter      = 3e-7   (not 7 + 3e-7)
```

Two tests pin it, including one asserting the reset precedes every emitter.

### P1 · `aerocom_lts` used the `thermo_run` temperature

Correct — radiation and gravity-wave drag never advance `thermo_run` even though they move the saved `ta`. Now uses `_post_physics`, same as the winds.

**This is the fourth instance of the operator-split staleness class, and it sat three lines from the winds I fixed two commits ago in the same function.** It is a different miss from the earlier ones though, and worth naming: the value arrived as a **parameter** (`temperature`), so my enumeration of `state.<field>` reads in the source could not have found it. Enumerating textual reads is not sufficient — the invariant has to be checked at each *use*, including values passed in from elsewhere. I have amended the sweep guidance accordingly.

### Three failures caught locally, not in CI

All consequences of adding a term rather than defects in the fix: `ResetEmissionFluxes` had to be a real `PhysicsTerm` (`required_tracers`), and two order/category assertions in `jam_terms_test` needed to account for it. None appeared in the targeted tests — only at full-suite scope. Each would otherwise have been a CI round.

**1101 physics tests pass**, ruff clean.

@codex review

duncanwp:
## Both fixed — and the convective one is a physics-coupling defect, not a diagnostic one

### P1 · Tiedtke never advanced `thermo_run`'s condensate

You are right, and the framing "fresh evidence after the post-microphysics fix" is exactly it: this became **observable** only because this PR pointed the satellite simulators and the AeroCom cloud diagnostics at `thermo_run` to escape step-start staleness — which moved them onto a view that was itself incomplete.

The scheme detrains convective qc/qi (it computes `tendency.tracers["qc"]/["qi"]` and applies them to `clouds` a few lines above) but passed only T and q to `advance_thermo_run`. So `thermo_run`'s condensate stayed **pre-convection**, the 2M scheme advanced that stale base by its own microphysics tendency alone, and every downstream consumer lost convective detrainment entirely. Fixed by passing the tendencies the scheme already has.

### P1 · Cloud-group temperature

Correct, and it is the sibling of the `_plev_group` binding I fixed two commits ago — I overrode the temperature there and left this one, three lines away. Now `_post_physics`. The condensate in that group still comes from `thermo_run`, which is the view carrying qc/qi and which now includes the convective contribution.

**1101 physics tests pass**, ruff clean.

---

## A process note, since this is the fifth round of one issue

Each of my sweeps has failed in a *different* way: not looking past the reported file; not consuming a mechanism I had just added; a value arriving as a parameter (invisible to a source-text enumeration); and now a view that was correct at its consumers but incomplete at its source.

That pattern says my per-round patching is finding **instances** rather than establishing the **invariant**. The underlying property is architectural: under operator splitting, every consumer must know which view it is reading and whether that view is complete. There is currently nothing that enforces it — `thermo_run` is opt-in per term, `_tendency_run` is automatic, and `CloudData`/`state` are step-start, with no rule saying which a given consumer should use.

If further instances surface, I would suggest the next step is not another patch but:

1. a design note in `docs/source/design/` stating the invariant and which view each kind of consumer must read; and
2. a test that asserts each published diagnostic matches the state actually saved at the same timestamp — which would have caught all five of these mechanically.

That is larger than this PR should carry, so I have flagged it to @duncanwp rather than starting it. Happy to open it as a separate issue.

@codex review

duncanwp (commented):
Full review at `e0368ef`, with a triage of all 22 Codex comments against the current head (every comment predates at least one fix commit, so each was re-verified in the code rather than trusted stale).

## Codex triage: all 22 addressed, none stands as written

Verified fixed at head, grouped: **CMOR writer** — `zg` now carries the explicit 1/g scale with `height_full` winning when present; `standard_name` comes from a curated CF table and is omitted (not guessed) elsewhere; all 13 CFMIP MODIS/CALIPSO fields plus `cospwarmrain` are in `NAME_MAP`. **Step-staleness family** (10 comments) — the `_tendency_run` accumulator in `ComposablePhysics` now gives the terminal diagnostics the full physics-step view: post-physics winds/T/q (`_post_physics`), post-physics tracers (`_post_physics_tracer`, floored non-negative per the P2), post-microphysics condensate via `thermo_run` (with Tiedtke now advancing qc/qi so convective detrainment isn't lost — verified at `tiedtke_nordeng.py:1199`), post-physics aerosol burdens and modal amounts (with the documented r_dry approximation), and LTS/cloud-top T from the accumulator rather than the opt-in view. **Droplet/ice number** — `_number_concentrations` prefers the prognostic 2M tracers, integrates per-kg fields with dp/g (the m⁻³ vs kg⁻¹ measure concern is resolved: n/ρ · ρdz ≡ n dz), and publishes `aerocom_cdnc3d` for the writer. **Others** — surface albedo from surface fluxes (field names verified on the radiation struct); cloud-borne `mc_*` mass in burdens; `ResetEmissionFluxes` placed structurally at the head of the emissions chain; aerodynamic-diameter PM cuts with the slip-correction omission argued. Tests at head: aerocom 41 pass, cosp 25 pass, incl. the protocol-agreement NumPy cross-check and the trajectory-bit-identical test.

## Residual findings

**1. (P2) The 1M path re-inflates an already in-cloud droplet number.** `Echam1MMicrophysics` sets `CloudData.droplet_number = base_cdnc * cdnc_factor` — a characteristic **in-cloud** value, nonzero everywhere including clear sky (`echam_1m.py:1249`). `aerocom.py` treats `cdnc_m3` as grid-mean throughout: `_cloud_group` divides it by `cf` to form the "in-cloud" sampler input (double-inflating cloud-top `cdnc` by 1/cf), `_column_group` integrates it over the whole column for `cdnum` (counting droplets in clear sky), and the CMOR'd `cdnc3d` is grid-mean under 2M but in-cloud-everywhere under 1M with nothing in the metadata distinguishing them. Both AeroCom presets inherit `cloud_scheme: 2m`, so this is latent rather than submission-facing — but the term's docstring claims scheme-agnosticism, and this is exactly the trap the earlier Codex droplet-number thread circled without landing on. Suggest: branch on which scheme populated the field (the same structural test `_number_concentrations` already does) and for the 1M path use `droplet_number` directly as the in-cloud sampler input, `droplet_number * cf` as grid-mean `cdnc3d`, and `Σ droplet_number · cf · dz` for `cdnum` — or document 1M as unsupported for the cloud/column groups and zero the fields.

**2. (P2) `phase3d` guard violates the repo's own underflow rule.** `aerocom.py:548`: `has_cond = total > 0.0` guarding `qc / where(has_cond, total, 1.0)`. Per CLAUDE.md ("guard on a physical floor, never `> 0`"), the division VJP forms `-g·qc/total²`, and `total²` underflows to exactly 0 for `total ∈ (0, ~1e-154)` — a window spectral ringing condensate demonstrably reaches (the Phase-2 campaign measured tails at 1e-287). Forward-only use is safe today, but these diagnostics are natural calibration observables (that is half the point of building them), and the first `jax.grad` through `aerocom_ttop`/`lcc` will NaN out of this line. One-character-class fix: `total > 1e-30`. The other `> 0.0` guards in the file (`rho`, `sfc_down`) are physically bounded away from the window and fine.

**3. (P3) `sigmas = (self.mode_sigma_g * n_modes)[:n_modes]`** silently *cycles* the mode widths if the modal scheme ever has more modes than the 4-entry default (mode 5 would get mode 1's σ_g). And `σ_g = 1.0` gives `ln σ = 0` → NaN in both lognormal integrals. An assert `len(mode_sigma_g) == n_modes` and `σ_g > 1` at `__init__` costs nothing.

**4. (P3) CMOR `skipped` report miscounts the optics fields.** `convert()` subtracts `NAME_MAP` sources and burden names from `ds.data_vars`, but not the `od550_*`/`abs550_*` per-species fields that `_collect_optics` *does* write — so a run with the optics preset reports its own written variables as "had no AeroCom mapping". Cosmetic, but that report is the coverage-gap watchdog the PR body advertises, so it shouldn't cry wolf.

**5. (P3) `_jam_state` zero-fallback shape mismatch.** In `_aerosol_group`, the `jam is None` branch emits `N70`/`PM1` etc. as 2-D surface-shaped zeros while the real branch emits 3-D `(nlev, ncols)` fields (CMOR maps them as ModelLevel). If the probe genuinely lacks `_jam_state` while real steps have it, the scan carry pytree changes shape; if a run enables the `aerosol` group without JAM ("silently inactive" per the docstring), it writes 2-D fields into a ModelLevel variable. Making the fallback `jnp.zeros((nlev, *horiz))` removes both.

**6. (P3, naming)** `u200/v200/u700/v700` and `lts` carry `VertCoord="Surface"` in the writer. They're 2-D, but "Surface" is semantically wrong for a 200 hPa wind; AeroCom's `ModelLevel/Surface/Column/TOA` vocabulary fits `Column` least badly. Worth one line in the submission notes either way.

## General impressions and optimization notes

The two structural decisions — the dict-not-struct terminal term with a static key set, and the `_tendency_run` accumulator solving the operator-split staleness class *once* instead of per-diagnostic — are the right ones, and the second retroactively hardens `CloudsatCosp` too. The measured cost table in the docstring (3% fixed overhead, +9.4% all groups) is exactly the evidence a reviewer wants. The scan-carry constraint (static keys, zero-filled) is documented at every site it bites, which will save the next contributor a day.

Optimizations, none blocking: (a) the burden loop's `_post_physics_tracer` calls inside the species×tracer double loop re-fetch `_tendency_run`/`_dt_seconds` per tracer — hoisting is trivial but XLA almost certainly CSEs it, so only worth it for trace-time; (b) `cloud_top_sample` carries 7 fields through the scan — packing the 6 accumulators into one stacked array would shrink the carry pytree, but at T63 this is noise; (c) the per-mode Python loops in `_aerosol_group` are fine at 4 modes and the comment already justifies them. I'd spend the effort on finding 1 instead.

https://claude.ai/code/session_01Aqshaa5nDrxcDWgg1FYw8Z

PR #587, opened 2026-08-04
fix(diagnostics): make MoistAirColumnState broadcasting-native

JOhnsonKC201:
`MoistAirColumnState` assumed exactly one trailing horizontal axis when building pressures:

```python
a_full[:, None] + b_full[:, None] * surface_pressure[None, :]
```

That broadcasts on a column-vectorized `(nlev, ncols)` host, but raises on a whole `(nlev, nlon, nlat)` grid:

```
ValueError: Incompatible shapes for broadcasting: shapes=[(8, 1), (1, 96, 48)]
```

Every current caller (the ECHAM terms, `rce.py`, Tiedtke-Nordeng) runs column-vectorized, so the term had never been exercised on a full grid. I hit it composing `MoistAirColumnState` onto `speedy_physics()` to supply `pressure_full` / `pressure_half` / `height_full` for `LottMillerSso`, which no SPEEDY term provides.

Fixed by reshaping the vertical coefficients against `surface_pressure.ndim`, the broadcasting-native pattern documented in `CLAUDE.md`:

> `phalf = a_half.reshape((-1,) + (1,)*ps.ndim) + b_half[...] * ps[None]`, so it too is agnostic to whole-grid vs column-vectorized hosts.

The single-horizontal-axis case is unchanged, so existing callers are unaffected.

## Test plan

The module had no test file. This adds one covering the property that was broken:

- a single column `(nlev,)`, a vectorized block `(nlev, ncols)`, and a whole grid `(nlev, nlon, nlat)`
- that a grid column and a vectorized column agree value for value
- pressure monotonic and finite, half-levels bracketing full levels, and the documented 10 m layer-thickness floor

Results:

- **7 of the 8 new tests fail without the fix**; all 8 pass with it. The one that passes either way is `test_column_vectorized_host`, the layout that already worked.
- `pytest jcm/physics/diagnostics/ jcm/physics/convection/tiedtke_nordeng/ jcm/physics/echam/ -m "not slow"` — 169 passed, 8 skipped.
- `ruff check jcm/physics/diagnostics/` clean.

### DIFF
--- a/jcm/physics/diagnostics/moist_air_state.py
+++ b/jcm/physics/diagnostics/moist_air_state.py
@@ -181,11 +181,21 @@ class MoistAirColumnState(PhysicsTerm):
 
         surface_pressure = state.normalized_surface_pressure * p0  # Pa
         # Hybrid-coordinate pressure: works for pure sigma (a=0) too.
+        #
+        # The vertical coefficients are reshaped against however many
+        # horizontal axes the host actually has. Indexing them as
+        # ``a_full[:, None]`` against ``surface_pressure[None, :]`` hardcodes
+        # exactly one trailing axis, which broadcasts on a column-vectorized
+        # ``(nlev, ncols)`` host and raises on a whole ``(nlev, nlon, nlat)``
+        # grid. See the broadcasting-native convention in CLAUDE.md.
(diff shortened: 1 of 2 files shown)


duncanwp:
LGTM, thanks!

PR #588, opened 2026-08-04
docs(skills): Derecho run workflow as a project skill

duncanwp:
Packages the Derecho PBS workflow from the online-aerosol campaign into `.claude/skills/derecho-jcm-runs/`, so runs are repeatable instead of re-derived — and so the failure modes we hit are not hit again.

## ⚠️ One judgement call to review first

`.gitignore` had `.claude/` ignored wholesale as "local claude files". I narrowed it to:

```
.claude/*
!.claude/skills/
```

so local state (settings, sessions, caches) stays untracked while **shared project skills are versioned**. Verified: `.claude/settings.json` is still ignored. If you'd rather keep `.claude/` entirely private, say so and I'll move the skill to `tools/derecho/` instead — the content is identical either way.

## Contents

| file | purpose |
|---|---|
| `SKILL.md` | the workflow + machine facts |
| `scripts/mkjob.py` | generate a PBS script from flags; pre-flight the config |
| `scripts/watch_job.sh` | completion/failure monitor |
| `scripts/settled_rate.py` | convergence-checked throughput |
| `reference/data_paths.md` | boundary/emissions/aux data: packaged, prepared, upstream |

```bash
python .claude/skills/derecho-jcm-runs/scripts/mkjob.py --name my_run --days 30 --check > runs/my_run.pbs
qsub runs/my_run.pbs
```

Flags cover queue, account, GPUs (adds `+grid.spmd_mesh` and the right memory fraction), walltime, grid, dt, physics package, radiation scheme, advection, aquaplanet, resume, `--bench` variant matrices, and raw Hydra passthrough.

## Why each piece exists

Every default and warning corresponds to something that actually cost time:

- **Hydra `+`/`++` prefixes** — `run=longrun` replaces the run group, so `run.checkpoint_path` needs `+` under it but not otherwise. Two dead jobs.
- **`--check`** composes the config on the login node before a queue slot is spent; `SKILL.md` also gives the coords-constructibility check, because `--cfg job` does not build coords (an invalid truncation only fails at runtime — see T127).
- **Missing-input guard** — the JAM aux inputs live on purge-eligible scratch.
- **`set -euo pipefail` and `-m abe`** — a `sed`-derived script silently dropped job mail, and a script without `-e` reached its `DONE` marker after failing.
- **Monitor hardening** — live NFS logs give stale re-reads; grepping twice produced phantom "failures" that printed empty. Also: chunk netCDFs are written *before* the NaN check, so file existence is not success.
- **`settled_rate.py`** — jcm's `N sim days/hr` is cumulative and includes compile. This is what resolved the recent cross-machine throughput investigation; the skill records the method (per-chunk walls, 5% convergence, and dtype/power checks when comparing machines, since a config flag does not establish precision — one f64 input promotes whole subgraphs).

## Portability

Site paths are env-overridable — `JCM_REPO`, `JCM_VENV`, `JCM_DINOSAUR`, `JAM_INPUTS`, `JCM_EMISSIONS`, `PBS_ACCOUNT`, `SCRATCH` — so this is not pinned to one home directory; defaults match the current setup.

No library code is touched. Scripts lint clean, and the chain was validated end-to-end by submitting a generated job to the real queue: it ran clean (0/186 NaN), the watcher detected completion, and the analyzer correctly declined to quote a rate from a 10-day run.


duncanwp:
## Added: the dev-box skills, with the overlap merged rather than duplicated

Brings over the skills built on the shared UCSD workstation. They did overlap yours in two places, and in one of them the two copies had **already diverged** — so this factors the shared knowledge out instead of shipping both.

### Layering

The problem was that a site skill naturally accretes both "how to drive jcm" and "how this machine works", and the first half is identical everywhere. Split:

| skill | layer |
|---|---|
| `jcm-run` | site-agnostic: config groups, validated stable T63L47 overrides, Hydra traps, watching a run |
| `jcm-benchmark` | site-agnostic: throughput methodology + `tools/benchmark.py` |
| **`devbox-jcm-runs`** (new) | shared UCSD workstation — 8x A100-80GB, **no scheduler** |
| `derecho-jcm-runs` | NCAR Derecho PBS — unchanged in substance, now cross-references the generic layer instead of restating it |

The two site skills contrast usefully: on Derecho PBS hands you exclusive GPUs; on the dev box you self-allocate alongside colleagues and nothing stops two people picking the same card.

### The two real overlaps

**1. Throughput reduction → `tools/chunk_timing.py`.** Your `settled_rate.py` and my `benchmark.py` each had their own copy of the same insight, and they had already drifted apart:

| | yours | mine |
|---|---|---|
| tolerance | 5% | 3% |
| which chunks | last two | first agreeing pair |

Reconciled to the stricter reading of each: **judge on the last two** (yours — a noisy series can match by accident early while still drifting, and what matters is that the run had settled by the end), at **3%** (mine), reporting their **mean** rather than a single chunk, with `--tol` exposed.

I re-checked this against the jax-rrtmgp#22 A/B I just published: 72.16 → 72.23 and 20.62 → 20.47 s/sim-day, ratio 3.50x → 3.53x. Conclusion unchanged, and there is now a test pinning those numbers so the shared module cannot silently move them.

The two entry points stay distinct because they answer different questions: `settled_rate.py` reads a log already back from the queue; `benchmark.py` drives a run and samples GPU telemetry alongside.

**2. Free-GPU detection → `tools/gpu_util.py`.** Runnable directly (`--free`, `--wait N`) and used by `benchmark.py` as a hard pre-flight gate. The uuid join between `--query-gpu` and `--query-compute-apps` is the reason this is a module and not a re-typed shell one-liner.

### `.gitignore` — we both made the same change

You narrowed `.claude/` to `.claude/*` + `!.claude/skills/`; I independently did the same on the AeroCom branch. Kept your wording, added one note: the pattern **must** be `.claude/*` and not `.claude/`, because git does not descend into an excluded *directory*, so a trailing slash makes the negation unreachable. (I hit exactly that.) Verified `.claude/settings.local.json` is still ignored. I have reverted my copy of this change on #582 so it lands once, here.

### Why `devbox-jcm-runs` earns its place

Same principle as yours — every warning is something that actually cost time:

- **Utilisation is not a freeness signal.** Live reading while writing this: GPUs 0 and 3 both showed **0% util** while carrying five and six of other people's processes and tens of GB.
- **Nor is memory alone.** A colleague's Jupyter kernel had been parked on GPU 1 for weeks holding 1.2 GiB, so the card read `1182 MiB, 0%` — indistinguishable from idle driver overhead. I judged it free and started a 3-hour benchmark on top of it. Both signals are now required, and `benchmark.py` refuses to start otherwise.
- **`pkill -f` matches your own shell.** It killed this session's twice. Kill confirmed PIDs.
- Editable installs mean the working tree is the running code — never `git checkout` in the shared clone to A/B a library; use a worktree + `PYTHONPATH`.

### Also

Fixed a bug I introduced: argparse `%`-expands help strings, so the formatted `{DEFAULT_TOL:.0%}` made `--help` raise `ValueError` in both CLIs.

Verified: ruff clean, 11 timing tests pass, both `--help` work, the gate correctly refuses a busy card *and names the free one*, and `settled_rate.py` reproduces the published A/B numbers from a real 30-day log.

One judgement call for you, mirroring yours on `.gitignore`: I put `chunk_timing.py`/`gpu_util.py` in `tools/` rather than inside a skill, since both skills use them and `tools/` is already where repo tooling lives. `settled_rate.py` reaches them via `parents[4]`. If you would rather skills be self-contained, say so and I will vendor them instead.

duncanwp:
## Codex findings addressed (8b8dea2) — all three were real

### P1 · `watch_job.sh` accepted the marker before checking failures

Correct, and it defeated the script's own stated purpose. `run_chunked` logs `unhealthy` and returns **normally** when the health gate trips, so the job script proceeds to write its marker and the watcher reported `COMPLETE` for a run that died on NaNs.

Failures are now evaluated **before** the marker is accepted. No debounce in that path — once the marker is down the log is final, so there is nothing to race against.

Added beyond the report: an explicit non-zero `NaN vars: N/M` check. The health line is printed **per chunk, before the gate reacts**, so a bad count can appear with nothing matching `FAIL_RE` at all. The same rule now guards the late-marker branch, which had the identical hole.

Also fixed something adjacent: the script violated its own first documented lesson. It grepped the log *twice* per check — once for the marker, once for `FAIL_RE` — while the header says "read ONCE per check, then decide and report from those same bytes". Now genuinely one read, which is what makes the stale-NFS reasoning hold.

Verified with synthetic logs:

| log | result |
|---|---|
| clean + marker | exit 0 `COMPLETE` |
| `unhealthy` + marker | exit 1 `FAILED (marker present, but the run is unhealthy)` |
| non-zero `NaN vars` only + marker | exit 1 |

### P2 · `--bench` variant failures did not propagate

Correct. Failures are now counted and the completion marker **withheld** with a non-zero exit.

Extended slightly: a variant that *ran to completion* but went unhealthy also counts, since that is the same normal-return behaviour as the P1 finding — otherwise the fix would only catch crashes and still pass a NaN'd variant.

Deliberately left as-is: `settled_rate.py ... || true`. Its non-zero exit only means "not enough chunks to quote a rate", which is not a job failure. Now commented so it does not read as an oversight.

### P2 · L47 oxidant file hard-coded for every grid

Correct — and the failure mode is the expensive kind, since the runner's level-count validation only fires *after* the job reaches the front of the queue.

Now derived from the requested grid via `grid_layers()`, reusing the parse that already existed for the coords check. The real win is that `check_inputs()` then catches a missing or misnamed file **before qsub**, which is the whole point of that guard.

```
echam_t63_l47_hybrid   -> oxidants_cam_echam_l47_2014_t63.nc
echam_t63_l95_hybrid   -> oxidants_cam_echam_l95_2014_t63.nc
echam_t106_l95_hybrid  -> oxidants_cam_echam_l95_2014_t63.nc
```

`dms`/`dust` are surface fields with no level axis, so they stay grid-independent.

---

## Also added: `jcm-dev-workflow` skill

Requested by @duncanwp — the loop this PR is itself an instance of:

> branch → atomic commits → local test + lint gate → PR linked to its issue → monitor CI **and** the automatic Codex review → fix and respond → `@codex review` if the response was substantial → hand back to a human only when green

One thing it calls out explicitly, because it is exactly how this round went unnoticed for a while: **Codex posts as a PR *review* with inline comments, not as an issue comment**, so `gh pr view --json comments` shows nothing. You need:

```bash
gh api repos/climate-analytics-lab/jax-gcm/pulls/<PR>/reviews
gh api repos/climate-analytics-lab/jax-gcm/pulls/<PR>/comments
```

It also carries the rule that a watcher must match **failure** signatures and not only success — a success-only filter is silent through a crash, which reads identically to "still running". Same class of bug as the P1 above.

### `CLAUDE.md` shrinks

Per @duncanwp, the `Running Tests` and `Linting` sections are folded into the skill. I left a pointer plus the two facts too expensive to lose from always-loaded context: `ruff check .` before every push, and `JAX_PLATFORMS=cpu` (without which every xdist worker fights for the same GPU and XLA dies with `CUDA_ERROR_OUT_OF_MEMORY`). Everything else — CI thresholds, `-n` guidance, coverage, slow-test marking — now lives in the skill only.

Judgement call worth a second opinion: `CLAUDE.md` is always in context whereas a skill is loaded on demand, so this trades guaranteed availability for one canonical copy. I kept the two highest-cost facts inline as a hedge. Happy to pull more back if you would rather not rely on the skill being invoked.

@codex review

duncanwp:
## Second Codex round addressed (45f378f) — all four correct

Two of these are cases where I fixed a bug in one place and left the identical hole one file over. Worth noting as a pattern.

### P2 · `benchmark.py` accepted an unhealthy or truncated run

Correct, and this is the same class as the P1 from the last round. `check_health` trips on more than NaN — temperature range, `q_max` — and `run_chunked` then stops early while `jcm.main` still exits 0. My wrapper looked only at the exit code and the NaN count, so a run that died at day 5 of 30 could print a **converged-looking rate and return success**.

Now also parses the unhealthy signature *and* compares completed days against requested, reporting each reason separately.

This was visible in our own CI log and I walked past it:

```
Chunk 1 | Day 90 (0.25 yr) | FAILED: T_min=66K (< 100K); T_max=598K (> 500K)
```

No NaN anywhere in that line — exactly the case the old check missed.

| log | unhealthy | truncated |
|---|---|---|
| healthy, 30/30 days | no | no |
| `FAILED: T_min=66K`, 5/30 | **yes** | **yes** |
| `atmosphere unhealthy`, 3/30 | **yes** | **yes** |
| clean but 10/30 | no | **yes** |

### P2 · `--bench` variants missed NaN-only logs

Correct, and you're right to call it "a fresh instance" — I fixed precisely this in `watch_job.sh` in the previous commit and left it open here. Each variant redirects jcm output to its own `run.log`, so the outer watcher never sees those per-chunk health lines; they have to be checked at the variant.

Non-zero `NaN vars: N/M` now counts even with no `unhealthy` line — the health line prints per chunk *before* the gate reacts, and a NaN in a non-gated variable produces no gate message at all. Verified: `NaN vars: 0/239` → ok, `NaN vars: 1/239` → UNHEALTHY.

### P2 · `dms`/`dust` pinned to T63 for every grid

Correct, and my own comment was the problem. I wrote that they were "grid-independent" because they are surface fields — which **conflated the vertical with the horizontal**. `read_dms_seawater`/`read_dust_source` validate lat/lon against the model Gaussian grid, so a T106 run passes `check_inputs()` and then dies after reaching the queue front.

All three aux inputs now derive from the requested grid — level count for oxidants, truncation for all — so a wrong-grid file is caught before `qsub`:

```
echam_t63_l47_hybrid    dms=..._t63.nc    ox=oxidants_..._l47_2014_t63.nc
echam_t106_l95_hybrid   dms=..._t106.nc   ox=oxidants_..._l95_2014_t106.nc
echam_t119_l95_hybrid   dms=..._t119.nc   ox=oxidants_..._l95_2014_t119.nc
```

### P2 · The specific-humidity warning was backwards

**This is the one that mattered most**, and thank you for catching it — as written it would have made an agent read a normal ~10 g/kg field as 0.01 g/kg and conclude the model was 1000× too dry.

I verified against the code rather than trusting either source:

- `jcm/dycore/dinosaur/state_bridge.py:120` — `dimensionalize(q, units.gram / units.kilogram)` on the way out.
- `jcm/diagnostics.py:40-50` — documents that an earlier `* 1000` was removed precisely because it double-counted and tripped `q_max > 100 g/kg` on every chunked SPEEDY/ECHAM run.

The label is correct and the values are genuinely g/kg. The warning was stale, carried over from a note that predates the fix. Corrected in the skill, and the note itself rewritten as the correction so it cannot resurface.

---

### Also in this push

Closes the provenance gap I had held over from the RRTMGP A/B. Reports now record resolved path, git SHA/branch and dirty state for `jcm`, `rrtmgp`, `dinosaur`, `mam4_jax` — probed **in the run's own environment**, so a `PYTHONPATH` override shows the code actually imported rather than the override requested. Packaged installs record a version instead. Verified it distinguishes `rrtmgp` 8703bb5 (baseline) from 27076c0 (perf worktree), which is exactly the claim a library A/B needs to be able to make.

@codex review

duncanwp:
## Third round addressed (f5057e7)

Three new findings (the other four in that review were already fixed in 45f378f — same findings re-reported with null line numbers). All three correct.

### P2 · Provenance probed the wrong interpreter

Correct, and it undermines the feature entirely. `_provenance` ran under `sys.executable`, but `--python` routinely points at a different environment from the one running the wrapper — the micromamba env is not on PATH on this box, which is exactly the documented workflow. A report could therefore attribute a benchmark to whichever `jcm`/`rrtmgp` the *wrapper* imports.

**A provenance record that is confidently wrong is worse than none**, since its whole purpose is to be trusted. Now probes with the benchmark's own interpreter and cwd.

### P2 · Emissions pinned to T63

Correct. `_validate_emissions_grid` rejects fields whose horizontal shape differs from the model grid, so the default killed every T85/T106/T119 job *after* it reached the queue front. Derived from `--grid` now:

```
echam_t63_l47_hybrid   -> emissions_echam_t63_l47_hybrid_2014.nc
echam_t85_l47_hybrid   -> emissions_echam_t85_l47_hybrid_2014.nc
echam_t106_l95_hybrid  -> emissions_echam_t106_l95_hybrid_2014.nc
```

`$JCM_EMISSIONS` still wins when set — that is an explicit site choice — but now warns when it does not name the requested grid, since silently reusing a T63 file on T106 is the exact failure being prevented.

### P2 · Ozone silently fell back to the analytic profile

**The worst of the three, because nothing fails.** `auto` resolves only *packaged* climatologies, this repo packages T63L47 only, and the fallback surrogate carries ~7.6× the tropospheric ozone column. An L95 or T106 job runs clean, exits 0, and produces radiation that cannot be compared with anything — no preflight error, no signature a watcher looks for.

`mkjob.py` now refuses any grid without a packaged climatology:

```
NO PACKAGED OZONE for echam_t63_l95_hybrid (t63, L95).
  forcing.ozone_file=auto would fall back to the ANALYTIC profile (~7.6x the
  tropospheric ozone column), which runs fine and silently invalidates the radiation.
  Pass --ozone /path/to/ozone_<grid>.nc — see reference/data_paths.md, or prepare one with
    python -m jcm.data.bc.interpolate_ozone --in T63_ozone_picontrol.nc --out ozone_t63_l95.nc --nlevels 95
```

Verified: T63L47 still generates; T63L95 refuses with instructions.

The pattern across all three rounds is one thing: **`check_inputs()` was verifying that files exist, not that they match the run.** Every prepared input is now derived from `--grid` — level-resolved (oxidants, ozone) and horizontally-resolved (emissions, DMS, dust) alike. `data_paths.md` and the skill table are updated.

---

## Also: benchmark output is now disposable (@duncanwp request)

Model netCDF goes to `/tmp/jcm-bench` (`--scratch-root` / `$JCM_BENCH_SCRATCH`) and is deleted once the report is written. A 30-day T63L47 arm writes GBs and the recent 3-pair A/B filled scratch with fields nobody read; the report, log, GPU telemetry and `result.json` all stay, and reclaimed size is recorded.

Two deliberate limits:

- **Not applied to `jcm-run`.** For a science run the output *is* the point. This belongs to benchmarking only, which is why it lives in the harness.
- **Output is kept when the run was unhealthy.** Deleting the evidence of a failure is how you end up unable to explain it. `--keep-output` forces retention always.

Also reordered the free-GPU gate ahead of directory creation so a refused run leaves nothing behind — noticed because the gate fired *for real* mid-testing when another user took GPU 4 in the seconds between my `--free` check and the launch. Working as intended, but it had already made its scratch dir.

@codex review

duncanwp:
## Fourth round addressed

### P2 · Ozone override nested under the emissions guard

Correct — and this was my own previous fix placed in the wrong scope. **Ozone belongs to radiation, not to the aerosol chain**, so it has to be emitted for every prescribed-forcing run. As written, `--physics echam-rrtmgp` or `--no-emissions` on an L95 grid would *accept* `--ozone` and then silently drop it, leaving `auto` → analytic profile. That is exactly the high-resolution radiation run the guard was added to protect, so the bug landed on the one case that mattered most.

Moved out beside terrain/forcing. Verified both paths now emit it:

```
--physics echam-rrtmgp --grid echam_t63_l95_hybrid --ozone /path/oz_l95.nc
  -> forcing.ozone_file=/path/oz_l95.nc
--no-emissions --grid echam_t63_l95_hybrid --ozone /path/oz_l95.nc
  -> forcing.ozone_file=/path/oz_l95.nc
```

### P2 · `--check` composed with the wrong interpreter

Correct, and the same class as the provenance finding in f5057e7 — I fixed that one in `benchmark.py` and left the identical assumption here. The preflight ran under whichever Python invoked the generator while the job activates `--venv` and runs its own.

The failure mode you name is the bad one: not "rejects a valid job" but **"passes using packages the real job will not import"**, which defeats the entire purpose of a preflight. Now uses the job's interpreter, warning and falling back if it does not exist (the generator can legitimately run off-node).

### P3 · Report written before the retention decision

Already fixed in the previous push — and you are right that it affected both branches. It is also why my own check of that report showed no `## Output` section and I did not question it.

---

### On the L95 ozone, for the record

The files exist and are validated — T63/T106/T119 L95, generated 2026-08-01 for the MA L95 benchmark, all NaN-free with latitudes already S→N. They are **deliberately not packaged**: the interpolated field is near-maximum entropy (17.7M unique values of 21M), so 84 MB only reaches 48 MB at zlib-9 against 0.99 MB for the packaged L47 file, and T106/T119 are several times larger again. `reference/data_paths.md` now records their location, shapes, the purge-eligibility caveat, and the polar-interpolation trap that bites on regeneration.

duncanwp:
## Nautilus (NRP Kubernetes) support added — CI green, requesting a review pass

Everything since the last review (5 Aug) is new and unreviewed: a Nautilus site skill, two job generators, a site-profile module, and a rewrite of report retrieval.

### Validated against the live cluster, twice

| config | dev box | Nautilus | |
|---|---|---|---|
| `t63-echam-rrtmgp` | 174.6 | **175.2** sim days/hr | 0.3 % apart |
| `ma-t63-l47` (full JAM) | 131.9 | **125.3** sim days/hr | 5 %, *explained* |

Both converged, full 30 days, no NaN. The 5 % is not platform noise: `dinosaur` and `rrtmgp` are byte-identical pins, but the dev-box figure was taken on MAM4-JAX `141b2ea` (a stale local branch) while Nautilus runs `main`, two commits later — which also accounts for the 1.3 GiB lower peak memory. The provenance block turned an unexplained discrepancy into two named commits, which is the argument in #591.

### The trap the skill exists to document

`nvidia.com/a100` is a **quota bucket, not a card type**. It spans SXM4-80GB (22 nodes), 80GB-PCIe (9), **PCIE-40GB (8)** and a **1g.10gb MIG slice**. A bare request lands anywhere; the 40 GB card OOMs every L95 config and the MIG slice OOMs nearly everything, and a benchmark hopping between products yields numbers that cannot be compared. The generators select on the memory label and record which product actually ran.

Related: `priorityClassName` must be left **unset** — the namespace bans every named class at 0 pods, including `default`, so naming one gets the pod refused by quota.

### Two design points from @duncanwp, both addressed

**Portability.** Site-specific values are extracted into `sites.py` (~15 of them). The boundary is concrete: Nautilus nodes advertise `nvidia.com/gpu` — the standard name GKE/EKS use — alongside `nvidia.com/a100`, and only the latter is quota-tracked here. I deliberately did **not** write a cloud profile: untested platform code that looks authoritative is worse than none. The file records what a new profile must supply.

**`fetch_reports.py` smell.** Correct, and the fix was not a tidier reader but not needing one. Jobs now echo `report.md` to stdout, so `kubectl logs` retrieves it — matching NRP's own guidance, which reserves `kubectl cp` for small files and points at S3 for bulk. `--from-pvc` remains for expired-TTL jobs. Jobs also dump the tail of `run.log` when no rate was produced; several debug cycles here went on the pod log showing only the harness's "truncated" verdict while the traceback sat on the PVC.

### Production runs

`mkrun.py` is separate from `mkjob.py` because the two want opposite things: output kept rather than discarded, health gate always on, `backoffLimit` 20 rather than 0. Eviction tolerance is the design — a pod is not guaranteed a node for days, so each retry resumes from jcm's checkpoint rather than restarting the year.

### Known gaps, stated rather than hidden

- **`mkrun.py` is not tested end-to-end.** It dry-runs clean; I would want a short run through it before trusting a year to it.
- **T106/T119 presets cannot run on Nautilus** — their prepared boundary data exists only on the dev box. The HF mirror work unblocks this.
- **`diffrax` is pip-installed at pod startup** because MAM4-JAX needs it and the image lacks it. It works and is followed by a hard GPU check (pip could otherwise swap the CUDA jax for a CPU build, which would "work" ~100x slower and read as a slow run). But the resolved version is unpinned, so it belongs in the image behind a `jam` extra.

@codex review

PR #589, opened 2026-08-04
fix(sso): make LottMillerSso work on any horizontal layout

JOhnsonKC201:
`LottMillerSso` opened with

```python
nlev, ncols = state.temperature.shape
```

then `vmap`ped over axis 1 and transposed its output, so it only ran on a column-vectorized `(nlev, ncols)` host and raised on a whole `(nlev, nlon, nlat)` grid:

```
ValueError: too many values to unpack (expected 2)
```

The drag itself is a per-column scheme, so it has no reason to care how the host lays out the horizontal.

`ComposablePhysics(vectorize_columns=True)` is not a workaround here. `__add__` inherits that flag from its **left** operand, and a grid-based package cannot simply switch: `speedy_shortwave.get_zonal_average_fields` computes zonal averages and needs a real longitude axis, which flattening destroys. So composing `LottMillerSso` with SPEEDY was not possible in either configuration.

Fixed by flattening whatever trailing axes the host uses into one column axis, running the existing `vmap` unchanged, and restoring the shape on the way out. No branching on rank, and the column-vectorized path is unchanged.

## Test plan

New `lott_miller_host_test.py` builds the pressure and height diagnostics directly rather than going through `MoistAirColumnState`, so it exercises this term alone:

- grid `(nlev, nlon, nlat)` and column `(nlev, ncols)` hosts both return correctly shaped tendencies
- the two layouts agree column for column
- tendencies are finite

It uses the real T30 orography deliberately. With flat orography the scheme returns identically zero and every comparison passes without testing anything, so there is an explicit assertion that the drag is non-zero.

Results:

- **4 of the 5 new tests fail without the fix**; all pass with it. The survivor is `test_column_host_returns_column_shaped_tendencies`, the layout that already worked.
- `pytest jcm/physics/gravity_waves/` — 70 passed.
- `ruff check jcm/physics/gravity_waves/sso/` clean.

Verified beyond unit tests by composing `speedy_physics() + MoistAirColumnState() + LottMillerSso()` and running 450 simulated days at T31, finite throughout.

This is a shape fix only. It makes no claim that the scheme's parameters or its `nktopg`/`ntop` level indices are appropriate for any particular host resolution.

Related: #587 fixes the same class of issue in `MoistAirColumnState`. The two are independent and this PR does not depend on it.

### DIFF
--- a/jcm/physics/gravity_waves/sso/lott_miller.py
+++ b/jcm/physics/gravity_waves/sso/lott_miller.py
@@ -913,13 +913,31 @@ class LottMillerSso(PhysicsTerm):
         terrain: TerrainData,
     ) -> tuple[PhysicsTendency, dict]:
         """Compute u/v/T tendencies from Lott-Miller SSO."""
-        nlev, ncols = state.temperature.shape
+        # The drag is a per-column scheme, vmapped over columns below, so it
+        # does not care how the host lays the horizontal out. Flatten whatever
+        # trailing axes it uses into a single column axis and restore the
+        # shape on the way out. Unpacking ``nlev, ncols`` instead restricted
+        # the term to a column-vectorized host, which a grid-based physics
+        # package cannot supply: SPEEDY's shortwave computes zonal averages
+        # and so needs a real longitude axis, and ``ComposablePhysics.__add__``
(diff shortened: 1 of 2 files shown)


duncanwp:
LGTM, thanks!

Issue #591, opened 2026-08-08
Record full run provenance in the output files (versions, hardware, forcing, config, hash)


duncanwp:
## Problem

A jcm output file currently does not say what produced it. You cannot tell, from a netCDF alone, which code version ran, what boundary data it used, what precision it was in, or what hardware it ran on. That makes results hard to reproduce and — more dangerously — easy to *mis*compare, because the failure is silent.

This is not hypothetical. All of the following happened in a single benchmarking session:

**A library changed underneath a running sweep.** `jax-rrtmgp` was switched from a feature branch to `main` between config 1 and config 2 of a six-config sweep, because a PR was merged mid-run. Every install here is editable, so *the working tree is the running code*. Nothing failed; the two configs simply measured different radiation. It was caught only because the benchmark harness happens to record resolved git SHAs — without that, a 1.77x "resolution scaling" number would have been reported and believed.

**Prescribed ozone fell back to the analytic profile.** `forcing.ozone_file: auto` resolves a packaged climatology and silently degrades to RRTMGP's analytic surrogate when none matches the grid. That surrogate carries ~7.6x the tropospheric ozone column. A run completes, exits 0, produces plausible fields, and its radiation cannot be compared with anything. The only evidence is one INFO line in a log that is not kept with the output.

**Precision was invisible.** f32 vs f64 (`MAM4_JAX_ENABLE_X64`) changes both memory and speed, and above T63 f32 is required. It was set by an environment variable that appears nowhere in the output, so two files that look identical in every recorded respect can differ by 30% in cost and by more than that in what fits on a card.

**Boundary data lives on purge-eligible scratch.** Prepared inputs (JAM emissions/oxidants, non-T63 ozone) sit under `$SCRATCH`. Six months from now the file a run used may not exist, and nothing in the output records what it was — not even a checksum that would let you confirm a regenerated file matches.

**Diagnostics changed meaning across commits.** Several diagnostics in #582 were reporting step-start rather than post-physics state. Output written before and after that fix is not comparable, and nothing in either file distinguishes them.

## Proposal

Capture provenance at run start and attach it to every output file, plus a sidecar and a log line.

### What to capture

**Code**
- `jcm` version + git SHA + branch + dirty flag
- Same for every editable dependency actually imported: `jax-rrtmgp`, `dinosaur`, `mam4-jax`, `jax-cosp`, `pyses`
[...]

PR #600, opened 2026-08-11
fix(k8s): scope the run completion gate to the current attempt

duncanwp:
`run.log` is append-only **across pod restarts** — deliberately, it is what
makes the eviction-resume design debuggable — but both gates in `mkrun.py`
grepped the cumulative log, so both gave the wrong verdict on a resumed run.

**Completion (silent success).** `LAST` took the max `_dayN.nc` over *all*
attempts, so an attempt that integrated nothing inherited the previous
attempt's progress and the Job was marked Complete.

Hit for real: a benchmark resumed from another preset's leftover checkpoint
at sim-day 6 with a 4-day target, integrated 0 days, exited 0. (The
underlying "should we be resuming at all" problem is added to #591; this PR
only stops the no-op being reported as success.)

**Health (fails forever).** One bad chunk from an earlier attempt that a
later attempt recovered from failed the Job on every subsequent restart.

## Change

Capture `run.log`'s byte offset before the run; grade both gates on that
slice only. An attempt with no output is no longer ambiguous:

- log shows it resumed at/past the target → genuinely finished, pod merely
  restarted → exit 0;
- otherwise → no-op, fail loudly and name the checkpoint to inspect.

## Testing

Behaviour-tested against synthetic logs — **8/8**, including reproducing
both old misverdicts:

| scenario | expected | result |
|---|---|---|
| restart after genuine completion | 0 | pass |
| no-op resume, stale ckpt behind target | 1 | pass |
| *old* cumulative grep on the same log | (silent success) | reproduced |
| normal completion | 0 | pass |
| evicted mid-year | 1 | pass |
| health trip this attempt | 1 | pass |
| recovered-from trip in earlier attempt | 0 | pass |
| *old* cumulative health grep | (fails forever) | reproduced |

Generated manifest's script passes `bash -n`; ruff clean.

### DIFF
--- a/.claude/skills/kubernetes-jcm-runs/scripts/mkrun.py
+++ b/.claude/skills/kubernetes-jcm-runs/scripts/mkrun.py
@@ -180,14 +180,19 @@ if grep -qiE "unhealthy|NaN vars: *[1-9]" /tmp/attempt.log; then
   grep -iE "unhealthy|NaN vars: *[1-9]" /tmp/attempt.log | tail -3
   exit 1
 fi
+# `|| true` is load-bearing under `set -euo pipefail`: "no output this
+# attempt" is a state we must INSPECT, but a no-match grep exits 1 and
+# pipefail propagates that out of the command substitution, which would abort
+# the script before the empty-LAST branch below could run — marking a genuine
+# completion-restart as failed. Same for RESUMED.
 LAST=$(grep -oE "_day[0-9]+\\.nc" /tmp/attempt.log | grep -oE "[0-9]+" \\
-       | sort -n | tail -1)
+       | sort -n | tail -1 || true)
 if [ -z "$LAST" ]; then
   # No output this attempt. Distinguish the one benign case — the run was
   # already finished and the pod merely restarted — from a no-op resume,
   # which must NOT look like success.
   RESUMED=$(grep -oE "Resumed from checkpoint .* at sim-day [0-9.]+" \\
-            /tmp/attempt.log | grep -oE "[0-9.]+$" | tail -1)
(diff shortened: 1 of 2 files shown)


duncanwp:
Good catch — that was a real bug, fixed in f431cf3.

Confirmed the mechanism: the generated script runs under `set -euo pipefail`,
and "this attempt produced no output" is a state the gate has to *inspect*,
but a no-match `grep` exits 1 and pipefail propagates it out of the command
substitution, so `set -e` killed the script before the empty-`LAST` branch
could run. A pod restarting after the run had genuinely finished was marked
**failed**. The `RESUMED` lookup had the identical hazard — both pipelines
now end `|| true`, with a comment saying why it is load-bearing.

The more useful half of the lesson is why my testing missed it: I had
reimplemented the gate in a plain shell function with no `set -e`, so the
test exercised the *logic* but not the *environment the logic runs in*, and
passed while the real script aborted.

Replaced it with `scripts/mkrun_gate_test.sh`, which extracts the gate from
the actual generated manifest and sources it inside `set -euo pipefail`.
Verified it catches this defect rather than just passing: reverting the
`|| true` turns "restart after genuine completion" red and leaves the other
six green.

```
  PASS  restart after genuine completion               (rc=0)
  PASS  no-op resume from stale checkpoint             (rc=1)
  PASS  no output and no resume line                   (rc=1)
  PASS  normal completion                              (rc=0)
  PASS  evicted mid-year is incomplete                 (rc=1)
  PASS  unhealthy this attempt                         (rc=1)
  PASS  recovered-from NaN in earlier attempt          (rc=0)
```

PR #632, opened 2026-08-14
Allow physics to change atmospheric mass (surface pressure) — enables Mars CO2 cycle

sameeerkashyap:
## Summary

Allows physics parameterizations to **change the atmospheric column mass** (global-mean surface pressure). Today the dycore enforces strict dry-mass conservation, so a physics term cannot add or remove atmosphere. This blocks the **Mars CO₂ condensation cycle**, where ~25–30% of the atmosphere freezes onto the winter pole each year and the global surface pressure swings seasonally (the Viking-observed signal).

The change is opt-in and defaults to a no-op, so existing dry-mass-conserving physics is unchanged.

## What changed

1. **`PhysicsTendency`** — new scalar field `log_surface_pressure` = d(log Pₛ)/dt in 1/s. Defaults to `0.0`; appended last for backward compatibility. Non-zero only for mass-changing physics (e.g. CO₂ condensation/sublimation).
2. **`state_bridge.py`** — `physics_tendency_to_dynamics_tendency` converts that scalar into the `(0,0)` spectral mode of the log-surface-pressure tendency (a uniform global-mean mass source), instead of hardcoding it to zero.
3. **`dycore.py`** — `_conserve_global_mean_ps` is now applied against the **post-physics** state rather than the pre-physics state. This still removes spectral-dynamics mass drift, but **retains** a physics mass source/sink. For dry physics the two references are identical, so behaviour is byte-for-byte unchanged.

## Why the previous approach was insufficient

The physics tendency was already forward-Euler added to the state, but `_conserve_global_mean_ps` then reset the `[0,0,0]` mode of `log_surface_pressure` back to the **pre-step** value each step — erasing any mass change. Un-hardcoding the bridge alone does nothing; the fix is to pin conservation to the post-physics reference.

## Verification

- **Correctness:** a prescribed uniform mass sink/source lowers/raises the global-mean surface pressure by `rate × time`. Observed −0.0423 vs expected −0.0432 over 0.5 day at −1e-6 s⁻¹ (~2% match). Zero rate conserves mass (mean normalized pₛ = 1.000000).
- **No regression:** full `model_test.py` passes, including the Speedy physics runs and the `vjp`/`jvp` gradient tests (differentiability preserved). `physics_interface_test.py` and `physics_interface_hybrid_test.py` pass.

## Notes / scope

- The mass source is currently the **global mean** (the `(0,0)` mode) — sufficient for the seasonal total-mass/pressure cycle. Spatially localized condensation (higher spectral modes) is a natural follow-up.

### DIFF
--- a/jcm/dycore/dinosaur/dycore.py
+++ b/jcm/dycore/dinosaur/dycore.py
@@ -251,8 +251,10 @@ class DinosaurDycore(DynamicalCore):
             ),
             level_orders=self.diffusion.level_orders_temp,
         )
+        # NOTE: global-mean surface-pressure conservation is applied explicitly
+        # in ``step`` (against the post-physics state), not as a filter here, so
+        # a physics mass source/sink is retained. See ``step``.
         return [
-            self._conserve_global_mean_ps,
             diffuse_div,
             diffuse_vor_q,
             diffuse_temp,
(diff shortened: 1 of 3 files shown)


duncanwp:
Thanks for the PR! Can you rebase on 'dev' which is our main development branch

sameeerkashyap:
> Thanks for the PR! Can you rebase on 'dev' which is our main development branch


Ah, sorry, I am still testing this on my fork. I did not realize Claude made an upstream PR; let me rebase, test, open a proper issue, and then raise this! 

Thanks for the reply @duncanwp

PR #643, opened 2026-08-16
fix(jam): refuse a mam4-jax without configure_gas_netprod — the H2SO4 stub created sulfur without bound (#642)

### LINKED ISSUE #642: JAM sulfur/SOA budget non-closure: unbounded super-linear growth of secondary aerosol (~10× emitted sulfur; mass created in the aerosol cycling loop)
Found by the year-2 release-validation extension (#638). **Pre-existing on dev — NOT introduced by #636**: the A/B with `jam_convective_transport=false` shows identical growth.

## Symptom

`echam-jam` T63L47: interstitial accumulation-mode sulfate (and SOA) grow super-linearly and without bound. Global means:

| | year-1 mean | day-365 (60d) | year-2 last-200d |
|---|---|---|---|
| burden so4 [mg/m²] | 56 | ~73 | **418** (349 of it `m_so4_acc`) |
| burden soa [mg/m²] | 30 | ~39 | **227** |
| AOD₅₅₀ | 0.42 | 0.50 | **1.76** |

Primary species are steady the whole time (bc 0.7, poa 5, ss 1.4, du 28) and TOA has started drifting (−1.7 → −3.9 W/m²) as the spurious aerosol becomes radiatively significant.

## Budget evidence (why this is mass creation, not weak removal)

From the AeroCom flux diagnostics and column integrals (day-365 / day-730 chunks):

- `emi_so2` = 0.52 mg/m²/day (constant, sane vs ~110 Tg/yr global).
- `wet_so4` = 2.5 → 3.9 mg/m²/day — the wet sink alone is **5–7× the total emitted sulfur**, yet the burden STILL grows at ~1.3 mg/m²/day ⇒ the aerosol-phase source is ≳5 mg/m²/day, ~10× what the gas phase can supply.
- Gas reservoirs are steady and small: `g_so2` 3.36 → 3.43 mg/m², `g_h2so4` ~0.02, `g_soag` 0.14 ⇒ the excess mass is NOT coming from gas-phase depletion.
- Growth is super-linear ⇒ a **multiplicative loop**: each cycle through (aqueous production → cloud-borne → resuspension/re-injection → scavenge …) appears to gain a few % of mass. Same behaviour for SOA (soag steady, m_soa_* growing) implicates the shared aerosol-phase cycling rather than sulfur chemistry specifically.

## Suspects (in rough order)

1. Cloud-borne ↔ interstitial exchange + wetdep re-injection accounting: aerosol scavenged into the carry re-injected to interstitial without the matching debit (or credited in both phases for a step).
2. Aqueous SO2→SO4 producing cloud-borne sulfate without (fully) consuming the dissolved SO2 at the same stoichiometry — though steady `g_so2` argues the gas side balances, an intra-step double-count between the aqueous split and the exchange would look exactly like this.
3. The implicit-exponential scavenging + re-injection ledger disagreeing about which mass was removed (removed_frac vs reinjected shares).

## Suggested attack

A single-column budget harness (the SCM from the release-validation skill) with all sources off and a seeded interstitial+cloud-borne so4 pair: total sulfur (gas + interstitial + cloud-borne + cumulative deposition) must be conserved to machine precision per step; bisect terms (aqueous off, exchange off, re-injection off) until the creating term is isolated. Then add a **column sulfur-closure test** to CI and a budget-closure gate (source vs sink vs burden tendency) to the release health check so this class can't recur silently.

Release-process consequence (#638): the `echam-jam` day-730 state is NOT usable as a benchmark init state until this is fixed; the tuning of wetdep/chemistry rates is blocked behind it (no point tuning against a non-conserving budget).

duncanwp:
Fixes #642.

## Root cause

mam4-jax carries CAM `driver.F90:1248`'s "other-process" gas-chemistry stub: a hard-coded **1e-16 mol/mol/s H2SO4 production** inside `amicphys`. jcm zeroes it at term construction via `configure_gas_netprod` — but that hook only exists on mam4-jax **main** (unreleased; the newest tag `v0.2.0-beta.1` predates it), and the wrapper's `hasattr` guard **skipped silently** on hook-less cores. Result: every JAM run created ~1e-7 kg-S/m²/day per column — ~10× the emitted sulfur globally — which is exactly the unbounded super-linear so4/soa growth of #642 (condensation of the spurious H2SO4 scales with existing aerosol surface area, hence super-linear).

Found by a per-term sulfur budget audit (composed column physics, total S = gas + interstitial + cloud-borne + deposition weighed around every term call): with the stub active, `jam_mam4_jax_microphysics` creates **+6.7% of column S per 2 days** and every other term conserves to float noise. With the hook applied: **−0.1% per 2 days** (150× reduction; the residual is core per-cell numerics, see below).

## Changes

- **pyproject**: pin `mam4-jax` to main SHA `5ed4656` (has `configure_gas_netprod`; its wheel also now ships `_coag_tables.npz`). A tagged upstream release should replace the SHA pin when cut.
- **wrapper**: the soft `hasattr` guard is now a **hard refusal** — a core without the hook cannot conserve sulfur, so constructing the term raises with the #642 explanation instead of silently corrupting a run. The hook call also zeroes the SOA netprod explicitly.
- **tests**: `SulfurConservationTest` pins the contract — netprod zeroed at construction, and per-cell S-weighted closure of the core's gas+interstitial+cloud-borne tendencies (tolerance 1e-3 of gross turnover: the core's own residual is ~2.5e-4, the stub sits orders above).

## Follow-ups (noted on #642)

- Core per-cell closure residual ~2.5e-4 of gross turnover (nucleation mass-adjustment numerics) — upstream question.
- `jam_dry_deposition` ledger mismatch: destroys S at ~−3.4e-9 kg-S/m² per 2 days in the audit column (removes more from tracers than it credits to `dry_so4`) — ~50× below the old stub; separate small issue.
- SOA grew in the year runs with steady SOAG — not explained by this stub (SOA netprod was already 0); the carbon-side audit is the next hunt if a fixed-core year still shows it.
- A budget-closure gate (source vs sink vs burden tendency) goes into the release health check.

A fresh 1-year echam-jam validation run with the fixed core is in flight; burden trajectory will be posted on #642.

### DIFF
--- a/jcm/physics/aerosol/jam/microphysics/mam4_jax.py
+++ b/jcm/physics/aerosol/jam/microphysics/mam4_jax.py
@@ -195,16 +195,24 @@ class Mam4JaxMicrophysics(ModalMicrophysicsTerm):
         self._condensation_backend = str(condensation_backend)
         self._n_substeps = int(n_substeps)
 
-        # Disable the core's hard-coded "other-process" H2SO4 production stub
-        # (driver.F90:1248, 1e-16 mol/mol/s). jcm seeds the gas-phase tracers to
-        # zero and does not yet feed prognostic sulfur chemistry into the core,
-        # so that stub is a spurious, non-conservative H2SO4 source that drives
-        # runaway binary nucleation. Zero it here; revisit when jcm couples its
-        # gas-phase sulfur (jam_sulfur_gas_chemistry) into the core's gas slots.
-        # Guarded on the capability so jcm still runs against a mam4_jax build
-        # without configure_gas_netprod (the currently pinned release).
-        if hasattr(_amicphys, "configure_gas_netprod"):
-            _amicphys.configure_gas_netprod(h2so4=0.0)
+        # Disable the core's hard-coded "other-process" gas production stub
+        # (driver.F90:1248, 1e-16 mol/mol/s on H2SO4). jcm supplies its own
+        # sulfur via jam_sulfur_gas_chemistry, so the stub is a spurious
+        # sulfur source: left on, it creates ~1e-7 kg-S/m²/day per column —
+        # ~10× the emitted sulfur globally — and drove the unbounded
+        # secondary-aerosol growth of jax-gcm#642 (a previous soft hasattr
+        # guard skipped silently on cores that predate the hook, which is
+        # how a full corrupted model year shipped). A core without the hook
+        # cannot conserve sulfur, so REFUSE it rather than run.
+        if not hasattr(_amicphys, "configure_gas_netprod"):
+            raise ImportError(
+                "The installed mam4-jax has no configure_gas_netprod, so its "
+                "hard-coded H2SO4 production stub (1e-16 mol/mol/s, "
+                "driver.F90:1248) cannot be disabled and every JAM run "
+                "creates sulfur mass without bound (jax-gcm#642). Install "
+                "the pinned version: pip install 'jcm[mam4]'."
+            )
+        _amicphys.configure_gas_netprod(h2so4=0.0, soa=0.0)
 
         # Precision — applied during construction so the dycore state built
         # afterwards (in bootstrap/run) inherits it; toggling it later would
--- a/jcm/physics/aerosol/jam/microphysics/mam4_jax_test.py
+++ b/jcm/physics/aerosol/jam/microphysics/mam4_jax_test.py
@@ -263,6 +263,68 @@ class Mam4JaxAdapterTest(unittest.TestCase):
 
 
 @pytest.mark.slow
+class SulfurConservationTest(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


duncanwp:
Validation year complete — burden trajectories on #642: sulfate and SOA both bounded with seasonal turnover (so4 peaks 27 mg/m² day ~275 then declines; the broken core reached 418 in year 2). TOA +0.3 W/m², NaN-free. The audit, the conservation test, and the year all agree the creator is gone.

duncanwp:
Per review: the SulfurConservationTest unit tests are removed — the hard raise-on-missing-hook stays as the only guard. Upstream has no new tag yet (release requested); the pyproject pin stays on the main SHA with a comment to flip to the version number the moment the tag is cut.

PR #657, opened 2026-08-19
fix(deps): flax floor is >=0.12.1 — the release that ADDED nnx get_value()

### LINKED ISSUE #655: `nnx.Variable.get_value()` no longer exists in flax>=0.12.0, breaking 95 call sites
## Summary

With the currently pinned `flax>=0.12.0`, `nnx.Variable` (and `nnx.Param`, which subclasses it) no longer exposes a `.get_value()` method — only a `.value` property (plus `raw_value` and `create_value`) remains. `jax-gcm` code still calls `.get_value()` throughout, e.g. `jcm/physics/speedy/speedy_terms.py:172` does `self._speedy_coords.get_value()`. Because `flax.nnx.Variable.__getattr__` proxies unknown attribute lookups to the wrapped raw value, most `.get_value()` calls only look like they work if the wrapped object happens to define its own `get_value` method; when it doesn't (e.g. `SpeedyCoords`), it raises `AttributeError: 'SpeedyCoords' object has no attribute 'get_value'` immediately.

## Environment

* flax: 0.12.0 (installed; `requirements.txt` pins `flax>=0.12.0`)
* Python: 3.13.7

## Root cause

Flax's NNX `Variable` API removed/renamed `.get_value()` sometime between April 2026 and the `0.12.0` release. This repo's `requirements.txt` was bumped to `flax>=0.12.0` in `3aba895` ("refactor(dycore)!: semi-Lagrangian transport only, remove the Eulerian path", 2026-08-13) — an unrelated dycore change that happened to raise the flax lower bound — but none of the `.get_value()` call sites were updated to match the new API. The `.get_value()` call site in `speedy_terms.py` itself was introduced earlier, in `567ecb3` ("Add composable physics infrastructure (#206) (#429)", 2026-04-27), when `.get_value()` was still valid Flax NNX API. As a result, the minimum flax version this package claims to support (`0.12.0`) is actually incompatible with the package's own code.

## Scope

`grep -ron "get_value()" jcm/ | wc -l` reports 95 occurrences across 41 files, including but not limited to: `jcm/physics/speedy/speedy_terms.py`, `jcm/physics/convection/tracer_transport.py`, `jcm/physics/convection/betts_miller/betts_miller_terms.py`, `jcm/physics/convection/tiedtke_nordeng/tiedtke_nordeng.py`, `jcm/physics/vertical_diffusion/tracer_diffusion.py`, `jcm/physics/vertical_diffusion/tte_tke/vertical_diffusion.py`, `jcm/physics/gravity_waves/spectral/term.py`, `jcm/physics/gravity_waves/sso/lott_miller.py`, `jcm/physics/gravity_waves/hines/hines.py`, `jcm/physics/gravity_waves/simple/simple_gwd.py`, `jcm/physics/dissipation/upper_sponge.py`, `jcm/physics/dissipation/upper_temperature_relaxation.py`, `jcm/physics/aerosol/macv2_sp.py` and the rest of `jcm/physics/aerosol/jam/\*\*`, `jcm/physics/radiation/rrtmgp.py`, `jcm/physics/radiation/nn_emulator_scheme.py`, `jcm/physics/radiation/grey_two_stream/radiation_scheme.py`, `jcm/physics/clouds/echam_1m.py`, `jcm/physics/clouds/sundqvist.py`, `jcm/physics/clouds/lohmann_2m/scheme.py`, `jcm/physics/chemistry/simple_chemistry.py`, `jcm/physics/diagnostics/moist_air_state.py`, `jcm/physics/diagnostics/state_sampler.py`, `jcm/physics/held_suarez/held_suarez_physics.py`, `jcm/physics/surface/echam/surface_physics.py`, `jcm/physics/forcing/echam_boundary_conditions.py`, `jcm/rce_test.py` and `jcm/physics/gravity_waves/spectral/term_test.py` / `jcm/physics/radiation/rrtmgp_test.py`.

## Minimal reproduction

```python
from jcm.physics.speedy.speedy_coords import get_speedy_coords
import flax.nnx as nnx

coords = get_speedy_coords()
v = nnx.Variable(coords)
v.get_value()  # AttributeError: 'SpeedyCoords' object has no attribute 'get_value'
```

## Where this surfaces in practice

Any run that constructs and steps a `jcm.model.Model` using the composable Speedy physics terms hits this during initialization: `Model.run_from_state_with_carry` → `Model._build_initial_physics_carry` → `ComposablePhysics.get_empty_data` → `ComposablePhysics.compute_tendencies` → `ComposablePhysics._compute_tendencies_3d` → `SpeedyFlags.__call__` → `SpeedyTermBase._build_data`, ending in the `AttributeError` above at `jcm/physics/speedy/speedy_terms.py:172`.

## Suggested fixes (pick one)

1. Repo-wide: replace `.get_value()` with `.value` across all 95 call sites to match the current Flax NNX API, keeping `flax>=0.12.0`.
2. Pin `flax` in `requirements.txt` back to the last version that still supported `nnx.Variable.get_value()`, reverting the incidental floor bump from `3aba895` until the call sites are migrated.

## Impact

Any workflow that constructs a `jcm.model.Model` with composable Speedy physics fails at initialization time under the currently pinned flax version.

duncanwp:
Closes #655.

## The report is right about the symptom, backwards about the cause

#655 says `nnx.Variable.get_value()` "no longer exists in flax>=0.12.0" and was removed "sometime between April 2026 and the 0.12.0 release". The opposite is true: **`get_value()` was ADDED in 0.12.1**. Verified by unpacking every wheel and grepping `flax/nnx/variablelib.py`:

| flax | `def get_value` |
|---|---|
| 0.10.0 – 0.12.0 | **absent** |
| 0.12.1 – 0.12.9 | present |

The claimed timeline is also impossible: 0.12.0 shipped **2025-09-25**, seven months *before* 567ecb3 (2026-04-27), the commit it credits with introducing the call sites "when `.get_value()` was still valid". At that commit flax was 0.12.7, which has it. The code has always been on the new API.

So the bug is not 94 stale call sites — it is that the `flax>=0.12.0` floor (incidentally raised by 3aba895) advertises a version jcm cannot run on. **It is off by one patch release.**

## Why neither suggested fix in the issue is right

- **"Replace `.get_value()` with `.value`"** would migrate onto the API flax is *removing*. In 0.12.1+ `.value` is a deprecation shim whose own warning reads: `"'.value' access is now deprecated. ... For other Variable types use: variable.get_value()"`.
- **"Pin flax back"** is impossible — no earlier release ever had `get_value()`.

## Why no dev box or CI ever hit it

Everything resolves to the newest flax (this dev box: 0.12.6; PyPI latest: 0.12.9). You only reproduce by installing the floor exactly, which is what the reporter did. Separately, 0.12.0 cannot even be imported against a current jax — it does `from jax._src.core import mutable_array`, long gone (`ImportError` on jax 0.10.2). So nothing of value is lost by raising the floor.

## What is in this PR

Two separable commits:

1. **`requirements.txt`: `flax>=0.12.0` → `>=0.12.1`**, with a comment recording why the floor is where it is and an explicit "do not "fix" this by moving to `.value`" note, so the next person reading #655 doesn't undo it. (`pyproject.toml` takes `dependencies` from this file via `dynamic`, so this is the installed floor.)

2. **Five test call sites off the deprecated `.value`** in `jcm/physics/echam/parameters_test.py` and `jcm/runners_test.py`. These were the *only* remaining `.value` readers in the tree — the ~94 production sites were already correct — so they were the sole source of flax deprecation noise and the only thing that breaks outright when `.value` is deleted.

For (2) I swept the category rather than the spellings I'd been shown: enumerated all 35 attributes bound to an `nnx.Param`/`Variable`/`List` and checked each for `.value` reads or writes. None remain; the only `.value` tokens left under `jcm/` are `ast.*` node attributes in `requires_audit_test.py`.

## Verification

- Full fast suite under a side-installed **flax 0.12.1** (the new floor): **1769 passed**.
- Both touched modules on flax **0.12.1 and 0.12.6** with `-W error::DeprecationWarning:flax.*`: **102 passed**, no warning raised.
- Reproduced the original `AttributeError` on 0.12.0 and confirmed it is gone on 0.12.1.
- `ruff check .` clean.

**Local suite caveat, so CI is read correctly:** this dev box has the released dinosaur, but `dev` now requires the `semi-lagrangian` fork, so 120 tests fail locally with `RuntimeError: the installed dinosaur has no semi-Lagrangian transport`. I ran the full suite on clean `origin/dev` and on this branch and diffed the failure sets — **identical, 120 on both, zero regressions and zero fixes.** CI installs from `requirements.txt` so it gets the fork and should be green.

### DIFF
--- a/jcm/physics/echam/parameters_test.py
+++ b/jcm/physics/echam/parameters_test.py
@@ -38,7 +38,7 @@ def test_echam_physics_default_kwargs():
     convection_term = next(
         t for t in physics.terms if t.category == "convection"
     )
-    assert abs(float(convection_term.params.value.entrpen) - 1.0e-4) < 1e-7
+    assert abs(float(convection_term.params.get_value().entrpen) - 1.0e-4) < 1e-7
 
 
 def test_echam_physics_per_scheme_kwargs():
@@ -50,13 +50,13 @@ def test_echam_physics_per_scheme_kwargs():
     convection_term = next(
         t for t in physics.terms if t.category == "convection"
     )
-    assert abs(float(convection_term.params.value.entrpen) - 5.0e-4) < 1e-7
+    assert abs(float(convection_term.params.get_value().entrpen) - 5.0e-4) < 1e-7
 
     # Untouched terms still see their defaults.
     cloud_fraction_term = next(
         t for t in physics.terms if t.category == "cloud_fraction"
     )
-    assert abs(float(cloud_fraction_term.params.value.crt) - 0.75) < 1e-7
+    assert abs(float(cloud_fraction_term.params.get_value().crt) - 0.75) < 1e-7
 
 
 def test_physics_terms_compute_tendencies():
--- a/jcm/runners_test.py
+++ b/jcm/runners_test.py
@@ -139,7 +139,7 @@ class TestBuilders(unittest.TestCase):
             t for t in physics.terms if t.category == "convection"
         )
         self.assertAlmostEqual(
-            float(convection_term.params.value.entrpen), 4e-4,
+            float(convection_term.params.get_value().entrpen), 4e-4,
         )
 
     def test_build_physics_curated_preset(self):
@@ -154,7 +154,7 @@ class TestBuilders(unittest.TestCase):
             t for t in physics.terms if t.category == "convection"
         )
         self.assertAlmostEqual(
-            float(convection_term.params.value.entrpen), 4e-4,
+            float(convection_term.params.get_value().entrpen), 4e-4,
         )
 
     def test_build_physics_swap_radiation_via_preset(self):
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,16 @@
 # ``dinosaur>=X.Y`` — a VCS requirement cannot be published to PyPI and
 # makes the wheel uninstallable for downstream users (see #577).
 dinosaur @ git+https://github.com/shoyer/dinosaur@semi-lagrangian
-flax>=0.12.0
+# Floor is 0.12.1: that release ADDED `nnx.Variable.get_value()` and deprecated
+# the `.value` property in its favour. jcm reads every nnx.Param/nnx.Variable
+# through `.get_value()` (~94 sites under jcm/physics/**), so on exactly 0.12.0
+# the lookup misses the class and falls through `Variable.__getattr__` to the
+# *wrapped* object — `AttributeError: 'SpeedyCoords' object has no attribute
+# 'get_value'` as soon as a Model is built (#655). Do NOT "fix" that by moving
+# to `.value`: that is the deprecated direction, and no flax ever released has
+# both. (0.12.0 is unusable here anyway — it imports the long-gone private
+# `jax._src.core.mutable_array`.)
+flax>=0.12.1
 jax-datetime
 tree-math
 hydra-core


duncanwp:
@codex review

Issue #390, opened 2026-03-04
Better documentation the surface flux parameterization and its fields


eldavenport:
There are often multiples of each field, one for land, one for ocean and one combined. It's not always clear which is which and we have to figure out how it works every time we look into it.

Issue #328, opened 2025-12-12
Clean up SpeedyPhysics surface fluxes logic


jvmadan:
It's a very confusing file to navigate at the moment, we should clarify the dependencies by reorganizing code and/or adding comments

PR #659, opened 2026-08-19
Publish SPEEDY surface fluxes as flat 2D maps (#645, #328, #390)

### LINKED ISSUE #645: `hfluxn` in `speedy_surface_flux.py` only has 2 channels — missing the land/sea weighted-average channel that all sibling flux arrays compute
# Summary

`get_surface_fluxes` in [`jcm/physics/surface/speedy_surface_flux.py`](jcm/physics/surface/speedy_surface_flux.py) computes five per-surface-type flux arrays — `ustr`, `vstr`, `shf`, `evap`, `rlus` — each shaped `(ix, il, 3)`, where channel 0 is land, channel 1 is sea, and channel 2 is a land/sea-fraction-weighted average computed at the end of the function:

```python
weighted_average = lambda var: var[:, :, 1] + fmask * (var[:, :, 0] - var[:, :, 1])

ustr = ustr.at[:, :, 2].set(weighted_average(ustr))
vstr = vstr.at[:, :, 2].set(weighted_average(vstr))
shf  = shf.at[:, :, 2].set(weighted_average(shf))
evap = evap.at[:, :, 2].set(weighted_average(evap))
rlus = rlus.at[:, :, 2].set(weighted_average(rlus))
```

`hfluxn` (net downward heat flux) is initialized with only 2 channels and is never included in this weighted-average block:

```python
hfluxn = jnp.zeros((ix, il, 2))   # land=0, sea=1 — no channel 2
...
hfluxn = hfluxn.at[:, :, 0].set(...)   # land
...
hfluxn = hfluxn.at[:, :, 1].set(...)   # sea
# no hfluxn[:, :, 2] weighted average — channel 2 doesn't exist
```

`SurfaceFluxData.zeros`/`.ones` in `jcm/physics/speedy/physics_data.py` bake in the same 2-channel default for `hfluxn`, so the shape mismatch is consistent throughout the module — it's not a one-off typo at the call site.

## Root cause

`hfluxn` was never given the same 3-channel treatment as `ustr`/`vstr`/`shf`/`evap`/`rlus`, and the weighted-average block at the end of `get_surface_fluxes` omits it.

# Proposed fix

1. In `get_surface_fluxes`, initialize `hfluxn = jnp.zeros((ix, il, 3))` (matching its siblings) and add `hfluxn = hfluxn.at[:, :, 2].set(weighted_average(hfluxn))` alongside the other four weighted-average assignments.
2. Update `SurfaceFluxData.zeros`/`.ones` defaults in `physics_data.py` to shape `(..., 3)` for `hfluxn`, consistent with `ustr`/`vstr`/`shf`/`evap`/`rlus`.
3. Consider adding a shape assertion (or switching indexing call sites to raise on out-of-bounds rather than relying on JAX's default clamp mode) so a future channel-count mismatch fails loudly instead of silently duplicating data.

# Where

- `jcm/physics/surface/speedy_surface_flux.py` (confirmed present on both `main` and `dev`)
- `jcm/physics/speedy/physics_data.py` (`SurfaceFluxData.zeros`/`.ones`)

# Reported via

Originally surfaced as [climate-analytics-lab/jax-esm#99](https://github.com/climate-analytics-lab/jax-esm/issues/99), which consumes `hfluxn[:, :, 2]` as `total_heat_flux`.

duncanwp:
Closes #645. Takes the linked #328 (*"confusing file … no clear separation"*) and #390 (*"not always clear which is which"*) with it.

## What was wrong

`get_surface_fluxes` carried land, sea and area-weighted values of five fluxes in a trailing channel axis. `hfluxn` had only two channels, so `hfluxn[:, :, 2]` clamped to the sea value rather than raising — and jax-esm's coupler consumed exactly that as its grid-mean heat flux (jax-esm#99). The docs example did `jnp.sum(hfluxn, axis=2)`, adding land and sea together. Both are the same failure: a positional axis where the position carries meaning nothing enforces.

## What this does

Land and sea are now named local values inside the module, and every published field is a 2D map — so `[:, :, 2]` raises instead of clamping. `surface_flux.shf.2` becomes `surface_flux.shf`; the `.0`/`.1` variables are gone.

**The merged values are bit-identical.** I captured the pre-refactor output for nine configurations (the five unit-test cases, pure land, aquaplanet, `fhum0=0.7`, `lskineb=False`) and the refactor reproduces every published field, every tendency, exactly — no ulp drift.

## Design decisions worth your attention

**1. `hfluxn` stays per surface type, as `hfluxn_land` / `hfluxn_sea`.** Dropping the unmerged dimensions entirely would have broken the coupler: `jem/components/jcm_component.py` reads `hfluxn[:, :, 0]` and `[:, :, 1]` to drive its slab **land** and slab **ocean** separately. Those are load-bearing, not diagnostics — `hfluxn` is the flux into the surface *medium*, so a surface component needs its own tile's value and the grid mean is the wrong forcing. The line I drew: fields the atmosphere feels are grid means; fields a surface component integrates are per-tile. `hfluxn` is the only one in the second category today. If Veros ever needs ocean-only wind stress, that argues for `ustr_sea` on the same principle rather than bringing the axis back.

A reviewer argued the deeper fix is to publish whole tile structs (`surface_flux.land.*` / `surface_flux.sea.*`) instead of a name-mangled pair. That is a fair point and I did not do it — worth deciding before a second field needs a tile.

**2. Output variables are renamed.** Unavoidable given the ask, but it is a breaking change for any analysis script keyed on `surface_flux.shf.2`. No in-repo tool reads these names (checked `tools/`, the release-validation health checks and the AeroCom mapper); three demo notebooks did and are fixed. Recorded in the release notes.

**3. SPEEDY checkpoints are invalidated.** `SurfaceFluxData` is part of the physics carry, so its shape change breaks every existing SPEEDY restart file — including `bundles/t31_l8/init_states/speedy_year1.msgpack` on the data mirror. `load_checkpoint` now names the file and the reason instead of surfacing flax's bare leaf-count error. **I'll regenerate and re-upload the mirror state once this merges** (SPEEDY T31 is minutes). Long ECHAM/JAM runs are unaffected — different physics package.

## Bonus: the units table was dead plumbing (#390)

`ComposablePhysics.UNITS_TABLE_CSV_PATH` was `None`, so `jcm/physics/speedy/units_table.csv` **never reached model output** — all 67 SPEEDY variables shipped with no units and no description, which is a large part of why "it's not always clear which is which". Terms now declare their own table and the container gathers them; a test fails if a diagnostic arrives without a row. ECHAM/ICON terms now have a place to put theirs.

## Review

Ran `/code-review high` locally before opening this (8 finder angles). It caught a real regression I had introduced: the land branch's zero arm took its dtype from the state while the computed arm follows the forcing, so a float64 forcing with a float32 state — precisely what the coupler passes — hit `cond branches must have equal output types`. Fixed via `jax.eval_shape`, with a test that fails without it. It also caught that every regression case ran at `fmask=0.5`, where the weighting is symmetric and a swapped merge order is undetectable; there is now a pure-land case.

Tests: 238 passed across the surface, SPEEDY, radiation, cloud, composable-physics, checkpoint and utils suites; `ruff` clean.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -28,6 +28,33 @@ Unreleased — transient AMIP forcing
 Unreleased — issue-backlog tidy-up
 ----------------------------------
 
+- **SPEEDY surface fluxes are published as flat 2D maps** (#645, #328,
+  #390). ``ustr``, ``vstr``, ``shf``, ``evap`` and ``rlus`` carried land,
+  sea and area-weighted values in a trailing channel axis, so output files
+  held ``surface_flux.shf.0/.1/.2``. Only the weighted grid mean ever
+  reached the atmosphere, and ``hfluxn`` had no channel for it at all —
+  ``hfluxn[:, :, 2]`` clamped to the sea value instead of raising, which a
+  coupled run consumed as its grid-mean heat flux. Each of these is now a
+  single 2D variable holding the grid mean: **``surface_flux.shf.2``
+  becomes ``surface_flux.shf``**, and the per-surface ``.0``/``.1``
+  variables are gone. The merged values themselves are unchanged.
+
+  ``hfluxn`` is the exception: it drives a surface component's own
+  temperature, so a coupled land or ocean model needs its own tile's
+  value rather than the grid mean. It is published as
+  ``surface_flux.hfluxn_land`` and ``surface_flux.hfluxn_sea`` beside the
+  merged ``surface_flux.hfluxn``.
+
+  **Existing SPEEDY checkpoints will not load**, including the ``t31_l8``
+  init state on the data mirror — the diagnostic struct changed shape.
+  ``load_checkpoint`` now names the file and the reason instead of
+  surfacing a bare leaf-count error. Regenerate the state, or pin the
+  previous release.
+- **Every SPEEDY output variable carries units and a description** (#390).
+  The SPEEDY units table never reached output at all: the composable
+  physics container defined no table, so all 67 variables shipped bare.
+  Terms now declare their own table and the container gathers them, with
+  a test that fails if a new diagnostic arrives without a row.
 - **ECHAM hyperdiffusion now covers every hybrid grid, including L95**
   (#579). ``diffusion.kind=auto`` matched on ``layers == 47``, so the L95
   middle-atmosphere grids — which exist precisely to resolve the
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -37,13 +37,12 @@ Unreleased — issue-backlog tidy-up
   coupled run consumed as its grid-mean heat flux. Each of these is now a
   single 2D variable holding the grid mean: **``surface_flux.shf.2``
   becomes ``surface_flux.shf``**, and the per-surface ``.0``/``.1``
-  variables are gone. The merged values themselves are unchanged.
+  variables are gone. ``hfluxn`` gains the grid mean it never had. The
+  merged values themselves are unchanged.
 
-  ``hfluxn`` is the exception: it drives a surface component's own
-  temperature, so a coupled land or ocean model needs its own tile's
-  value rather than the grid mean. It is published as
-  ``surface_flux.hfluxn_land`` and ``surface_flux.hfluxn_sea`` beside the
-  merged ``surface_flux.hfluxn``.
+  A coupled surface model that was reading per-tile heat fluxes out of
+  ``hfluxn``'s channels now needs them from its own land/ocean components
+  rather than from the atmosphere's diagnostics.
 
   **Existing SPEEDY checkpoints will not load**, including the ``t31_l8``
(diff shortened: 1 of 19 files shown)


duncanwp:
Dropped `hfluxn_land`/`hfluxn_sea` as requested (a38e6cb) — every published field is now the fmask-weighted grid mean with no exceptions, so there is one semantic in `SurfaceFluxData` and nothing to get wrong about which fields are per-tile. JEM will need its per-tile heat fluxes from its own land/ocean components instead.

The weighting is still pinned without publishing the tiles: `test_fluxes_are_linear_in_the_land_fraction` runs fmask 0, 0.5 and 1 and requires every flux to sit on the straight line between the endpoints, having first asserted the endpoints differ so the check can't go vacuous.

Re-verified after the change: still **bit-identical** to `dev` on every published field across all nine configurations.

Issue #692, opened 2026-08-19
Flaky: pyses_dycore_test.py::TestCoupledEchamSmoke::test_model_drives_float32_physics segfaults intermittently under xdist


duncanwp:
`jcm/dycore/pyses/pyses_dycore_test.py::TestCoupledEchamSmoke::test_model_drives_float32_physics` intermittently kills its xdist worker with `Fatal Python error: Segmentation fault` during the slow gate on a Derecho `develop` node. The faulthandler dump shows only threadpool and xdist frames plus `model.run -> resume -> run_from_state_with_carry`, i.e. the fault is inside compiled/native code, not Python.

## Observations

Five full slow-gate runs on a `develop` node (16 cpus / 200 GB, `-n 4`, `JAX_PLATFORMS=cpu`), two worktrees:

| worktree | without `--cov` | with `--cov` |
|---|---|---|
| `origin/dev` @ f6d1bcd | 93 passed | 93 passed |
| a branch touching only `jcm/physics/clouds/lohmann_2m/` | 93 passed | **segfault**, **segfault** |

The branch correlation is spurious: `echam_physics(radiation_scheme="grey")`, which is what this test builds, resolves `cloud_scheme` to its `"1m"` default, so the stack contains `echam_1m_microphysics` and no Lohmann 2M term at all — the branch's code is never constructed, traced or executed by this test. Coverage is not the variable either (dev passes with it).

Run in isolation the test passes on both worktrees in near-identical time (59.80 s vs 59.56 s), so it needs the full-suite context (memory pressure, or accumulated XLA JIT mappings — cf. the `LLVM ERROR: Unable to allocate section memory!` note in the jcm-local-ci skill, which is the same class of failure at a different threshold).

A different pySES test, `TestPysesDycoreProtocol::test_forced_step_advances_and_stays_finite`, also failed in a (separately invalid) login-node run, so the instability may not be specific to this one test.

## Why it matters

It fails the PR gate non-deterministically, and the segfault takes the coverage data with it ("workers failed to return coverage data" → total coverage reported as ~29 %, tripping `--cov-fail-under=80` on a suite that otherwise passes). That turns one flake into two red signals and makes the gate hard to read.

## Suggested next steps

- Reproduce under `faulthandler` with `-p no:randomly` and a fixed `-n` to see whether it is order-dependent.
- Check whether the pySES native extension is being initialised in more than one worker simultaneously, and whether it is fork-safe.
[...]

Issue #689, opened 2026-08-19
2M: a negative qc/qi undershoot is repaired by condensing vapour, a sign-definite spurious heat + moisture sink


duncanwp:
`cloud_microphysics_2m` clips the number tracers at entry, explicitly because the dycore can ring negative:

```python
qnc = jnp.clip(qnc, 0.0, _cdnc_max_phys_per_m3 * inv_rho_safe)   # scheme.py:121
qni = jnp.clip(qni, 0.0, params.icemax * inv_rho_safe)
```

The **mass** tracers `qc`/`qi` are not clipped. Since #662 finding 6 they are passed straight to the ledger as `liq_mmr_prev`/`ice_mmr_prev`, so a negative value reaches the negative-mass guard:

```python
zdxlcor = jnp.where(liq_mmr_next < ccwmin, -ztmst_rcp * liq_mmr_next, 0.0)
liq_tendency = liq_tendency + zdxlcor
specific_humidity_tendency = specific_humidity_tendency - zdxlcor
temp_tendency = temp_tendency + lvdcp * zdxlcor
```

With `qc = -1e-6` and cf = 0.6 that is `dqc·dt = +1e-6`, `dq·dt = -1e-6`, `dT·dt = +0.0025 K` per 1800 s step — about **+0.12 K/day** of warming and a matching vapour sink, wherever the undershoot occurs. Previously `liq_mmr_prev` was the in-cloud value, which `update_in_cloud_water` clamps at `>= 0`, so the guard saw 0 and did nothing.

## This is not obviously a bug

The repair is thermodynamically consistent — it removes unphysical negative liquid and pays for it out of vapour with the correct latent heat — and it is what ECHAM's `zdxlcor` is for (ECHAM's `pxlm1` is the real tracer, negatives included). Clipping at entry instead would create mass with no ledger entry, which is worse for conservation.

The concern is that it is **sign-definite**: every undershoot becomes warming and drying, never the reverse, so it biases rather than just adding noise.

## What to decide

1. Whether the semi-Lagrangian transport (#521, which replaced the Eulerian core precisely because it rang negative on sharp sources) leaves undershoots large enough to matter. If they are at the 1e-12 level this is noise; if 1e-6 occurs it is a real heat source.
2. If it matters: diagnose the repair as a budget term so it is visible, rather than silently folding it into `dtedt`/`dqdt`. A column-integrated `negative_mass_repair` diagnostic would make it measurable in a real run.

[...]

Issue #688, opened 2026-08-19
2M: the negative-mass guard now clears thin stratiform cloud (grid-mean condensate < ccwmin) — check ccwmin is the right threshold


duncanwp:
#662 finding 6 changed `update_tendencies_and_important_vars`'s `ice_mmr_prev`/`liq_mmr_prev` from the post-process **in-cloud** condensate to the **grid-mean** `qi`/`qc`. That is ECHAM's convention (`pxim1`/`pxlm1` are grid-mean, and so is every increment the ledger adds to them), and it is what makes the guard bound the host's actual post-step condensate. But it changes when the guard fires, by a factor of 1/cf.

```python
ll_ice_neg = ice_mmr_next < params.ccwmin      # assembly.py:204, ccwmin = 1e-7 kg/kg
zdxicor    = jnp.where(ll_ice_neg, -ztmst_rcp * ice_mmr_next, 0.0)
```

Note it triggers on small **positive** values, not just negatives, and it does four things at once: sublimates/evaporates the condensate to vapour with the matching latent heat, zeroes the CDNC/ICNC tracer tendency, and (with the write-back) clears the cloud fraction.

## The case to check

Optically thin cirrus: cf = 0.05, in-cloud IWC 1e-6 kg/kg — an ordinary sub-visible cirrus — is grid-mean 5e-8 < ccwmin, so it is removed in one step along with its crystals. Under the previous in-cloud comparison, 1e-6 > ccwmin and it survived. The same mechanism clears any cf ≤ 0.1 stratocumulus whose grid-mean LWC falls below 1e-7.

Freshly nucleated cirrus is the sharpest version: the #494 deposition-nucleation hook can set ICNC before the crystals have grown enough mass to clear a grid-mean 1e-7, in which case the cell cannot bootstrap — each step's new crystals are deleted before they can grow.

## What to decide

Whether `ccwmin = 1e-7` is the right *grid-mean* threshold. It was previously being applied to in-cloud magnitudes, so it has never been tuned against grid-mean values in this scheme. ECHAM uses `ccwmin = 1e-7` against grid-mean `zxlp1`/`zxip1`, so the pairing is reference-correct; the question is whether jcm's thin-cirrus and #494 nucleation paths need a lower value or a cf-aware floor.

Suggested evidence: a T63 run with IWP/cirrus-cover diagnostics before and after #662, and a single-column test that nucleates cirrus from zero and checks it survives to a radiatively meaningful IWC.

[...]

Issue #687, opened 2026-08-19
clouds.cloud_fraction means different things under cloud_scheme='1m' and '2m'


duncanwp:
After #662, `Lohmann2MMicrophysics` writes the post-microphysics cover back into the `clouds` carry (ECHAM's `paclc` write-back: cells emptied of both condensates, or driven below `clc_min`, stop being cloudy). `Echam1MMicrophysics.__call__` does **not** — its `clouds.copy(...)` leaves `cloud_fraction` at the value `SundqvistCloudFraction` diagnosed.

So the same field now means:

- **1m**: the RH-diagnosed cover, before microphysics.
- **2m**: the cover after microphysics has cleared emptied cells.

Every shared consumer of the carry switches semantics with the cloud scheme, with nothing in the code or docstrings recording it:

- `aerosol/jam/cloud_borne.py` — the cloud-borne/interstitial partition target
- `aerosol/jam/aqueous.py` — the `cloud_fraction > 0` gate on in-cloud SO4 production
- `aerosol/jam/wetdep_term.py` — `cf_clip` for in-cloud scavenging and `clear_fraction = 1 - cf` for below-cloud impaction
- `diagnostics/cosp_cloudsat.py`
- `diagnostics/aerocom.py` — the exported `clt`

## Why it matters

A 1M-vs-2M A/B is the standard way this repo attributes a cloud or aerosol-forcing change. That comparison now differs by a *definition* as well as by physics, and the difference will read as a physics signal. The aerosol side is the sharper half: 2M columns get systematically less in-cloud scavenging in cells the microphysics emptied, which is exactly where the precipitation formation rate is largest.

COSP is the one place the 2M behaviour is clearly the better of the two — it pairs the cover with the post-microphysics condensate — which makes 1M the inconsistent one there.

## Options

1. Make the 1M scheme write its post-microphysics cover back too, so both agree. Needs checking that `echam_1m` has an equivalent of the `paclc` write-back (ECHAM's 1M `cloud` routine does).
2. Or publish the post-microphysics cover under a distinct key and leave `cloud_fraction` as the diagnosed cover for everyone.

Option 1 is closer to both references; option 2 is the smaller change and lets each consumer choose. Either way the choice should be documented on `CloudData.cloud_fraction`.

Related: `tools/release_validation/health.py` gates `clouds.cloud_fraction.max('level')` against a 0.4–0.8 band that was calibrated on the pre-microphysics cover — worth re-checking after this lands either way.

[...]

PR #693, opened 2026-08-19
fix(clouds): route every 2M process transfer into the tendency ledger (#662)

duncanwp:
Findings 1, 2, 3, 4 and 6 of #662, plus the column enthalpy-conservation test
that gates them. **Finding 5 is deliberately not here** — see the last section.

## Root cause

Five independent symptoms, one shape: a process routine computes a transfer,
applies it to its *local in-cloud* arrays, and the grid-mean tendency ledger in
`assembly.py` never hears about it — or hears only half of it. Water stays
conserved in three of the five, which is why nothing caught them.

The ledger (`update_tendencies_and_important_vars`) is ECHAM's: it starts from
the grid-mean condensate the step began with and adds grid-mean per-step
increments. Every fix below is "convert the transfer to grid-mean and put it in
the accumulator the ledger reads".

### 1. WBF applied its latent heat and discarded the mass

`WBF_process` computes the Bergeron transfer once and returns it three ways: a
liquid debit, an ice credit, and the matching fusion warming. The sweep kept
only the warming (`_liq_tend_wbf`, `_ice_tend_wbf` were dropped on the floor).
Supercooled liquid was heated as though it had glaciated but never left `qc`.
All three now seed the three INOUT accumulators.

### 2. Cloud ice above 0 °C melted twice

`melting_snow_and_ice` reconstructs the ice available to melt as
`max(pxim1 + dt·pxite, 0)`. It was handed the **pre**-sedimentation ice with
`pxite = 0`, so ice that sedimentation had already moved into the falling flux
melted a second time — and the routine's own ice debit was discarded on top.
The sedimentation tendency is now threaded *through* the routine (which is how
ECHAM keeps the two sinks from claiming the same mass) and its output tendency
seeds the ledger.

The sediment → melt *ordering* is unchanged; it matches MG/PUMAS. The bug was
the discard, not the order.

### 3. Heterogeneous freezing moved crystal number but not mass

`frozen_mass` was applied to `in_cloud_ice_het` / `in_cloud_liquid_het` but
never added to the freezing accumulator, so raising `ice_nuclei` created
crystals with zero mass and destroyed droplets with zero mass. Water conserves
perfectly — which is why this one hides from every budget check — but it meant
the JAM aerosol → ice coupling of #494 moved number and nothing else, while the
effective radii were computed on liquid the tracers never lost.

A single grid-mean `freezing_rate` (ECHAM `pfrl`) now carries both the
homogeneous and the heterogeneous leg; the ledger converts it to fusion heat
automatically.

### 4. `cloud_fraction_final` was discarded — cherry-picked, then clipped

Cherry-picked from @duncanwp's `fix/662-cloud-fraction-final` (`c29a820`) —
thanks. Cells the microphysics has just emptied no longer keep a non-zero cloud
fraction for radiation and the aerosol partition to read.

**One change on top of it.** The published cover is now clipped to the incoming
cover: the microphysics may *remove* cloud, never create it. `update_in_cloud_water`
has an upward branch — a clear cell with any condensation gets
`cf = clip(RH, 0.01, 1)` — which is a second, RH-based cover closure competing
with `SundqvistCloudFraction`. Publishing it raw substituted that closure. Measured:
an ice-supersaturated column at 5 hPa, **above `cloud_top_pressure_pa`, which
Sundqvist deliberately reports as cloud-free for exactly this reason**, came back
**overcast (cf = 1.0)** on ~1e-6 kg/kg of ice — and COSP, AeroCom and the JAM
cloud-borne / aqueous / wetdep terms all read that field.

### 6. The negative-mass guard was fed in-cloud values

`ice_mmr_prev` / `liq_mmr_prev` are ECHAM's `pxim1` / `pxlm1`: the **grid-mean**
condensate the step started from. They were being passed the post-process
in-cloud values, so `ll_liq_neg = liq_mmr_next < ccwmin` compared an in-cloud
magnitude (larger by 1/cf) against a grid-mean threshold — the guard
under-triggered, and where it did fire it injected an in-cloud-sized correction
into a grid-mean tendency and into `dqdt`/`dtedt`.

### Also: one cloud fraction for every in-cloud → grid-mean conversion

`update_in_cloud_water` is where ECHAM rewrites `paclc` — a clear cell with
positive condensation becomes cloudy, and its in-cloud condensate is defined
against that *new* fraction. Conversions downstream were still using the
pre-activation `cloud_fraction`, so in exactly those cells they converted at
cf = 0 and the whole transfer vanished from the ledger while the in-cloud state
moved. Everything after that call now uses `cloud_fraction_uicw`. Findings 1
and 3 do not close without it.

## The gate: `TestColumnEnthalpyConservation2M`

Internal phase changes leave

```
E = Σ ρ·dz·(cpd·T − Lv·qc − Ls·qi)
```

invariant — condensation trades Lv of vapour enthalpy for Lv of warming,
freezing trades Lf for Lf. Vapour cancels out of `E` entirely, which is what
makes this independent of the existing `TestColumnWaterConservation2M`. The
only way the column changes `E` is by exporting condensate, so

```
Σ ρ·dz·(cpd·dT/dt − Lv·dqc/dt − Ls·dqi/dt) == Lv·rain_sfc + Ls·snow_sfc
```

Modelled on CAM's `check_energy_chng` (column energy against
`previous + dt·boundary_flux`), not on PUMAS's `omsm` limiter — a negativity
guard cannot see a budget that is open in both directions.

Four fixtures, as the issue specifies: warm liquid, WBF-active mixed phase,
cold ice reaching the surface, and cloud ice above 0 °C. The first and third
are controls — they passed before and must keep passing. The water test is
extended with the fourth fixture too.

**Verified to fail without the fix**: with `scheme.py` reverted, the four
finding-targeted tests fail and the two controls still pass.

## Evidence

Column enthalpy residual (`dE/dt − boundary`), single-column probes:

| fixture | before | after |
|---|---|---|
| warm liquid → rain *(control)* | −3.2e−4 W/m² (−0.00 %) | −3.2e−4 W/m² |
| cold ice → surface snow *(control)* | −6.5e−5 W/m² (−0.00 %) | −3.4e−5 W/m² |
| WBF-active mixed phase | **+39.9 W/m² (+3.81 %)** | −5.3e−5 W/m² |
| cloud ice above 0 °C | **−587.8 W/m² (−425.7 %)** | −2.8e−5 W/m² |

The WBF residual was `Lf × wbf_col` to 0.2 % (3.337e5 × 1.198e−4 = 39.96),
exactly as #662 predicted.

Cloud ice above 0 °C, water budget — with **zero** surface precipitation:
residual +2.07e−4 kg/m²/s (≈18 mm/day) → +6.6e−12. `Σ dqcdt·m` was exactly
2× the ice debit (4.147e−4 vs 2.074e−4); it is now equal and opposite.

Heterogeneous freezing, `ice_nuclei` 0 → 1e6 m⁻³, WBF suppressed, level 5
(Σ Δ`dqnidt` = +2355 kg⁻¹s⁻¹ in both, so the crystals were always being made):

| | before | after |
|---|---|---|
| Δ `dqcdt` | +2.45e−11 — *wrong sign*, 60× too small | −1.486e−9 |
| Δ `dtedt` | −8.13e−9 — *wrong sign* | +4.925e−7 |
| Δ`dtedt` / −Δ`dqcdt` | — | **331.46** = `(alhs−alhc)/cpd` exactly |

## Caught by review, fixed here

A `/code-review high` pass over the diff found a defect I had introduced, which
three independent finder angles located at the same line:

**`freezing_below_238K` already area-weights its own transfer**
(`pfrl += pxlb·paclc`, `deposition_freezing.py:508`) and zeroes the in-cloud
liquid outright. Folding it into the same `* cloud_fraction` conversion as the
heterogeneous leg weighted it twice, so below 238 K only `cf` of the liquid was
debited from `qc` while the in-cloud state said the cloud was fully glaciated —
grid-mean liquid water persisting below the homogeneous freezing point, halving
each step. Only `frozen_mass` (the het leg) is a bare in-cloud increment.

**The enthalpy test structurally could not catch it**: the liquid debit, the ice
credit and the fusion heat all come from that one number, so water and enthalpy
both close on the wrong value. It needs a *state* assertion, which is what
`test_homogeneous_freezing_removes_all_the_liquid` is — below `cthomi` every drop
freezes, in a partly-cloudy box too. Verified to fail on the double-weighted form
(froze 0.0417 of 0.0714 kg/m², i.e. exactly `cf`).

That is a general limitation worth stating: a budget test pins the *ledger's*
self-consistency, not the agreement between the ledger and the in-cloud state.
Both new state-level tests (this one and the cloud-fraction clip) exist because
of it.

## Expected impact on a running model

These are physics changes, not bookkeeping:

- supercooled liquid now actually leaves `qc` where WBF fires — LWP should drop
  in mixed-phase decks;
- warm columns carrying cloud ice stop manufacturing liquid and stop cooling
  twice;
- JAM's `ice_nuclei` (#494) now removes droplet mass and releases fusion heat,
  so the aerosol → ice coupling has a thermodynamic effect for the first time;
- the post-microphysics cloud fraction reaches radiation and the aerosol
  partition (finding 4);
- newly-formed cloud cells no longer convert their transfers at cf = 0.

**No coupled integration has been run** — validation here is the unit suite plus
single-column probes, by agreement on scope. Worth watching LWP/IWP in the next
release-validation pass.

## Not in this PR — #662 stays open

**Finding 5** (`qr`/`qs` are permanently zero, so accretion by rain from above
is dead) needs `precip_formation_warm`/`_cold` moved *inside* the flux-coupled
scan, because the rain flux they need only exists in the scan carry. Warm
precip is the first step of the chain and everything downstream depends on it,
so that means moving the whole level-independent chain into the scan body and
turning ~12 arrays into scan `ys` — a restructure of the orchestrator, not a
patch. It also changes precipitation formation by ~3×, which deserves its own
evidence rather than being buried here. The Marshall–Palmer inversion to port
is `echam_1m.py:892-903` (per @duncanwp's comment on #662).

Issues filed for everything found but not fixed (repo rule):

| | |
|---|---|
| #681 | `c.alhf` (3.34e5) and `alhs − alhc` (3.33e5) disagree by 0.3 % and are used interchangeably. Surfaced because the enthalpy identity only closes against the latter — which is what the scheme's `lsdcp − lvdcp` is. |
| #685 | `pclcstar`/`pauloc` are placeholders, so accretion in partly precipitation-covered boxes is overestimated. Needs the same scan refactor as finding 5. Cross-linked from both call sites. |
| #686 | `cold_mask` gates on the step-start `qi`, so ice created *this* step (WBF, deposition, #494 het freezing) gets no aggregation or riming sink. Latent before, load-bearing now that the WBF ice credit reaches `dqidt`. |
| #687 | `clouds.cloud_fraction` now means the post-microphysics cover under `2m` and the diagnosed cover under `1m` — a 1M/2M A/B differs by a definition as well as by physics. |
| #688 | The grid-mean `ccwmin` guard clears thin stratiform cloud (cf = 0.05, in-cloud IWC 1e-6). Reference-correct, but the threshold has never been tuned against grid-mean values here. |
| #689 | A negative `qc`/`qi` undershoot is repaired by condensing vapour — consistent, but sign-definite (~+0.12 K/day per 1e-6 undershoot). Worth a diagnostic. |

## Testing

- `ruff check .` clean.
- Full CI gates (lint, fast `-m "not slow"` cov≥90, slow cov≥80) on a Derecho
  `develop`-queue node — **not** the login node, which produces ~50 phantom
  failures across unrelated subsystems (the `jcm-local-ci` skill documents this;
  I reproduced it before noticing `nproc` = 1).
- `/code-review high` (8 finder angles + verification) — findings above.
- `jcm/physics/clouds/lohmann_2m_test.py`: 62 pre-existing + 8 new, all passing.

**The slow gate is red on an unrelated flake — #692.**
`pyses_dycore_test.py::TestCoupledEchamSmoke::test_model_drives_float32_physics`
segfaults its xdist worker intermittently. Five full slow-gate runs:

| worktree | without `--cov` | with `--cov` |
|---|---|---|
| `origin/dev` @ f6d1bcd | 93 passed | 93 passed |
| this branch | 93 passed | **segfault**, **segfault** |

The branch correlation is spurious, and the mechanism settles it rather than the
statistics: that test builds `echam_physics(radiation_scheme="grey")`, which
resolves `cloud_scheme` to its `"1m"` default, so its stack holds
`echam_1m_microphysics` and **no Lohmann 2M term** — this diff is confined to
`jcm/physics/clouds/lohmann_2m/` and is never constructed, traced or executed by
it. Run alone it passes on both worktrees in near-identical time (59.80s vs
59.56s). The segfault also destroys the coverage data, which is why the run
reports ~29 % and trips `--cov-fail-under=80`.

I would not merge on a red gate without that being understood — flagging it
rather than asking anyone to take "it's flaky" on faith.

### DIFF
--- a/jcm/physics/clouds/lohmann_2m/scheme.py
+++ b/jcm/physics/clouds/lohmann_2m/scheme.py
@@ -363,10 +363,13 @@ def cloud_microphysics_2m(
 
     # Grid-mean freezing ledger (ECHAM pfrl): the single accumulator the
     # assembly step debits from liquid, credits to ice, and converts to fusion
-    # heat. Carries both legs — homogeneous below cthomi and the immersion
-    # freezing driven by ``ice_nuclei`` (#494) — since both routines above
-    # work on in-cloud condensate, hence the cloud-fraction weighting.
-    freezing_rate = (freezing_rate_hom + frozen_mass) * cloud_fraction_uicw
+    # heat. Carries both legs — homogeneous below cthomi, and the immersion
+    # freezing driven by ``ice_nuclei`` (#494).
+    #
+    # Only the heterogeneous leg is converted here. ``freezing_below_238K``
+    # already area-weights internally (``pfrl += pxlb·paclc``), whereas
+    # ``frozen_mass`` above is a bare in-cloud increment.
+    freezing_rate = freezing_rate_hom + frozen_mass * cloud_fraction_uicw
 
     # ------------------------------------------------------------------
     # WBF (Wegener-Bergeron-Findeisen): liquid → ice in mixed-phase
@@ -656,6 +659,7 @@ def cloud_microphysics_2m(
     # ------------------------------------------------------------------
     liquid_cloud_flag = temperature > params.tmelt
     ice_cloud_flag = temperature <= params.tmelt
+    cloud_fraction_in = cloud_fraction
 
     (
         cloud_fraction_final,
@@ -754,6 +758,20 @@ def cloud_microphysics_2m(
     # destruction (review finding 2.18).
     dqrdt = jnp.zeros_like(qc)
     dqsdt = jnp.zeros_like(qc)
+
+    # The microphysics may REMOVE cloud, never create it. ECHAM's write-back
+    # to ``paclc`` exists to clear cells the scheme has just emptied of both
+    # condensates; the upward branch of ``update_in_cloud_water`` — which
+    # sets cf = clip(RH, 0.01, 1) wherever a clear cell has any condensation
+    # or deposition — is a second, RH-based cloud-cover closure, and
+    # ``SundqvistCloudFraction`` is the one this stack uses. Publishing the
+    # raw value substitutes it: an ice-supersaturated stratospheric column
+    # above ``cloud_top_pressure_pa``, which Sundqvist deliberately reports
+    # as cloud-free, comes back overcast (cf = 1) on ~1e-6 kg/kg of ice, and
+    # COSP, AeroCom and the JAM cloud-borne / aqueous / wetdep terms all read
+    # it. Clipping to the incoming cover keeps the emptying behaviour and
+    # drops the closure substitution.
+    cloud_fraction_final = jnp.minimum(cloud_fraction_final, cloud_fraction_in)
 
     # update_tendencies' tracer_tendency_{cdnc,icnc} is already in per-kg-
     # of-air per second once we pass qnc/qni (per-kg) as the tm1 tracers
--- a/jcm/physics/clouds/lohmann_2m/scheme.py
+++ b/jcm/physics/clouds/lohmann_2m/scheme.py
@@ -145,7 +145,8 @@ def cloud_microphysics_2m(
     cdnc = jnp.maximum(cdnc, cdnc_min)
 
     # pauloc==1 and pclcstar==cloud_fraction are conservative first-pass
-    # approximations — refine in later 5b steps.
+    # approximations. The true pclcstar is min(cloud cover, precip cover),
+    # which only exists inside the flux-coupled scan — see #685.
     autoconv_factor = jnp.ones_like(qc)
     min_cloud_precip_fraction = cloud_fraction
 
@@ -299,13 +300,18 @@ def cloud_microphysics_2m(
         params,
     )
 
+    # ``update_in_cloud_water`` rewrites ``paclc`` (a clear cell with positive
+    # condensation becomes cloudy), and the in-cloud condensate it returns is
+    # defined against that fraction. Every in-cloud → grid-mean conversion
+    # below therefore uses ``cloud_fraction_uicw``.
+
     # ------------------------------------------------------------------
     # Freezing below 238 K (homogeneous freezing, level-independent)
     # ------------------------------------------------------------------
     freezing_condition = temperature < params.cthomi
     (
         icnc_frz, _droplet_freezing_rate, cdnc_frz,
-        _freezing_rate, in_cloud_ice_frz, in_cloud_liquid_frz,
+        freezing_rate_hom, in_cloud_ice_frz, in_cloud_liquid_frz,
     ) = freezing_below_238K(
         freezing_condition,
         cloud_fraction_uicw,
@@ -355,6 +361,13 @@ def cloud_microphysics_2m(
         het_condition, jnp.maximum(cdnc_frz - new_crystals, params.cqtmin), cdnc_frz,
     )
 
+    # Grid-mean freezing ledger (ECHAM pfrl): the single accumulator the
+    # assembly step debits from liquid, credits to ice, and converts to fusion
+    # heat. Carries both legs — homogeneous below cthomi and the immersion
+    # freezing driven by ``ice_nuclei`` (#494) — since both routines above
+    # work on in-cloud condensate, hence the cloud-fraction weighting.
+    freezing_rate = (freezing_rate_hom + frozen_mass) * cloud_fraction_uicw
+
     # ------------------------------------------------------------------
     # WBF (Wegener-Bergeron-Findeisen): liquid → ice in mixed-phase
     # ------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


duncanwp:
@codex review

Issue #706, opened 2026-08-21
Cloud schemes use dry-air cp in lvdcp/lsdcp; ECHAM uses the moist heat capacity


duncanwp:
Both ECHAM cloud ports compute the ledger latent-heat factors as constants:

- `lohmann_2m/scheme.py`: `lsdcp = c.alhs / c.cpd`, `lvdcp = c.alhc / c.cpd`
- `echam_1m.py`: same convention (`zlvdcp`/`zlsdcp` built from `cpd`)

ECHAM builds them from the **moist** heat capacity (mo_cloud_micro_2m.f90:846-848):

```
ztmp1 = 1/(cpd + zcons1*MAX(pqm1,0))     ! zcons1 = cpd*vtmpc2
zlvdcp = alv*ztmp1 ;  zlsdcp = als*ztmp1
```

i.e. `L/(cpd·(1+vtmpc2·q))` — up to ~1.3 % smaller latent heating/cooling in the moist tropics (vtmpc2 ≈ 0.86, q up to ~0.018). Systematic, not noise: every condensation event in the tropical boundary layer over-heats by that fraction.

Deliberately NOT fixed in the #662/#667 consolidation because the column enthalpy gate (`TestColumnEnthalpyConservation2M`) integrates `cpd·dT` — switching the schemes to moist cp requires switching the gate (and any future `check_energy_chng`-style budget) to the same moist cp in the same change, or the budget opens by exactly the fraction being fixed. Do both together, for 1M and 2M at once, and re-verify the four enthalpy fixtures.

Found during the #662/#667 restructure; recorded in `docs/source/design/lohmann_2m_column_processes.md` (deliberate omissions).

Issue #705, opened 2026-08-21
2M microphysics: large-scale vertical velocity is not plumbed (zvervx, zauloc knvb gate, het_mxphase pvervel)


duncanwp:
Found while restructuring the 2M scheme into ECHAM's `column_processes` loop (#662/#667/#685/#686 PR). Three consumers of the large-scale vertical velocity are running without it:

1. **`zvervx` (updraft for the WBF/lo2 gates)** — ECHAM composes it from `pvervel` (large-scale omega converted to w) plus the TKE term (and an orographic gust term via `get_util_var`); jcm computes only `fact_tke·sqrt(2·TKE)·100` (`scheme.py`, updraft_velocity). The Korolev/Mazin threshold comparison `0.01·zvervx < zvervmax` therefore misses the resolved-ascent contribution — mixed-phase clouds in strong large-scale ascent should escape WBF glaciation and currently cannot.
2. **The `zauloc` inversion exception (#685 remainder)** — ECHAM zeroes the local accretion fraction at the two levels around a diagnosed low-level inversion when `pvervel > 0` (subsidence) and `lonacc` (mo_cloud_micro_2m.f90:1621-1627, `knvb`/`jbmin`/`jbmax`). Skipped in the port; noted on `CloudParams2M.cauloc`. The inversion level `knvb` is diagnosed by the Sundqvist cover scheme (jcm has the inversion finder in `sundqvist._stratocumulus_zsat` but does not export the level index).
3. **`het_mxphase_freezing`** (the faithful ECHAM contact/immersion machinery, currently not called from the orchestrator — see #552 notes) takes `pvervel` directly.

Hooks: the dycore knows omega; a diagnostics key (e.g. `vertical_velocity` on the moist-air state term) would let the 2M term read it the same way it reads TKE. Reference: `mo_cloud_micro_2m.f90` section 1 (`zvervx` assembly) and 1614-1627.

Issue #686, opened 2026-08-19
2M: cold_mask gates precip_formation_cold on the step-start qi, so ice made this step gets no aggregation/riming sink


duncanwp:
`cloud_microphysics_2m` builds the cold-precip mask from the **input** grid-mean ice:

```python
cold_mask = (temperature <= params.tmelt) & (qi > params.ccwmin)   # scheme.py:419
```

ECHAM's equivalent (`ll_prcp_cold`) is `ld_cc .AND. zxib > cqtmin` — the *current in-cloud* ice, i.e. after deposition, homogeneous/heterogeneous freezing and WBF have run. Because `qi` here is the value the sweep started with, any ice **created during the step** is invisible to `precip_formation_cold`: no aggregation to snow (`zsaut`), no riming (`zsaclin`), no ice-snow accretion (`zsaci`).

## Why it matters more after #662

A pure supercooled liquid deck — `qc = 3e-4`, `qi = 0`, `cf = 0.9`, T = 258 K — glaciates via WBF to `in_cloud_ice_wbf ≈ 3.3e-4`, but `cold_mask = (258 <= 273) & (0.0 > 1e-7) = False`, so `snow_formation_gridmean = 0` and `psacl = 0` for that step.

Before the tendency-ledger fix the WBF ice credit was discarded, so the phantom ice never reached `qi` and the missing sink was unobservable. Now that `ice_tendency = ice_tendency_scan + ice_tend_wbf` routes it into `dqidt`, the ice is real prognostic mass whose only escape that step is sedimentation. The same holds for any cell where `deposition_rate > 0` lifts `qi` off zero, and for the heterogeneous-freezing path of #494.

## Fix

Gate on the post-freezing in-cloud ice, i.e. build the mask from `in_cloud_ice_wbf` (the value actually passed to `precip_formation_cold`) rather than the entry `qi`:

```python
cold_mask = (temperature <= params.tmelt) & (in_cloud_ice_wbf > params.cqtmin)
```

Note the threshold differs too (`ccwmin` = 1e-7 grid-mean vs ECHAM's `cqtmin` = 1e-12 in-cloud), so this is two changes; worth checking both against `mo_cloud_micro_2m.f90` before landing, and worth a fixture that seeds `qi = 0` into an active WBF/deposition cell and asserts snow forms.

Found by review during #662 (not fixed there — it changes which cells precipitate, so it wants its own evidence).

Issue #685, opened 2026-08-19
2M microphysics: pclcstar and pauloc are hardwired to 1 / cloud fraction, overestimating accretion in partly precipitation-covered boxes


duncanwp:
`cloud_microphysics_2m` (`jcm/physics/clouds/lohmann_2m/scheme.py:147-150`) sets two ECHAM accretion geometry factors to placeholders:

```python
# pauloc==1 and pclcstar==cloud_fraction are conservative first-pass
# approximations — refine in later 5b steps.
autoconv_factor = jnp.ones_like(qc)
min_cloud_precip_fraction = cloud_fraction
```

Both are consumed as-is:

- `pclcstar` (`min_cloud_precip_fraction`) is ECHAM's `min(cloud cover, precipitation cover)` — documented as such in `precip.py:274-276`. It weights accretion of cloud water by rain falling from above (`zrac1`) in `precip_formation_warm`, and the riming/aggregation weighting in `precip_formation_cold`.
- `pauloc` (`autoconv_factor`) is the fraction of the box participating in accretion with *newly formed* rain (`zrac2`), pinned to 1.

Setting `pclcstar = cloud_fraction` assumes the precipitation cover is always at least the cloud cover. In a box where rain from above covers less than the cloud does — a decaying cell, or precipitation shafts narrower than the deck — the accretion sink is weighted by the larger of the two areas and is systematically overestimated.

This is a pre-existing gap, but the current PR extends it to the cold chain: `precip_formation_cold` is now called with `cloud_fraction_uicw` for both the cloud fraction and `pclcstar`, making the placeholder explicit at a second call site (`scheme.py:440`).

Filed because the in-code note ("refine in later 5b steps") carries no issue number and nothing tracks it — `grep -rn 'pclcstar|pauloc' jcm/` returns only those two comments plus the parameter docs.

## What a fix needs

The real `pclcstar` needs the precipitation-covered fraction, which the scheme already computes: `precip_cover` (ECHAM `pclcpre`) is carried through the `_flux_coupled_step` scan and updated by `gridbox_frac_falling_hydrometeor` in `update_precip_fluxes`. It only exists *inside* the scan, whereas the precip-formation calls are currently outside it — the same structural obstacle as #662 finding 5 (`qr`/`qs` diagnosed from the fluxes). Both are unblocked by the same refactor: moving the level-independent chain into the flux-coupled scan. Worth doing together.

[...]

PR #707, opened 2026-08-21
2M clouds: the faithful column_processes consolidation (#662-f5, #667, #685, #686, #687, #688, #689, #681)

### LINKED ISSUE #662: Lohmann 2M: process routines mutate in-cloud state without feeding the tendency ledger
Found by a re-audit of the July 2026 ECHAM+RRTMGP review against `dev` @ `f6d1bcd`. Parent: #614.

Six findings, one root cause and one fix pass over `lohmann_2m/scheme.py` and `assembly.py`: **process routines compute a transfer, mutate their local in-cloud state, and the tendency ledger never receives it** — or the reverse, a value is computed and discarded. The conservation test in the last section is the gate for all of them.

## 1. WBF applies its latent heat but discards the mass transfer

`scheme.py:380-395` drops `_liq_tend_wbf` and `_ice_tend_wbf` while keeping `dtedt_wbf` (seeded at `:698`). `WBF_process` (`deposition_freezing.py:830-838`) returns all three; ECHAM passes `pxlte, pxite, ptte` as INOUT.

Probe, T=258 K, qc=3e-4, cf=0.9, TKE=0: `dqcdt = 0.000e+00` **exactly** while `wbf_col = 3.088e-04 kg/m²/s`. Column enthalpy residual **+102.7 W/m² = 33 % of gross**, exactly `Lf × wbf_col`. Introduced by the fix that added the Bergeron threshold-velocity gate.

## 2. Cloud ice above 0 °C melts twice

The melting routine's ice debit is discarded (`scheme.py:526`) and `ice_tendency` is seeded only from the sedimentation delta (`:705`). Because sedimentation runs *before* melting and the **pre**-sedimentation `qi_k` is passed as `ice_cloud_previous`, `pimlt` and `pximlt` both claim the same mass.

Probe, column whose only condensate is cloud ice at T>tmelt: `dqcdt = 2 × qi/dt`, water residual **+3.079e-04 kg m⁻² s⁻¹ (≈26 mm/day) with zero precipitation**, enthalpy residual −102 W/m².

Note the melt-after-sediment *ordering* is fine — it is MG/PUMAS's (`micro_pumas_v1.F90:3093` sediment → `:3293` melt). PUMAS applies the ice debit correctly (`qitend = ((1-dum)*dumi - qi)*rdeltat`, `:3323`); jcm discards it. The bug is the discard, not the order.

## 3. Heterogeneous freezing moves crystal number but not mass or fusion heat

`scheme.py:344-350` computes `frozen_mass`, applies it to `in_cloud_ice_het`/`in_cloud_liquid_het`, but never adds it to the `freezing_rate` accumulator. `assembly.py:176-183` builds the tendencies purely from the rate terms, so the transfer is invisible to the state.

Probe, raising `ice_nuclei` 0 → 1e6 m⁻³: new crystals 8.5e5–9.9e5 m⁻³ (matches ΔICNC exactly), but Δ`dqcdt` = **+1.6e-12 against an expected −3.30e-9** — three orders of magnitude small and the wrong sign — and Δ`dtedt` = −5.3e-10 against +1.09e-6 K/s.

This is a *missing pathway*, not a budget violation: crystals are created with zero mass and droplets destroyed with zero mass, so water still conserves. But `assembly.py:246-263` passes the **post**-freezing in-cloud values to the effective-radius block, so exported `liq_eff_radius`/`ice_eff_radius` are computed on liquid the tracers never lost and ice they never gained. This is the whole point of #494 — the aerosol→ice coupling currently moves crystal number and nothing else.

## 4. `cloud_fraction_final` is discarded

`scheme.py:639-645` unpacks it; it appears exactly once in the file. `assembly.py:215-219` computes it correctly (zeroes cf where both condensates fall below `ccwmin`, and where `cf < clc_min`), mirroring ECHAM's write-back to `paclc` — the array radiation reads. Cells the microphysics has just emptied keep a non-zero cloud fraction.

## 5. `qr`/`qs` inputs are permanently zero — accretion by rain from above is dead

`scheme.py:917-918` reads `state.tracers.get("qr"/"qs", zeros)`, but no term declares them any more (`grep TracerSpec("qr")` → nothing), so both are identically zero in every run. They feed `precip_formation_warm` (`:185`) and `_cold` (`:424`): `zrac1 = qc*(1 - exp(-3.7*dt*qr))` is now **exactly 0**, and riming/aggregation lose their snow-from-above term.

Probe: rain flux 1.36e-5 (qr=0) vs 4.19e-5 (qr=1e-4) — a 3× difference the model can never realise. This is a regression from removing the prognostic tracers: the reads weren't removed with them. ECHAM does not use tracers here — it diagnoses `zxrp1`/`zxsp1` from the *fluxes* by Marshall–Palmer inversion inside the level loop (`mo_cloud_micro_2m.f90:1638-1654`), which needs these calls moved inside `_flux_coupled_step`.

## 6. The negative-mass guard is fed in-cloud values where grid-mean is expected

`scheme.py:648-649` passes `ice_mmr_prev=in_cloud_ice_cold`, `liq_mmr_prev=in_cloud_liquid_cold`; ECHAM passes grid-mean `pxim1`/`pxlm1` (`mo_cloud_micro_2m.f90:1779-1780`). `dqcdt`/`dqidt` are unaffected, but `ll_liq_neg = liq_mmr_next < ccwmin` (`assembly.py:192, 204`) tests an in-cloud magnitude against a grid-mean threshold, so the guard under-triggers — and when it fires, `zdxlcor = -liq_mmr_next/dt` injects an in-cloud-magnitude correction into a grid-mean tendency and into `dqdt`/`dtedt`.

## The gate: a column enthalpy-conservation test

None exists — `grep` for `alhf|enthalp|moist static` in `lohmann_2m_test.py` returns nothing. Water is covered (`TestColumnWaterConservation2M`), but its fixtures never reach the two states that break: cloud ice above 0 °C, and an active WBF cell.

Model it on CAM's `check_energy_chng_run` (`atmos_phys/schemes/conservation_adjust/check_energy/check_energy_chng.F90:309-340`) — column-integrated total energy and total water against `previous + dt·boundary_flux` — **not** on PUMAS's `omsm` sink limiter (`micro_pumas_v1.F90:2218-2520`), which is a negativity guard and would not have caught any of this. CAM's thresholds are 1e-14 (energy) / 1e-10 (water); findings 1 and 2 here are ~1e-1 relative errors.

Assert `Σ_k Δp/g·[cp·ΔT/Δt] == -Lv·cond - Lf·(freeze/melt) + surface precip enthalpy` over four fixtures: warm liquid, WBF-active mixed phase, cold ice reaching the surface, and **cloud ice above 0 °C**. Extend the water test with the last fixture too.

## Provenance

Checked against ECHAM 6.3-HAM2.3 and ESCOMP/CAM `cam6_4_196`. Both references put the mass term in for every one of these — ECHAM `pfrl` at `mo_cloud_micro_2m.f90:3600-3615`, CAM `mnuccc = nnuccc*mi0l` → `qctend`/`qitend`/`tlat*xlf` at `micro_mg2_0.F90:1478, 1899-1911`. A full literal sweep found `lohmann_2m/` to be essentially pure ECHAM-HAM, so there is no alternative-reference defence for any of these.

## Refs

Review findings 0.2 (resolved), 2.17 (resolved), 2.18, 2.19, 2.23, 2.24-b/d/g. Related: #494, #539, #537, #552.

duncanwp:
The July-2026 cloud audit filed its findings separately for provenance, but many touch the same code and **counteract each other** in full simulations (an inert internal saturation adjustment masked by an external bolt-on; dead rain-from-above accretion partly masked by an over-generous accretion geometry), so fixing them piecemeal would have shifted the climate at every step. This PR lands the remaining cluster in one branch, with each issue an atomic commit.

Closes #662 (finding 5 — the last one open), #667, #685, #686, #687, #689.
Refs #688 (threshold verified reference-correct, kept; evidence below), #681 (landed here as the first commit), #658 (unblocks the cloud-borne scavenging numerator), #552, #705, #706.

## The central change: the 2M chain is ECHAM's `column_processes` loop

The reference runs the **entire** process chain per level, top-down, because every process at level jk consumes the precipitation state (`prfl`/`pssfl`/`zclcpre`/`zxiflux`) the levels above produced *this step*. The port had factored the "level-independent" processes out of the flux-coupled scan — which silently severed exactly those couplings. Moving the whole chain into the `lax.scan` (design doc: `docs/source/design/lohmann_2m_column_processes.md`) closes four issues at once:

- **#662 finding 5** — `zxrp1`/`zxsp1` are now the Marshall–Palmer inversions of the carry fluxes (Roeckner et al. 2003 eqs. 10.70/10.74), replacing reads of `qr`/`qs` tracers that no term declares (both were identically zero: accretion by rain from above and snow-from-above riming were dead code). Probe: rain formation now compounds down a warm deck (2.3e-7 → 3.1e-7 kg/kg per level) instead of staying flat.
- **#685** — `zclcstar = min(paclc, zclcpre)` from the in-scan precipitation cover, and `zauloc = clip(cauloc·dz/5000, clmin, clmax)`. The `knvb`/`lonacc` exception needs `pvervel` (not plumbed) → #705.
- **#686** — cold precip is gated by ECHAM's `ll_cc` only; the internal `zxib > cqtmin` check runs on the *current* in-cloud ice, so ice deposited/frozen/WBF-transferred this step meets aggregation/riming in the same step. Gated by a new test: a qi=0 deck glaciating via WBF exports a frozen flux the same step.
- **#667.4** — warm rain runs *after* condensation and activation (both references); the docstring no longer claims "ECHAM6 order" for the old layout. The sediment→melt order deliberately stays MG/PUMAS's (reviewed in #662).

## #667: the scheme owns saturation adjustment

- **zdqsdt fixed** (commit 2): ECHAM's derivative is a finite difference over one lookup step (0.001 K); the port stepped 1.0 K but kept the ×1000 factor, so `zqcon = 1/(1+L/cp·zdqsdt)` sat at ~1/50…1/650 and the internal adjustment was inert. Both points now come from the shared `jcm.physics.thermodynamics` Tetens pairs with the analytic derivative. Measured: zqcon at 285 K goes 1/644 → 0.357.
- **Section-5 condensation ported**: the `zqcdif` moisture-convergence closure (zdqsat with the melt/evap heat corrections, the `zifrac` cnd/dep split, the dissipation branch) seeds `mixed_phase_deposition_and_corrections`.
- **The Sundqvist bolt-on is removed** from the Term — with zdqsdt fixed, keeping it would remove supersaturation twice per step with double the latent heating, exactly as the issue predicted. Gate: `TestSaturationGate2M` (10 % supersat → ECHAM's 1 % zoversat tolerance in one step).
- **`zxlevap`/`zxievap` ported verbatim** — condensate in cloud-free cells and the clear-sky share of positive increments evaporate with matching latent cooling. Probe: a cf=0 cell holding 5e-4 qc returns it to vapour in one step (previously untouchable, `max|dqcdt| = 0`).
- **One saturation convention**: the binary `lo2` switch governs the scheme; `sundqvist._qs_and_dqs_dt` takes `t_mix_min` from the config so the es-ramp and L-ramp cannot desynchronise under a Hydra override; the dead `t_ice` field now drives `_qs_cover`.

### Latent defects the restructure surfaced (all fixed verbatim from the Fortran)

- `peta`: `threshold_vert_vel` received a dimensionless 0–1 saturation-ratio clip where ECHAM's diffusional-growth factor ζ (mo_cloud_micro_2m.f90:856) belongs — the WBF weak-updraft gate and the lo2 phase decision were miscalibrated by orders of magnitude. ζ is now computed at the t−1 anchors, and the WBF threshold is recomputed from the post-freezing ice (ECHAM 1580-94).
- `prid`: `update_in_cloud_water`'s mean ice radius must be the Schumann (2011) **volume-mean radius in metres**; it received an effective radius in **microns** (r³ off by 1e18), pinning the `nic_cirrus=1` diagnostic ICNC at the `icemin` floor.
- `zqrho = 1.3/ρ` (not 1/ρ) in the sublimation ventilation factor and the cold-chain aggregation.
- `ll_liqcl`/`ll_icecl` are condensate+number flags (ECHAM 1755-60), not a temperature split — fixes which cells export effective radii.

### State-splitting convention

Primary inputs are the post-upstream provisional state (what the returned tendencies are relative to); optional `*_m1` inputs are the step-start state = ECHAM's (ptm1, pqm1, pxlm1, pxim1) anchors, with the differences playing `ztmst·pqte` in the closure. The ledger reconstruction is identical either way, so the host's additive tendency sum telescopes exactly as ECHAM's INOUT accumulation. Documented on the function and in the design doc.

## The rest

- **#687** — the 1M term now writes ECHAM's `paclc` write-back too (mo_cloud.f90:1280), so `clouds.cloud_fraction` means the post-microphysics cover under both schemes; semantics documented on `CloudData`. Note: `tools/release_validation/health.py`'s 0.4–0.8 cover band was calibrated on the pre-microphysics cover — should be re-checked, though the write-back only clears near-empty cells.
- **#689** — the negative-mass repair (`zdxlcor`/`zdxicor`) is exported as `clouds.negative_mass_repair` [W/m², column-integrated latent heating], so its sign-definite warming is measurable in any run. The repair itself is unchanged.
- **#688** — verified against the reference: ECHAM tests grid-mean `zxlp1`/`zxip1` against the same `ccwmin = 1e-7`, so the post-#662 pairing is reference-correct and kept. Whether thin-cirrus/#494 bootstrap needs a cf-aware floor is now measurable via the new diagnostic + IWP in real runs.
- **#681** — `alhf` is a derived `@property` (`alhs − alhc`), exactly ECHAM's `alf = als − alv`; the 0.3 %-inconsistent independent base value is gone. Tiedtke's melting heat shifts 0.3 % (within its test tolerance).

## Out of scope, deliberately

- **#552** (Kärcher–Lohmann / Liu–Penner cirrus nucleation) is a full new-scheme port, not a counteracting fix — left open; the `nic_cirrus=2` section-5 branch returns zero deposition as with a missing external source in ECHAM.
- **#692** (pySES xdist segfault) is test infra.
- **#705** (vertical-velocity plumbing) and **#706** (moist cp — must move together with the enthalpy gate) filed from findings here.

## Verification

- `ruff` clean; fast suite (1884) at 95 % coverage; slow suite at PR coverage — all green locally.
- Conservation gates: `TestColumnWaterConservation2M` + `TestColumnEnthalpyConservation2M` (4 fixtures each) pass through the restructure. Three test expectations were architecture-bound and updated to reference-invariant form (below-cloud sedimented ice arrives as vapour via `zxievap`; melting is "exactly once", not "purely to qc"; het-INP warming bounded below by the fusion heat with the enthalpy gate excluding fabrication) — each documented in the test.
- New gates: `TestSaturationGate2M`, `TestColdChainSameStepCoupling2M`, `TestCloudFractionWriteBack1M`.
- Gradients: AD-vs-FD on d(surface precip)/d(ccraut) through the full scan agree to 4 significant figures; d/dT finite everywhere.
- Representative runs: 60-day T63L47 1M-vs-2M A/B (identical dry-JW init, real orography/SSTs, RRTMGP), results in the thread below.

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -15,6 +15,7 @@ JAX-GCM is designed to be a fully differentiable climate model that balances eas
    design/dinosaur_sl_jam_configuration
    design/speedy_variable_levels
    design/frontal_gravity_wave_drag
+   design/lohmann_2m_column_processes
    design/aerosol_optics_diagnostics
    design/aerocom_erfari_sampling
    design/data_mirror
--- a/docs/source/design/lohmann_2m_column_processes.md
+++ b/docs/source/design/lohmann_2m_column_processes.md
@@ -1,0 +1,118 @@
+# The Lohmann 2M scheme is ECHAM's `column_processes` loop
+
+*(2026-08; issues #662, #667, #685, #686, #687, #688, #689; parent audit #614)*
+
+## The decision
+
+`cloud_microphysics_2m` runs the **entire** two-moment process chain inside one
+top-down `lax.scan`, one scan step per level, in the section order of ECHAM's
+`mo_cloud_micro_2m.f90` `column_processes` loop. Section 8 (the tendency
+ledger, `update_tendencies_and_important_vars`) has no cross-level coupling and
+runs vectorized on the stacked per-level outputs afterwards.
+
+The previous layout split the chain: "level-independent" processes
+(condensation partition, freezing, WBF, warm/cold precipitation formation) ran
+vectorized outside the sweep, and only sedimentation/melting/sublimation ran
+inside it. That looked like a harmless performance-friendly factorization. It
+was not, because in the reference **every process at level `jk` consumes the
+precipitation state the levels above produced *this step*** — the rain/snow
+fluxes (`prfl`/`pssfl`), the precipitation cover (`zclcpre`), and the falling
+ice flux (`zxiflux`) are loop-carried. Splitting the chain silently severed
+those couplings:
+
+- accretion by rain/snow from above read `qr`/`qs` tracers that no term
+  declared any more — both identically zero, so the pathway was dead code and
+  precipitation formation was understated ~3× (#662 finding 5, and through
+  `rate_cb = p_form/(qc+qi)` the leading suspect for the cloud-borne aerosol
+  drainage failure in #658);
+- the cloud∩precipitation overlap `zclcstar = min(paclc, zclcpre)` could not
+  be formed, so accretion geometry fell back to `paclc` (#685);
+- ice created mid-step (deposition, freezing, WBF) never met its
+  aggregation/riming sink that step (#686);
+- warm rain ran *first*, before condensation and activation, matching neither
+  ECHAM nor CAM (#667 finding 4).
+
+The restructure is therefore not an optimization choice reversed — it is the
+minimal faithful shape. Wall-clock cost is unchanged at leading order: the
+same arithmetic moves from a `(nlev,)`-vectorized context into the scan body,
+and the scan already existed.
+
+## What the sweep does per level
+
+In ECHAM section numbering: 4 sedimentation → 3.1 melting → 3.2/3.3
+sublimation/rain-evaporation → in-cloud prep with clear-sky evaporation
+(`zxlevap`/`zxievap`) → 5 the `zqcdif` condensation closure → 5.4
+supersaturation corrections → 5.5 in-cloud update + activation/nucleation →
+6.1 homogeneous freezing → 6.2 heterogeneous freezing + WBF → 7 precipitation
+geometry (`zclcstar`, `zauloc`, the Marshall–Palmer inversions
+`zxrp1`/`zxsp1`) → 7.1 warm rain → 7.2 cold precipitation → 7.3 flux update.
+
+One deliberate deviation, reviewed and kept: sedimentation runs **before**
+melting (MG/PUMAS order, `micro_pumas_v1.F90`), with the running ice tendency
+threaded through the melt routine so the two sinks cannot claim the same mass
+(#662 finding 2). ECHAM melts first; both orderings are internally consistent
+ledgers.
+
+## The state-splitting convention
+
+ECHAM is leapfrog: the scheme receives the t−1 state (`ptm1`, `pqm1`,
+`pxlm1`, `pxim1`) plus accumulated tendencies (`ptte`, `pqte`, …) and *adds*
+its own contributions to the tendencies. jcm is additive operator-split: each
+term returns its own tendency against a provisional post-upstream state.
+
+The mapping used (and documented on `cloud_microphysics_2m`):
+
+- primary inputs = the **post-upstream provisional** state (`thermo_run` T/q,
+  `clouds.qc/qi` with convective detrainment folded in) — what the returned
+  tendencies are relative to;
+- optional `*_m1` inputs = the **step-start** state — ECHAM's t−1 anchors.
+  Saturation and all section-1 fields evaluate there, and the differences
+  `(x − x_m1)` play the role of `ztmst·pqte` in the condensation closure and
+  of `ztmst·pxlte` in the clear-sky-evaporation split.
+
+The ledger reconstruction is identical either way:
+`pxlm1 + Δt·(upstream+own) ≡ qc_provisional + Δt·own`, so the negative-mass
+guard bounds the true end-of-step state and the host's tendency sum
+telescopes exactly as ECHAM's INOUT accumulation.
+
+## What this made live (previously inert or miscalibrated)
+
+| pathway | before | after |
+|---|---|---|
+| internal saturation adjustment | `zqcon ≈ 1/50…1/650` (zdqsdt ×1000; #667.1) | Newton damper 0.36–0.93; 10 % supersat → 1 % in one step |
+| grid-scale condensation | external Sundqvist bolt-on (double-adjustment risk) | ECHAM section-5 `zqcdif` closure in-scheme; bolt-on removed |
+| clear-sky condensate sink | none (`pxlevap = pxievap = 0`) | `zxlevap`/`zxievap` verbatim; cf=0 cells evaporate in one step |
+| rain/snow-from-above accretion | dead (`qr`/`qs` ≡ 0) | Marshall–Palmer inversion of carry fluxes |
+| WBF threshold updraft | fed a 0–1 clip as `peta` | ECHAM's diffusional-growth ζ (line 856); recomputed post-freezing |
+| diagnostic ICNC (`nic_cirrus=1`) | `prid` in µm (r³ off by 10¹⁸) → pinned at floor | Schumann volume-mean radius in metres |
+| cold-chain gate | step-start `qi > ccwmin` | `ll_cc` + in-scan `zxib > cqtmin` (#686) |
+
+## Deliberate omissions (tracked)
+
+- **Large-scale vertical velocity is not plumbed** (`zvervx` is TKE-only; the
+  `knvb`/`lonacc` inversion gate on `zauloc` is omitted; `het_mxphase_freezing`
+  likewise lacks `pvervel`). Tracked in #705.
+- **`nic_cirrus = 2`** still expects the Kärcher–Lohmann `pnicex`/`zqinucl`
+  source jcm does not compute (#552); its section-5 deposition branch returns
+  zero, as in the reference with a missing external source.
+- **Moist heat capacity**: ECHAM's `zlvdcp = Lv/(cpd + cpd·vtmpc2·q)`; jcm
+  uses `Lv/cpd`. ~1 % on latent heating; kept so the column enthalpy gate has
+  a single cp. Tracked in #706.
+
+## The gates
+
+`TestColumnWaterConservation2M` and `TestColumnEnthalpyConservation2M`
+(modelled on CAM's `check_energy_chng`) close the water and enthalpy budgets
+against the surface fluxes for warm-liquid, WBF, cold-fallout, and
+melt-in-place fixtures. `TestSaturationGate2M` pins that no supersaturation
+survives a step (the property whose absence let the double adjustment hide);
+`TestColdChainSameStepCoupling2M` pins that a deck glaciating via WBF exports
+a frozen flux the same step. Budget tests pin the ledger's self-consistency —
+a defect that mis-states the in-cloud state on both sides of a transfer needs
+a *state* assertion, which is why the het-INP test bounds the per-level
+fusion heat from below rather than trusting closure alone.
+
+`clouds.cloud_fraction` now means the post-microphysics cover under both the
+1M and 2M schemes (#687; documented on `CloudData`), and the 2M
+negative-mass repair is exported as `clouds.negative_mass_repair` [W/m²]
+(#689) so its sign-definite heating is measurable in any run.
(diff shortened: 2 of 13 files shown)


duncanwp:
**On the Codex P1 (cover write-back vs same-step JAM scavenging)** — verified against the code and the reference; refuted as a defect of this PR, and the substance is now tracked in #708:

- **Not reachable as a regression here.** JAM composes only with `cloud_scheme: 2m` (`echam-jam.yaml`), and the 2M cover semantics are byte-identical to `dev` (#693 published the write-back; this PR only adds the equivalent to 1M, which no JAM config uses).
- **The write-back itself is the reference behaviour for scavenging too.** ECHAM's own wet chemistry (`cloud_subm_2`, mo_cloud_micro_2m.f90 §9) runs after `column_processes` and receives the POST-write-back `paclc` — "preserve the process-time cloud fraction for these consumers" would deviate from ECHAM.
- **The real gap is on the JAM side and pre-existing**: the implicit interstitial in-cloud pathway keys to `cf_clip` where ECHAM keys scavenging to the process-time ledger (`zmlwc`/`zmratepr`), which jcm already exports as `clouds.precip_formation_rate`. The explicit cloud-borne pathway is already ledger-keyed (`p_form/condensate`) and is unaffected. Two precisions on the comment's mechanics: `CloudBorneExchange` at cf=0 is a τ=900 s relaxation toward zero, not instantaneous resuspension, and it races the same step's ledger-keyed rainout rather than simply winning.

Filed as #708 with the reference-faithful fix direction (key the implicit pathway to the formation ledger); it needs its own JAM burden A/B because it interacts with #658 in the opposite direction.

duncanwp:
## Representative-run evidence: 60-day 1M-vs-2M A/B

T63L47, real orography/SSTs, RRTMGP, dt=12 min, identical dry-JW inits; `physics=echam` vs `physics=echam-rrtmgp-2m` on two A100s. Days 30–60 means (chunk files on the dev box at `/scr/dwatsonparris/2m_faithful_ab/`, logs `run_logs/ab{1m,2m}_260821_213758.log`):

| observable | 1M | 2M | 2M/1M |
|---|---|---|---|
| LWP [g/m²] | 254 | **94** | 0.37 |
| IWP [g/m²] | 15.5 | 5.4 | 0.35 |
| total cloud cover | 0.75 | 0.79 | 1.06 |
| LS rain [mm/day] | 0.92 | 1.08 | 1.18 |
| LS snow [mm/day] | 0.074 | 0.019 | 0.26 |
| column p_form [kg/m²/s] | 1.67e-5 | 2.00e-5 | **1.20** |
| CRE_SW [W/m²] | −79 | −55 | 0.69 |
| CRE_LW [W/m²] | 34 | 16 | 0.48 |
| neg-mass repair [W/m²] mean / p99 | 0 / 0 | 0.068 / 2.2 | — |

Both runs: **0 NaN vars in every health check through day 60.**

What this settles:

- **The #667 step-4 re-test.** The bolt-on's justifying A/B ("2M LWP ~50× smaller than 1M") does not reproduce with the core owning condensation: 2M LWP is 0.37× the 1M value — a plausible inter-scheme difference, with 2M the one closer to observed global LWP (~80–100 g/m²) and CRE_SW (obs ≈ −45…−50 W/m²; the 1M's over-cloudiness is the known bias from the climatology sweeps). The bolt-on removal stands on evidence, not just on principle.
- **The #658 unblocking claim.** Column `p_form` — the cloud-borne scavenging numerator — is now 1.2× the 1M value under 2M (previously suppressed by the dead `zrac1`); the drainage pathway has a working rate.
- **The #689 decision question.** The negative-mass repair measures 0.068 W/m² global-mean (p99 = 2.2 W/m²) against ~55 W/m² of CRE — the sign-definite bias is real but ~0.1 % of the cloud radiative budget at T63; no cap or redesign warranted, and it is now permanently visible in output.
- **#688**: 2M IWP (5.4 g/m²) and cirrus survive the grid-mean ccwmin guard at climatological magnitudes; no evidence the threshold needs a cf-aware floor at this resolution.

Issue #712, opened 2026-08-22
1M scheme: publish the ECHAM scavenging ledger (mo_cloud zmlwc/zmratepr/zmrateps/zmsnowacl) so JAM wet removal can drop its cover-keyed fallback


duncanwp:
#708 keyed JAM's stratiform in-cloud wet removal and cloud-borne resuspension to the cloud scheme's process-time scavenging ledger (the ECHAM-HAM `cloud_subm_2` interface: in-cloud pools captured before precipitation formation, in-cloud formation rates, process-time cover, and the condensate-evaporation ledger — `CloudData.incloud_*`, `process_cloud_fraction`, `condensate_evaporation_rate`). Only the 2M scheme publishes it; under `cloud_scheme: 1m` the factory wires `scavenging_ledger=False` and both JAM terms fall back to the legacy cover × `p_form/(qc+qi)` reconstruction — which is unbounded in near-empty cells, reads the post-write-back cover (zero in cells the step fully converted to precipitation, so their removal is silently skipped), and gives the cloud-borne drain the resuspension-vs-rainout race #708 fixed for 2M.

## What is missing

The 1M port (`jcm/physics/clouds/echam_1m.py`) computes every ingredient in its per-level cascade — the in-cloud `zxlb`/`zxib` (qc/cf, qi/cf), the in-cloud autoconversion/accretion (`zraut`, `zrac`), aggregation (`zsaut`), riming and ice accretion, and the `zxlevap`/`zxievap` clear-sky evaporation — but does not thread them out. The reference is `mo_cloud.f90` (same `zmlwc`/`zmiwc`/`zmratepr`/`zmrateps`/`zmsnowacl` arrays passed to `cloud_subm`, with the same post-write-back zeroing at mo_cloud.f90:~1280), so the port is the 1M mirror of what #708's first commit did for 2M:

- capture the pools before the precip formation stage and pass them through the assembly zeroing,
- accumulate the in-cloud formation rates (note ECHAM 1M also seeds the snow ledger from ice sedimentation, as 2M does at mo_cloud_micro_2m.f90:1243),
- publish the seven `CloudData` fields the 2M scheme now writes.

## Hooks already in place

- `CloudData` carries the fields (zeros under 1M today) and documents their semantics, including the fraction=1 marker for a zeroed pool with positive formation.
[...]

Issue #713, opened 2026-08-23
Per-species aerosol budget-closure ledger: make slow burden drifts attributable in days, not months


duncanwp:
The SO4 runaway in the 2-year JAM run took ~9 sim-months to become visible in burdens, because a burden drift is a small residual of large opposing process fluxes — invisible in float32 5-day-mean output until it compounds. Requested by Duncan while validating #660+#708: a way to diagnose these from tendencies quickly, given that the very small values accumulate.

## Proposal

1. **Close the per-species budget in the output.** The AeroCom `wet_*`/`dry_*` ledgers already record two sinks per step; add the missing pathways so `d(burden)/dt − (emis − wet − dry − …)` is computable exactly from any pair of saves:
   - per-species emission fluxes (`emis_*`) in the same column-integrated convention,
   - the turbulent dry-deposition contribution when it exists (today `dry_*` is sedimentation alone, documented in `flux_diagnostic.py`),
   - gas-phase/aqueous chemical production of so4 (the `g_*` → `m_*` conversion flux).
   The residual is then dycore advection + numerics only. **The drift equals the ledger imbalance and is visible in a single 5-day window** at full precision, instead of after months of compounding.

2. **A per-chunk f64 imbalance line in the runner log.** Accumulate column burdens and the ledger integrals in float64 across each chunk (the accumulators live outside the f32 model state) and log one line per species: `budget so4: dB=+0.13 mg/m2/chunk, emis=…, wet=…, dry=…, chem=…, resid=…`. Greppable by the existing run watchers; no netCDF post-processing needed to see a drift start.

3. **A sink-vs-size gate (the runaway signature test).** The blow-up mechanism was sinks failing to grow with particle size (#658/#660). A single-column probe sweeping the accumulation mode over D_g = 0.1→0.5 µm and measuring the TOTAL sink rate (sedi + drydep + wet at fixed met) catches any size window where growth outruns removal — as a test, without any long integration. #660's PR did this by hand for sedimentation; make it a pinned gate covering the summed sinks.

## Hooks

- `jcm/physics/aerosol/jam/emissions/flux_diagnostic.py` (the `DEPOSITED_SPECIES` ledger + per-step-reset keys) is the pattern to extend.
- `tools/release_validation/health.py` / the runner's chunk loop for the f64 per-chunk line.
- #660's `test_matches_cam_modal_aero_depvel_part` fixtures for the size sweep.

Refs #658, #660, #708.

[...]

Issue #658, opened 2026-08-19
JAM sulfate runaway: no removal for fine aerosol between ~300 and ~5 hPa (lifetime 2 d in the lower troposphere, 9,600 d in the UT)


duncanwp:
## Summary

The 730-day `echam-jam` T63L47 init-state run on merged `dev` (HEAD `dfc4672`, i.e. **after** #643 and #656) has a healthy physical climate but **accumulation-mode sulfate still runs away**, with onset in year 2. #656 fixed the SOA runaway (SOAG is now correctly zero); it did not fix this one.

Run: `/glade/derecho/scratch/duncanwp/jam_runs/jam_init_t63_l47/`, PBS 7145769, 730 d, NaN-free throughout.

## Physical climate is fine

```
PASS  NaN scan: 0/250 variables non-finite
PASS  toa_net_wm2 = 0.78        PASS  precip_mm_day = 2.45
PASS  cloud_cover = 0.61        PASS  near_surface_T = 284.49
PASS  burden_bc = 0.72   PASS  burden_du = 32.67   PASS  burden_poa = 4.04
FAIL  burden_so4 = 112.05 mg/m2 (gate 0.67-12)
FAIL  aod_550 = 0.36
```

## The runaway is species-, mode- and phase-discriminating

Global-mean column burdens [mg/m2]:

```
 day     so4     soa      bc      ss      du
   35    5.01    0.00    0.50    1.40   46.24
  185    4.87    0.00    0.71    1.85   31.29
  365    6.42    0.00    0.73    1.54   48.04     <- ~1 year, still fine
  425    9.41    0.00    0.72    1.94   61.96
  485   15.12    0.00    0.69    2.42   51.51
  545   34.86    0.00    0.74    1.90   23.14
  635   83.37    0.00    0.75    1.74   24.12
  695  141.94    0.00    0.69    2.37   37.35
```

Every other species is flat with a plausible seasonal cycle on **identical** transport, deposition and convection. Splitting sulfate by mode and phase [mg/m2]:

```
 day    g_so2  g_h2so4    m_so4_ait  m_so4_cor   m_so4_acc   mc_so4_acc
   45    3.478    0.007        1.274      0.089       3.599        0.338
  365    3.426    0.006        1.621      0.063       4.738        0.576
  485    1.902    0.007        1.226      0.045      13.851        3.088
  605    1.689    0.006        1.420      0.002      43.738        6.878
[...]

Issue #602, opened 2026-08-12
JAM cloud-borne aerosol: never populated, never deposited, but fully transported — and no physics-side aerosol vertical transport exists


duncanwp:
Two linked findings from profiling the pySES ne30 cost (#595). They are
independent of the dycore — the spectral path carries the same tracers — but
ne30 is where the cost is visible.

## 1. The cloud-borne population is incomplete, yet we pay full transport for it

`physics=echam-jam` declares **58 tracers**: 50 modal aerosol (`tracer_specs`
emits one number + one mass-per-species per mode, **doubled for the
cloud-borne mirror**, `jam/tracer_layout.py:55-78`), 4 sulfur gases, 4 cloud.
So **25 of the 58 are cloud-borne** `mc_*`/`nc_*`.

Measured by HLO cost analysis of one `dycore.step`: **tracers are ~94 % of
the step's FLOPs and ~93 % of its memory traffic.** The cloud-borne mirrors
are ~43 % of the tracer count, and at `tracer_substeps: 5` every one of them
is advected, PPM vertically remapped and hyperdiffused five times per
coupling step.

They are carrying almost nothing, because the population is only partly
wired up:

- **No activation transfer.** `activation/arg_term.py` writes only the
  diagnostics `activated_cdnc` and `activated_fraction` — nothing moves mass
  or number from interstitial to cloud-borne. Activation is the *only*
  physical source of cloud-borne aerosol, so the reservoir is never filled.
  The codebase already says so, in `chemistry/aqueous.py:288`:

  > *"Where a cell has no cloud-borne number (spin-up, or — today —
  > **everywhere**, since no activation transfer populates the cloud-borne
  > tracers yet) the production goes to INTERSTITIAL accumulation-mode
  > sulfate instead"*

- **No resuspension.** Nothing returns cloud-borne aerosol to interstitial
  when a droplet evaporates.

- **No deposition.** Sedimentation, wet dep and dry dep all build their name
  lists *without* `cloud_borne=True` (`sedimentation/sedi_term.py:156`,
  `wetdep/wetdep_term.py:233`, `drydep/drydep_term.py:114`), so cloud-borne
  mass has no removal path either.

The only writers are the MAM4 `qqcw` round trip (`microphysics/mam4_jax.py`)
and the aqueous SO4 term — which itself falls back to interstitial precisely
because cloud-borne number is zero.

**So we are transporting 25 near-empty fields at full cost, and the
[...]

PR #714, opened 2026-08-23
JAM wet scavenging keys to the microphysics' process-time ledger (#708)

### LINKED ISSUE #708: JAM scavenging in cells the microphysics empties: interstitial in-cloud rate keys to end-of-step cover, not the formation-time ledger
Raised by the Codex review of #707 (P1). Verified: partially real, pre-existing (not a #707 regression), and the suggested fix direction there ("preserve the process-time cloud fraction") would deviate from ECHAM — this issue records the reference-faithful alternative.

## The scenario

A stratiform cell whose condensate fully converts to precipitation within one step ends the step with `clouds.cloud_fraction = 0` (the ECHAM `paclc` write-back — under 2M since #693, under 1M since #707/#687). Two JAM consumers read that cover in the same step:

- `wetdep_term.py`: the IMPLICIT interstitial in-cloud pathway scales as `rate_ic_unit ∝ cf_clip` — zero cover ⇒ zero in-cloud scavenging of interstitial aerosol, in exactly the step whose `p_form` is largest. (The explicit cloud-borne pathway is fine: `rate_cb = p_form/condensate` is not cover-keyed.)
- `cloud_borne.py`: `target = cf·frac·(q_int+q_cb)` → cf = 0 relaxes the cloud-borne phase toward zero with `resuspension_timescale = 900 s` — a large fraction per 720–1800 s step — racing the same step's rainout.

## Why #707 did not change this

`echam-jam` composes only with `cloud_scheme: 2m`, whose cover semantics are identical before and after #707. And the write-back itself is reference-correct: ECHAM's own scavenging (`cloud_subm_2`, called after `column_processes`) receives the POST-write-back `paclc` too.

## The reference-faithful fix

ECHAM does not reconstruct scavenging rates from cover × grid-mean state: `cloud_subm_2` receives the process-time ledger — `zmlwc`/`zmiwc` (in-cloud condensate before precip formation, zeroed only where `paclc < clc_min`) and `zmratepr`/`zmrateps`/`zmsnowacl` (the in-cloud formation rates). jcm already exports the equivalents (`clouds.precip_formation_rate` is the grid-mean ledger; the 2M sweep also has `zmlwc`/`zmiwc` per level, currently not published). Keying the implicit interstitial pathway (and the CloudBorneExchange target during precipitating steps) to that ledger instead of `cf_clip` removes the emptied-cell dead zone without deviating from the cover semantics.

Needs its own evidence (a JAM A/B on burdens/lifetimes — interacts with #658, where under-scavenging is the standing suspect in the OPPOSITE direction, so measure before moving).

duncanwp:
**#660 has merged** (squashed into `dev` as 7762b53), so this PR now stands alone: the five commits in the diff are all and only this PR's work. The two are the sink half of #658 and were validated together (90-day T63L47 `echam-jam` lifetime A/B against `dev`; results in the thread).

Closes #708. Refs #658, #602, #713 (filed: budget-closure diagnostics for fast drift attribution). #712 (the 1M `mo_cloud` ledger port) was filed by this PR and then closed as moot by the review decision below.

## The problem (#708, Codex P1 on #707 — verified pre-existing, not a regression)

A stratiform cell whose condensate fully converts to precipitation in one step ends it with `cloud_fraction = 0` (the reference-correct `paclc` write-back) and condensate 0. Both JAM consumers read that post-microphysics state:

- `WetScavenging`'s implicit in-cloud rate scaled with the **end-of-step cover** — zero removal in exactly the step with the largest removal;
- `CloudBorneExchange` drained the reservoir toward zero on `resuspension_timescale` whenever cover read 0 — and cover 0 cannot distinguish *evaporated* (release the aerosol) from *rained out* (it left with the precip). Running just before wetdep, the drain let up to `1−exp(−Δt/τ)` ≈ 86% (Δt=1800 s) of the cloud-borne reservoir escape scavenging through exactly the fully-converting cells.

The explicit cloud-borne phase (the CAM/MAM4 `qqcw` inheritance) is **kept** — MAM4's own convention (remove at the conversion fraction, resuspend where cloud water evaporates) is precisely the ledger keying this PR implements, so the fix and the phase point the same direction.

## What changed

**1. The 2M scheme publishes the ECHAM-HAM scavenging interface** (`ScavengingLedger` → seven `CloudData` fields). Everything `cloud_subm_2` receives was already computed in the #707 scan as discarded locals: `zmlwc`/`zmiwc` (in-cloud pools captured *before* formation depletes them, post the faithful emptied-cell zeroing), `zmratepr`/`zmrateps`/`zmsnowacl` (in-cloud formation rates), the process-time cover, and the `zxlevap+zxievap` condensate-evaporation ledger. Two fidelity notes: `zmrateps` is now seeded from `sedimentation_ice` (ECHAM 1243→1703 — sedimenting ice **is** a scavenging carrier in ECHAM-HAM; the cold chain overwrites the seed where it runs, per the Fortran MERGE at 3310), and the evaporation ledger is exported as the resuspension key.

**2. Wetdep in-cloud removal is HAMMOZ `prep_wetdep_hydro`** — `peffwat = (zmratepr+zmsnowacl)·Δt/zmlwc`, `peffice = zmrateps·Δt/zmiwc`, clipped to [0,1] and split by the in-cloud ice mass fraction. Numerator and denominator are captured at process time, so the fraction is bounded **by construction** — no unbounded rate ratio being "accidentally correct" via a floor in near-empty cells. The implicit interstitial in-droplet share keys to the process-time cover.

**3. Cloud-borne resuspension keys to the evaporation ledger.** The released share per step is the share of the droplet population that evaporated, `E/(E+pool)`, capped by `1 − f_form` (the same step's rainout claim) so the two sinks cannot jointly overdraw the reservoir. A sky cleared by evaporation resuspends everything in one step (CAM's instant resuspension); a sky cleared by rainout resuspends nothing; WBF/homogeneous glaciation move condensate within the pool and release nothing — the aerosol rides into the ice and meets the snow pathway (#686) in the ledger. Only cells with no cloud process at all (advected-in `q_cb` in clear air) keep the slow timescale drain.

## The one documented deviation from the reference

ECHAM-HAM zeroes `zmlwc`/`zmiwc` *after* the `paclc` write-back (mo_cloud_micro_2m.f90:3655 → 3660), so `prep_wetdep_hydro` maps the fully-converting cell to `peffwat = 0` — **the reference itself has the dead zone** (one leapfrog step's worth of scavenging missed per full conversion). jcm keeps the faithful zeroing but reads the marker (zero pool + positive formation) as scavenged-fraction **1**, which is what the issue prescribed and what physical bookkeeping requires. Called out in `ScavengingLedger`, `incloud_scavenged_fractions`, and the design doc.

## JAM requires the 2M scheme (review decision)

An earlier revision kept a `scavenging_ledger` constructor switch so JAM could fall back to the legacy cover-keyed reconstruction under the 1M scheme. Review (this thread) ruled JAM+1M an unsupported configuration — prognostic aerosol with aerosol-blind microphysics — so f222619 removes the switch chain and the legacy reconstruction entirely: the ledger pathways are unconditional, and `echam_physics(aerosol_module='jam')` rejects `cloud_scheme='1m'` at compose time. (Verified the combination otherwise constructs and runs — nothing structurally requires 2M, it would just silently scavenge nothing stratiform off the 1M scheme's all-zero ledger fields — so the explicit error is the only guard.)

## Verification

- `ruff` clean; fast suite 1960 passed @ 94.8% (≥90) after f222619; slow tests in the touched areas (aerosol/clouds/echam) 14 passed — all locally at CI parity.
- New gates: `TestScavengingLedger2M` (in-cloud semantics, boundedness-by-construction, the sedimentation seed, term publication), `FormationLedgerTest` (the emptied cell scavenges under the ledger; process-cover keying; fraction bounds incl. the fraction-1 marker), five `CloudBorneExchangeTest` ledger gates (rained-out ≠ evaporated; partial-evaporation share; the rainout cap; phase transfer releases nothing), and `test_jam_module_rejects_1m`.
- All pre-existing wetdep/cloud-borne/2M/1M tests pass.
- 90-day T63L47 `echam-jam-aerocom` A/B (this stack vs `dev`, identical config/init/seeds) in flight for per-species burden/sink-flux/lifetime ratios — thread below when done.

### DIFF
--- a/docs/source/design/lohmann_2m_column_processes.md
+++ b/docs/source/design/lohmann_2m_column_processes.md
@@ -1,8 +1,6 @@
 # The Lohmann 2M scheme is ECHAM's `column_processes` loop
 
-*(2026-08; issues #662, #667, #685, #686, #687, #688, #689; parent audit #614)*
-
-## The decision
+## The design
 
 `cloud_microphysics_2m` runs the **entire** two-moment process chain inside one
 top-down `lax.scan`, one scan step per level, in the section order of ECHAM's
@@ -10,32 +8,20 @@ top-down `lax.scan`, one scan step per level, in the section order of ECHAM's
 ledger, `update_tendencies_and_important_vars`) has no cross-level coupling and
 runs vectorized on the stacked per-level outputs afterwards.
 
-The previous layout split the chain: "level-independent" processes
-(condensation partition, freezing, WBF, warm/cold precipitation formation) ran
-vectorized outside the sweep, and only sedimentation/melting/sublimation ran
-inside it. That looked like a harmless performance-friendly factorization. It
-was not, because in the reference **every process at level `jk` consumes the
-precipitation state the levels above produced *this step*** — the rain/snow
-fluxes (`prfl`/`pssfl`), the precipitation cover (`zclcpre`), and the falling
-ice flux (`zxiflux`) are loop-carried. Splitting the chain silently severed
-those couplings:
-
-- accretion by rain/snow from above read `qr`/`qs` tracers that no term
-  declared any more — both identically zero, so the pathway was dead code and
-  precipitation formation was understated ~3× (#662 finding 5, and through
-  `rate_cb = p_form/(qc+qi)` the leading suspect for the cloud-borne aerosol
-  drainage failure in #658);
-- the cloud∩precipitation overlap `zclcstar = min(paclc, zclcpre)` could not
-  be formed, so accretion geometry fell back to `paclc` (#685);
-- ice created mid-step (deposition, freezing, WBF) never met its
-  aggregation/riming sink that step (#686);
-- warm rain ran *first*, before condensation and activation, matching neither
-  ECHAM nor CAM (#667 finding 4).
-
-The restructure is therefore not an optimization choice reversed — it is the
-minimal faithful shape. Wall-clock cost is unchanged at leading order: the
-same arithmetic moves from a `(nlev,)`-vectorized context into the scan body,
-and the scan already existed.
+The monolithic sweep is forced by the reference's data flow: every process at
+level `jk` consumes the precipitation state the levels above produced *this
+step* — the rain/snow fluxes (`prfl`/`pssfl`), the precipitation cover
+(`zclcpre`), and the falling ice flux (`zxiflux`) are loop-carried. Lifting
+apparently "level-independent" processes (the condensation partition,
+freezing, WBF, precipitation formation) out of the sweep severs those
+couplings: accretion by rain and snow from above loses its collector fluxes,
+the cloud∩precipitation overlap `zclcstar = min(paclc, zclcpre)` cannot be
+formed, ice created mid-step (by deposition, freezing, or WBF) never meets
+its aggregation/riming sink that step, and warm rain cannot run after
+condensation and activation, which is where both ECHAM and CAM place it. The
+whole chain therefore lives in the scan. Wall-clock cost is unchanged at
+leading order: the same arithmetic runs inside the scan body rather than in a
+`(nlev,)`-vectorized context.
 
 ## What the sweep does per level
 
@@ -48,10 +34,26 @@ geometry (`zclcstar`, `zauloc`, the Marshall–Palmer inversions
 `zxrp1`/`zxsp1`) → 7.1 warm rain → 7.2 cold precipitation → 7.3 flux update.
 
 One deliberate deviation, reviewed and kept: sedimentation runs **before**
-melting (MG/PUMAS order, `micro_pumas_v1.F90`), with the running ice tendency
-threaded through the melt routine so the two sinks cannot claim the same mass
-(#662 finding 2). ECHAM melts first; both orderings are internally consistent
+melting (the MG/PUMAS order, `micro_pumas_v1.F90`), with the running ice
+tendency threaded through the melt routine so the two sinks cannot claim the
+same mass. ECHAM melts first; both orderings are internally consistent
 ledgers.
+
+Formulation choices inside the sweep, for provenance:
+
+- Grid-scale condensation/evaporation is the section-5 `zqcdif` closure with
+  ECHAM's Newton saturation damper (`zqcon`; 0.36–0.93 through the
+  troposphere), so no external saturation adjustment is composed alongside
+  the scheme and no supersaturation survives a step.
+- Clear-sky condensate — including cells whose cover reached zero this step —
+  evaporates through `zxlevap`/`zxievap` verbatim.
+- Accretion by rain and snow from above converts the loop-carried fluxes to
+  local mixing ratios with the Marshall–Palmer inversions (`zxrp1`/`zxsp1`).
+- The WBF threshold updraft `peta` is ECHAM's diffusional-growth ζ
+  (`mo_cloud_micro_2m.f90` line 856), recomputed after freezing so the
+  post-freezing crystal population sets the threshold.
+- Diagnostic cirrus ICNC (`nic_cirrus = 1`) uses the Schumann volume-mean
+  radius, in metres.
 
 ## The state-splitting convention
 
@@ -75,18 +77,6 @@ The ledger reconstruction is identical either way:
 guard bounds the true end-of-step state and the host's tendency sum
 telescopes exactly as ECHAM's INOUT accumulation.
 
-## What this made live (previously inert or miscalibrated)
-
-| pathway | before | after |
-|---|---|---|
-| internal saturation adjustment | `zqcon ≈ 1/50…1/650` (zdqsdt ×1000; #667.1) | Newton damper 0.36–0.93; 10 % supersat → 1 % in one step |
-| grid-scale condensation | external Sundqvist bolt-on (double-adjustment risk) | ECHAM section-5 `zqcdif` closure in-scheme; bolt-on removed |
-| clear-sky condensate sink | none (`pxlevap = pxievap = 0`) | `zxlevap`/`zxievap` verbatim; cf=0 cells evaporate in one step |
-| rain/snow-from-above accretion | dead (`qr`/`qs` ≡ 0) | Marshall–Palmer inversion of carry fluxes |
-| WBF threshold updraft | fed a 0–1 clip as `peta` | ECHAM's diffusional-growth ζ (line 856); recomputed post-freezing |
-| diagnostic ICNC (`nic_cirrus=1`) | `prid` in µm (r³ off by 10¹⁸) → pinned at floor | Schumann volume-mean radius in metres |
-| cold-chain gate | step-start `qi > ccwmin` | `ll_cc` + in-scan `zxib > cqtmin` (#686) |
-
 ## Deliberate omissions (tracked)
 
 - **Large-scale vertical velocity is not plumbed** (`zvervx` is TKE-only; the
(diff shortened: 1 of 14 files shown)


jcm/physics/convection/tiedtke_nordeng/tiedtke_nordeng.py
```diff
@@ -976,9 +1457,37 @@ class TiedtkeConvection(PhysicsTerm):
...
+            +physics.terms.tiedtke_convection.params.cu_lmfmid=false
+
+        Column-mode callers with no dycore at all are unaffected: they
+        bypass the contract check, and a zero omega correctly means "no
+        resolved ascent, so no mid-level convection".
+        """
+        params = params or ConvectionParameters.default()
+        self.params = nnx.Param(params)
+        if bool(params.cu_lmfmid):
+            self.requires_dycore_fields = ("omega",)
```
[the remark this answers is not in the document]
duncanwp: Confirmed real — but this is dev code (#690's `cu_lmfmid` omega contract, commit 66b2d75), untouched by this PR; it appears in the diff only because this PR is stacked on #660's branch and merges current dev. Verified: `pyses_dycore_test.py:361` documents pySES publishes no omega, and `_pyses_preset` lacks the `cu_lmfmid=false` override, so `ma-ne30-l{47,95}` fail at Model construction. Filed as #715 with both fix options (preset override now, #698 omega provider as the real fix).

tools/profile_terms.py
```diff
@@ -0,0 +1,838 @@
...
+    probes = {profiling.DYNAMICS, profiling.BRIDGE_TO_PHYSICS,
+              profiling.BRIDGE_TO_DYNAMICS}
+    # EVERY probe must be complete, not just the best of them. The probes are
+    # ordered within a step (bridge_to_physics runs before dynamics), so a
+    # buffer that fills partway through the final step leaves the earlier probe
+    # at the full count while a later one is short; taking the max would accept
+    # that trace and then divide short totals by the full step count.
+    cycle = result.get("radiation_subcycle_steps", 1)
+    seen = {label: n for label, n in att.steps_seen.items() if label in probes}
+    if seen and min(seen.values()) < result["steps"]:
```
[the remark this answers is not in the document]
duncanwp: Confirmed real — dev code (#709, commit 0ee92ea), untouched by this PR (stacked-diff artifact, as above). Verified: `if seen and min(...)` skips the completeness gate entirely when attribution recovers no probe labels, and checks only the subset of the three mandatory probes it happened to find. Filed as #716 (require `seen.keys() == probes`, hard-error on a missing probe, regression test with stripped metadata).

duncanwp:
Codex loop closed: both findings are **real, and both are pre-existing dev code** (#690's Tiedtke omega contract and #709's profiler gate) that appear in this diff only because the stacked base predates the dev merge — neither file is touched by this PR's four commits. Each is verified against the code and filed with a fix plan: #715 (ma-ne30 presets vs `cu_lmfmid`) and #716 (profile_terms completeness gate). Threaded replies on both inline comments. No changes to this PR.

duncanwp:
**Status update from the lifetime A/B — one real defect found in this PR, fixed and pushed (8486954), plus a standalone runaway on the base branch (#660, reported there).**

The 90-day T63L47 validation (production emissions) caught the full stack exploding every activatable aerosol species exponentially from day ~32. Bisection (four runs, table on #660): the primary bomb is the base branch's un-capped activation target — `dev`+this-PR-only stays far from the runaway — but this PR's original resuspension keying **amplified it**: keying the downward exchange to the evaporation ledger *everywhere* zeroed it in persistent non-evaporating cloud, turning the reservoir into a ratchet (rise toward the activation target, never fall; `mc_du_acc` 14× the base-branch-alone loading by day 30).

The fix (8486954) scopes the ledger keying to where it belongs — the **cleared-sky branch** (cover < 1e-3, target = 0), which is exactly where the #708 race lived: rained-out cells still don't resuspend, evaporated cells still release in one step, clear-air advected `q_cb` keeps the slow drain. Under persistent cloud the pre-existing equilibrium relaxation is restored. New `test_persistent_cloud_still_relaxes_excess_down` pins the under-cloud relaxation law (the anti-ratchet gate). Full local gate re-run: ruff clean, fast 1939 @ 94.9%, slow 86 @ 84.7% (one unrelated RRTMGP fast-suite failure under a heavily loaded box verified passing standalone — xdist load flake).

A 90-day validation of the full stack + this fix + the #660 target commit reverted is in flight; the lifetime table lands here when it completes. Note for the eventual merge order: **this PR should not merge while its base carries the un-reverted activation-target commit** — evidence and suggested action on #660.

The parts of the stack already validated by the same runs: coarse-mode dust and sea salt drop 2–4× in burden with τ_dry 13 → 1.1 d (#660's moment-weighted settling doing its job), and the day-0–30 window of every run shows the ledger-keyed scavenging behaving (burdens tracking or below dev with the sulfur chain live).

duncanwp:
Correction to the bisection table now probe B's full 60 days are in: `dev` + this PR's ORIGINAL commits (capped target) does not stay clean — the ratchet alone drives a bounded, oscillatory pathology (dust burden 141→363→225 mg/m² over days 30–60, max `mc_du_acc` spiking to 1.4e-4, ~10³ over dev). So the original resuspension keying was a real standalone defect, not merely an amplifier of the #660 target change; the exponential runaway needs both. Either way 8486954 removes the mechanism — the definitive series is the in-flight fixed-stack run (probe C), reporting here when its 90 days complete.

duncanwp:
Bisection final (full matrix on #660): this PR's fixed consumers are bounded and conserving on `dev` alone, but **moments (#660) × these consumers explodes even with the ratchet fix and the #660 target commit reverted** — one of two independent sufficient pairings with the moment-weighted settling (the other needs no #714 code at all). Merge order: blocked on the interaction diagnosis on #660; the #713 budget instrumentation is the recommended tool. The lifetime table this PR promised is meaningful only after that — the day-0–30 windows and the sea-salt-only pair (where all pathways behaved: ledger-keyed scavenging active in emptied cells, sea-salt τ 0.42→0.26 d, coarse dust 3× lower burden) are the interim validation.

duncanwp:
Unblocked pending #720: the moments-interaction 'explosion' was the semi-Lagrangian transport mass leak (measured and fixed there — full stack incl. this PR's consumers stable and budget-closed through day 75+ with the fixer). Merge order: #720 → #660 → this.

duncanwp:
Rebased onto `dev` after #660 merged.

#660 was **squash-merged**, so `dev` carries it as a single commit (7762b53) while this branch still held its nine original commits — git saw two unrelated histories touching the same lines, hence the conflict. Rebasing cut at the dev-merge commit rather than at #660's tip, so only this PR's five commits replayed:

```
git rebase --onto origin/dev 036204b fix/jam-scavenging-formation-ledger-708
```

Zero conflicts. Verified before pushing:
- `git range-diff` — all five patches identical to their pre-rebase versions (`=` on every row)
- diff vs `dev` is exactly the 12 scavenging-ledger files, no sedimentation content duplicated from #660
- `ruff check jcm/` clean; `lohmann_2m`, `wetdep`, `cloud_borne` suites pass (137 tests)

Pre-rebase tip was `8486954` if anything needs recovering.

docs/source/design/lohmann_2m_column_processes.md
```diff
@@ -116,3 +116,41 @@ fusion heat from below rather than trusting closure alone.
 1M and 2M schemes (#687; documented on `CloudData`), and the 2M
 negative-mass repair is exported as `clouds.negative_mass_repair` [W/m²]
 (#689) so its sign-definite heating is measurable in any run.
+
+## The wet-scavenging interface (#708)
+
+ECHAM-HAM does not reconstruct aerosol scavenging from cover × grid-mean
```
duncanwp: Don't tell me what it doesn't do; don't refer to previous not-working implementations and don't reference specific issues or PRs unless critical. This is a design document supposed to reflect the design of the implementation and why certain choices were made for scientific provenance - not a log

jcm/physics/aerosol/jam/wetdep/wetdep_term.py
```diff
@@ -13,9 +13,14 @@
...
 
 * **In-cloud nucleation scavenging** — aerosol residing in cloud droplets is
-  removed at the local rate cloud condensate converts to precipitation,
-  ``precip_formation_rate / (qc + qi)`` (both grid-mean, so the cloudy-area
-  factor cancels inside the ratio).
+  removed at the fraction cloud condensate converted to precipitation this
+  step, read from the cloud scheme's process-time scavenging ledger
+  (``incloud_scavenged_fractions`` — HAMMOZ's ``peffwat``/``peffice``, #708),
+  which stays alive in cells the microphysics emptied. The legacy
+  reconstruction ``precip_formation_rate / (qc + qi)`` (grid-mean, cover
```
duncanwp: The 1m scheme is never used with online aerosol - it makes no sense. Remove this switch and the whole 'legacy reconstruction'. I don't think you can even construct JAM with the 1m scheme because I suspect at least one of the components needs something 1m doesn't provide (activation?), but we should check

jcm/physics/aerosol/jam/wetdep/wetdep_term.py
```diff
@@ -332,20 +440,37 @@ def __call__(self, state, diagnostics, forcing, terrain):
...
-        # The implicit stratiform rate is weighted by the cloudy area
-        # fraction: the grid-mean p_form/condensate ratio is the IN-CLOUD
-        # conversion rate (cf cancels between them), but only cf·af of the
-        # grid-mean interstitial tracer is in droplets. ``in_cloud_rate`` is
+        # Stratiform in-cloud (nucleation) scavenging rates. With the
+        # process-time ledger (#708) the per-step scavenged fraction of an
+        # in-droplet tracer is HAMMOZ's peffwat/peffice split by the phase
+        # mass fraction — bounded by construction, and alive in cells the
+        # microphysics emptied (post-write-back cover 0), where the legacy
+        # reconstruction below reads cover 0 x end-of-step condensate 0
```
duncanwp: Remove reference to legacy

jcm/physics/aerosol/jam/cloud_borne.py
```diff
@@ -200,14 +231,77 @@ def __call__(self, state, diagnostics, forcing, terrain):
...
+        phi_up = -jnp.expm1(
+            -dt / jnp.maximum(params.activation_timescale / cloudy_cf, 1.0)
+        )
+        phi_slow = -jnp.expm1(
+            -dt / jnp.maximum(params.resuspension_timescale, 1.0)
         )
-        # 1 − exp(−Δt/τ) ∈ [0, 1]: the move never overshoots the target, so
-        # neither phase can go negative (|Δ| ≤ |target − q_cb| ≤ donor).
-        phi = -jnp.expm1(-dt / jnp.maximum(tau, 1.0))
+        if self._evaporation_ledger:
```
duncanwp: Can this logic be simplified now?

jcm/physics/aerosol/jam/wetdep/wetdep_term.py
duncanwp: Removed in f222619 — the switch, the legacy reconstruction (`in_cloud_rate`), and its `CloudBorneExchange` twin (`evaporation_ledger`) are all gone; the ledger pathways are unconditional and the factory no longer takes `scavenging_ledger`.

Checked construction as you asked: JAM+1M **does** construct and run today — nothing structurally requires 2M (ARG's `activated_cdnc` and `ice_nuclei` are simply ignored by 1M, and `activated_fraction` comes from the MACv2-SP term composed alongside). It only 'worked' through the fallback: with the ledger unconditional it would run but silently scavenge nothing stratiform, since 1M leaves the `CloudData` ledger fields all-zero. So `echam_physics` now rejects `aerosol_module='jam'` with `cloud_scheme='1m'` at compose time (ValueError, test added). Closed #712 (the 1M ledger port) as moot — its only consumer was this combination.

docs/source/design/lohmann_2m_column_processes.md
duncanwp: Rewritten as a design document in f222619: it now states what the implementation is and why — the monolithic scan forced by the reference's loop-carried precipitation state, the per-level section order, the formulation choices with their Fortran provenance, the state-splitting convention, and the ledger interface with its one documented deviation. The fix-campaign narrative, the before/after table, and non-critical issue references are gone; issue numbers remain only on the tracked deliberate omissions.

jcm/physics/aerosol/jam/wetdep/wetdep_term.py
duncanwp: Removed in f222619 — the legacy path itself is deleted (thread above), and the comment now describes only the current ledger keying.

jcm/physics/aerosol/jam/cloud_borne.py
duncanwp: Yes — with the `evaporation_ledger` switch gone it collapses to a single select in f222619: `phi_down = where(cleared & live, min(f_evap, 1 - f_form), phi_slow)` — ledger release where the sky cleared and a cloud process ran, the equilibrium relaxation timescale everywhere else.

Issue #741, opened 2026-08-28
load_states_from_xarray reads surface-first output files into the top-first physics frame — no flip


duncanwp:
## What is wrong

`jcm/utils.py::load_states_from_xarray` passes `ds[var].values` straight into a
`PhysicsState`. Its natural input is a file written by
`ModelPredictions.to_xarray()`, whose `level` axis is **surface-first** and has
been since long before #710. The physics and the dinosaur dycore are
**top-first**. Nothing flips in between, so a state loaded this way is
vertically inverted.

## Concrete instance

`jcm/data/test/echam_t63l47/spinup_state.nc` is written by
`generate_default_stats.py:213` from `spin_up.to_xarray().isel(time=-1)` and
read back by `_load_spinup_state()` as the initial condition for the slow test
`model_test.py::test_echam_model_default_statistics`. Checking the file:

```
level        [0.99615 0.98265]  ...  [4.28e-05 9.94e-06]   # surface-first
geopotential [3822.8  4951.3 ]  ...  [836437.1 978025.4]   # surface-first
```

so the model receives geopotential increasing with index, i.e. an upside-down
column.

**Impact on that test today is probably nil, and this is not a regression.**
The saved state is isothermal 288 K with zero winds and zero humidity, so
geopotential is the only field with vertical structure, and the reference
`default_statistics.nc` was generated through the same inverted path — the test
is self-consistent. The defect matters because (a) the reader is wrong in
general and would silently invert any genuinely spun-up state, and (b) the
"spun-up state" is not actually spun up, which is worth knowing separately.

## Where the hooks are

- `jcm/utils.py::load_states_from_xarray` — the reader.
- `jcm/cf_metadata.py::orient_surface_first` — the inverse operation already
  exists; a reader-side counterpart (`orient_top_first`) is the natural fix,
  keyed off the `level` coordinate rather than assumed.
- Since #710 the file is self-describing (`level` carries `positive` and
  descending sigma values), so the reader can *detect* the input orientation
  instead of assuming one. Pre-#710 files also have descending `level` values,
  so the same check works on them.
- `jcm/data/test/echam_t63l47/generate_default_stats.py:95` `_load_spinup_state`
  — regenerating `spinup_state.nc` and `default_statistics.nc` is part of the
  fix, since the stored reference bakes in the current behaviour.

Related: #739 (`PrescribedStateModel.to_xarray` writes the same `level` dim
name in the opposite, top-first frame, and round-trips through this same
[...]

Issue #740, opened 2026-08-28
Physics diagnostics reach the output netCDF with no units or standard_name


duncanwp:
## What is missing

Most physics diagnostics land in the output file with `attrs == {}`. On a
T63L47 ECHAM+RRTMGP run that includes the whole radiation flux set —
`radiation.lw_flux_up`, `radiation.lw_flux_down`, `radiation.sw_flux_up`,
`radiation.sw_flux_down`, the heating rates — plus the cloud, convection and
JAM aerosol diagnostics. A reader cannot tell W/m² from K/s, or kg/kg from
kg/m², without reading the source.

Two mechanisms attach metadata today and neither covers these:

- `jcm/predictions.py:302-315` attaches `units`/`description` from the
  per-physics CSV tables (`jcm/dynamics_units_table.csv`, plus whatever
  `Physics.units_table_paths()` returns). The composable ECHAM terms contribute
  few entries, so most diagnostic keys miss.
- `jcm/cf_metadata.py::_VARIABLE_ATTRS` (added in #710) sets CF
  `standard_name`/`units`/`long_name`, but only for the vertical-coordinate
  neighbourhood — `pressure_*`, `height_*`, `surface_pressure` — and the core
  prognostics.

## What to do

Give each `PhysicsTerm` a way to declare units and, where one exists, a CF
`standard_name` for the keys it `provides`, so the metadata lives next to the
code that computes the quantity rather than in a CSV that drifts. The
radiation fluxes have exact CF names
(`surface_upwelling_longwave_flux_in_air`,
`toa_outgoing_longwave_flux`, ...) worth using where the diagnostic really is
that quantity, and a plain `units` + `long_name` elsewhere.

## Where the hooks are

- `jcm/physics/physics_term.py` — `PhysicsTerm.provides` is the existing
  per-term declaration to hang units off.
- `jcm/physics/composable_physics.py:570` `units_table_paths` — the current
  aggregation point.
- `jcm/cf_metadata.py::apply_cf_attributes` — where per-variable attributes are
  stamped onto the Dataset; a term-supplied mapping would feed in here.
- `jcm/physics/radiation/radiation_types.py:159-167` — the flux field set.

Found while adding CF metadata to the output coordinates in #710, which
deliberately scoped itself to the coordinates and the vertical variables.

Issue #739, opened 2026-08-28
PrescribedStateModel.to_xarray writes a top-first `level` axis, opposite to every other output product


duncanwp:
## What is missing

`jcm/prescribed_state_model.py::PrescribedStateOutput.to_xarray` emits column
variables under the dim name `level` in the **top-first physics frame**, with no
coordinate values and no attributes. Since #710, every other output product —
the trajectory file (`ModelPredictions.to_xarray`), both dycore backends'
`to_xarray`, and observer profile curtains — writes `level` **surface-first**,
with a nominal-sigma coordinate and CF `positive = "down"`.

So the same dim name means opposite things depending on which product you
opened, which is exactly the class of trap #710 was about; it just happens to
sit in the offline-tendency dump rather than the trajectory file.

## Why it is like that today

The dump round-trips: `jcm/utils.py::load_states_from_xarray` reads it straight
back into a `PhysicsState` without flipping either, so the pair is
self-consistent as long as nobody pairs it with a trajectory file. The
docstring on `to_xarray` now says so explicitly.

## Where the hooks are

- `jcm/prescribed_state_model.py::PrescribedStateOutput.to_xarray` — `_dims_for`
  hardcodes the positional dim names.
- `jcm/utils.py::load_states_from_xarray` — the reader; would need the matching
  flip if the writer changes.
- `jcm/cf_metadata.py::finalize_output` — the shared entry point the other
  products use. It needs the hybrid `(a, b)` table, which this struct does not
  currently carry (it has no `coords`); threading `coords` through
  `PrescribedStateOutput`, or storing the table on it at construction, is the
  main piece of work.

## Options

1. Flip the writer and the reader together (keeps the round-trip, makes the
   file consistent with everything else). Breaks any existing dump on disk.
2. Rename the dim to something that cannot be confused with the file
   convention (e.g. `level_physics`), leaving orientation as-is. Cheaper, but
   leaves two vertical conventions in the codebase.

Option 1 is the one that removes the trap rather than labelling it.

Found while unifying the output vertical convention in #710.

PR #742, opened 2026-08-28
Output netCDF: one vertical direction, and CF metadata that says which

### LINKED ISSUE #710: Output netCDF runs its two vertical dimensions in opposite directions, with no metadata to say so
## What is wrong

In a single JCM output file, the full-level dimension (`level`) is **surface-first** and the interface dimension (`level_i`) is **TOA-first**. Every variable on each dimension is internally consistent, so nothing looks wrong until you combine the two — at which point the result is silently upside down.

From a T63L47 ECHAM+RRTMGP snapshot (`traj_day10.nc`, 47 levels / 48 interfaces), global means:

| variable | dim | index 0 | last index | orientation |
|---|---|---|---|---|
| `pressure_full` | `level` | 100701 Pa | 0.99 Pa | **surface-first** |
| `level` (coord) | `level` | 0.99615 | 9.9e-06 | **surface-first** |
| `pressure_half` | `level_i` | 0.0 Pa | 97028 Pa | **TOA-first** |
| `height_half` | `level_i` | 93066 m | 347 m | **TOA-first** |
| `radiation.lw_flux_up` | `level_i` | 300.1 W/m² | 364.9 W/m² | **TOA-first** (OLR at 0, surface at −1) |
| `radiation.sw_flux_down` | `level_i` | 324.7 | 240.8 | **TOA-first** |
| `radiation.lw_flux_down` | `level_i` | 0.06 | 223.8 | **TOA-first** |

`pressure_full` and `pressure_half` disagree in **18432 of 18432 columns** — it is systematic, not a corner case.

## Why it bites

`level_i` is a **bare integer index 0..47 with no coordinate values and no attributes** (`ds['level_i'].attrs == {}`), and `pressure_half` / `pressure_full` carry no attributes either. So there is nothing in the file that tells a reader the two dimensions run opposite ways; the only way to find out is to inspect the pressure values of each group separately.

The natural operations all pair the two dimensions and are all silently wrong:

```python
dp = np.diff(ds.pressure_half, axis=...)        # TOA-first order
heating = (g/cp) * np.diff(F_net) / dp          # fluxes: TOA-first, fine
mass_weighted = (field * dp).sum()              # field on `level`: SURFACE-first -> reversed
```

Computing a heating rate from the saved fluxes and comparing it against a saved full-level temperature, or mass-weighting a tracer with `diff(pressure_half)`, both come out vertically flipped with no error and a plausible-looking magnitude.

This is the same class of trap `CLAUDE.md` already warns about after the 2026-07-05 "oxidant flip" investigation — but that note documents only the **full-level** convention ("the model's saved output is surface-first"). It does not say that interface variables are the opposite, so following it faithfully still produces the bug.

## Suggested fix

Either is defensible; pick one deliberately:

1. **Write interfaces surface-first too**, so one file has one direction. Cleanest for users, but flips the sign convention of anything already reading `pressure_half`, and the radiation flux diagnostics are new enough that now is the cheap moment to do it.
2. **Give `level_i` a real coordinate and attributes** (interface sigma or reference pressure, plus a `positive = "down"`/`"up"` attribute per CF conventions) and add `positive`/`long_name` to `pressure_full`/`pressure_half`. Keeps existing readers working and makes the orientation self-describing, which is what the file lacks today.

Whichever is chosen, the convention should be asserted in a test on a written file rather than left to documentation, and the `CLAUDE.md` note should be extended to name both dimensions.

## Existing hook

`tools/radiation_emulator/generate_training_data.py::_orient_toa_first` already defends against this: it orients the full-level and interface groups **each against their own pressure array** rather than a single file-wide flag, precisely because they disagree. That is the workaround, not the fix.

## How it was found

While building the radiation-emulator training pipeline (#702), which reads both groups from model output and has to put them in one frame.

duncanwp:
Closes #710, #739, #740, #741, #744.

The issue offered two fixes and said to pick one deliberately. This does **both**, because they answer different halves of the problem: flipping the interfaces makes the file *correct*, and the CF metadata makes it *checkable*. Doing only the flip would leave the next silent-orientation bug just as invisible; doing only the metadata would leave every naive pairing of the two axes wrong-but-documented.

## What was wrong

`ModelPredictions._trajectory_dataset` reversed the `level` axis inline and left `level_i` in the physics-internal TOA-first frame it arrived in. So in one file `pressure_full` ran surface→top and `pressure_half` ran top→surface, in 18432 of 18432 columns, with nothing in the metadata to say so.

The real defect is not a missing flip — it is that the flip was expressible per-variable at all.

## What changed

**One place owns the conversion.** New `jcm/cf_metadata.py`; every backend's `to_xarray` ends in `finalize_output(...)`, and no other code flips a vertical axis on the way out. `jcm/utils.py::data_to_xarray` stays a shape→dims mapper with no opinion about direction. A new interface diagnostic is now oriented correctly by construction rather than by someone remembering.

**Both axes surface-first.** `level[k]` lies between `level_i[k]` and `level_i[k+1]`, so `-diff(pressure_half)` is the Δp for a `level` field as written.

**The file says so.** Both axes are real coordinate variables holding nominal sigma (`a/p0 + b`) from one hybrid table, with `standard_name = atmosphere_hybrid_sigma_pressure_coordinate`, `units`, `axis = "Z"`, `positive = "down"`. The hybrid `(a, b)` tables travel with the file (`hybrid_a_full`/`hybrid_b_full`/`hybrid_a_half`/`hybrid_b_half`) in its own surface-first order, so `p = a + b·p_s` is reproducible from the file alone, and `formula_terms` names them. `lat`/`lon`/`time` and the other axes get their standard names too; `pressure_*`/`height_*`/the core prognostics get `standard_name`/`units`.

**Same fix in the other two output products.** The pySES backend renamed `level_interface` → `level_i` (it also had no coordinate values at all) and now shares the metadata construction. Observer profile curtains were emitting whole columns top-first under a bare `level` axis — same trap, different product — so `to_dataset` now goes through the same finalisation. `sample`'s contract is unchanged: `preds.observations` stays raw physics-frame.

**The redundant test is gone.** `tools/jam_burden_report.py::_layer_dp` detected the mismatch by comparing pressures at each end of each axis and flipped `dp` when they disagreed. That branch is now dead.

## Decisions worth second-guessing

- **`positive` is on the coordinate variables only, not on `pressure_full`/`pressure_half`.** The issue asked for it on both. CF-1.11 §4.3 defines `positive` for vertical *coordinate* variables, and `pressure_*` are data variables; a checker would flag it. Their orientation is that of the axis they sit on, and both axes now agree and are labelled — which is the CF-correct place to say it. Reasoning is in the module docstring.
- **`formula_terms` is emitted only when `surface_pressure` is in the file.** A dynamics-only run has no `ps` variable, and a dangling `ps:` reference is worse than no `formula_terms`.
- **`level` is overwritten by `attach_vertical_coordinates`, not left as the dycore wrote it.** Same values, but it guarantees both axes provably come from one table rather than two paths that happen to agree today.
- **One `standard_name` for both coordinate families.** Pure sigma is the `a = 0` case of the hybrid formula, so `atmosphere_hybrid_sigma_pressure_coordinate` is correct for both rather than branching.

## Breaking

Anything reading `pressure_half`, `height_half` or the radiative fluxes from a **new** file gets the opposite order to before — that is the point. Files written before this are distinguishable: `level_i` there is a bare integer index with `attrs == {}`. `tools/jam_burden_report.py` no longer compensates, and its docstring says so.

`level_interface` no longer exists; use `level_i` on both backends.

## Verified

- `ruff check .` clean. Fast gate (after the folded-in fixes): **2028 passed, 15 skipped, coverage 94.81%** (≥90). Slow gate: all 28 slow-test files green file-by-file, **0 failures** (per-file fresh processes — the dev box has ~3 GB available RAM and one long-lived pytest process gets `SIGABRT` inside XLA compilation; the earlier 80%-gate coverage measurement stands at 83.17%).
- New `jcm/cf_metadata_test.py` asserts the convention on a **written netCDF** from a real run (not an in-memory Dataset), per the issue: both axes surface-first, every full level strictly bracketed by its two interfaces, `-diff(pressure_half) > 0` everywhere, `a_half + b_half·p_s` reproducing `pressure_half` to 1e-5, and the attributes present.
- `jcm/observers_test.py` pins the observer `level` axis to the trajectory file's, and to `observations[:, ::-1]`.
- Manual T21L47 ECHAM run written and reopened: `pressure_full` and `pressure_half` now agree in every column.

## Folded-in follow-ups (originally filed as issues, brought into this PR for one consistent review)

- **#741** — `load_states_from_xarray` now detects the file's orientation from the `level` coordinate *values* (descending sigma = surface-first) and always returns the top-first physics frame; a bare-index axis passes through with a logged warning. All callers (SCM mode in `runners.py`, the T63L47 spinup loader) verified. One consequence filed as **#744**: `default_statistics.nc` was generated through the old inverted loader, so the GPU-gated climatology test may need its reference regenerated.
- **#739** — `PrescribedStateModel` output joins the file convention: the vertical table is threaded into the output struct and `to_xarray` ends in `cf_metadata.finalize_output`, so its `level` axis is surface-first with real sigma values. Round-trips through the now orientation-aware reader.
- **#740** — new `PhysicsTerm.output_attrs` declaration (aggregated by `ComposablePhysics`, applied in `to_xarray` with CSV < term < cf_metadata precedence). Populated for the radiation flux/heating set (exact CF standard names), cloud and convection diagnostics — only where units were already stated in code; vertical diffusion, surface tiles, JAM/MACv2-SP, chemistry, gravity waves and SPEEDY deliberately left for later rather than guessed.
- **#744** — `default_statistics.nc` regenerated on GPU through the orientation-aware loader. The old bands provably described an inverted-state run (global-mean normalized surface pressure 9.75e-6 — the model-top value; cloud fraction 0.988; humidity identically zero). `spinup_state.nc` untouched (byte-identical). One test change rode along: where a stored band's temporal std underflows to exactly 0 (the ~1e-24 humidity tail, constant pure-`a` pressure levels), the ±3σ band collapses to a point that cannot absorb cross-process XLA autotuning drift, so those levels — and only those — fall back to rtol=0.25/atol=1e-8; every physically meaningful level keeps strict ±3σ. Verified passing on GPU.
- **New `pressure_thickness` diagnostic** [Pa]: the per-layer Δp, emitted by `MoistAirColumnState` directly on the `level` axis. A Codex review caught the documented burden example silently evaluating to 0.0 (the diff-and-rename kept interface sigma labels, so xarray's alignment matched nothing) — the mass-weight is commonly needed and the model already computes it, so it is now simply in the output. Docs lead with it; `jam_burden_report` prefers it; the corrected `diff(pressure_half)` form stays documented as the fallback for SPEEDY/pre-diagnostic files.
- Also fixes the second Codex P2 (parametric `standard_name` without `formula_terms` on dynamics-only/observer files — CF requires the pair together; the standard name is now emitted with `formula_terms` or not at all).


jcm/cf_metadata.py
```diff
@@ -0,0 +1,297 @@
...
+    },
+    "lat": {
+        "standard_name": "latitude",
+        "units": "degrees_north",
+        "axis": "Y",
+        "long_name": "latitude",
+    },
+    # ``units`` for time is owned by xarray's datetime encoding — setting it
+    # here would collide with the encoding on write.
+    "time": {"standard_name": "time", "axis": "T", "long_name": "time"},
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in 37ff352.

You're right, and the claim was actively harmful: `Conventions = CF-1.11` plus `standard_name = "time"` on an axis with no decodable reference-time units is worse than carrying no metadata at all.

Fixed at both ends:

1. **Source** — `PysesCamSEDycore.to_xarray` now applies the same sim-day → `datetime64[ns]` conversion that `ModelPredictions._trajectory_dataset` uses on the dinosaur path, so every backend emits a real CF time axis and xarray's encoding supplies the units on write.
2. **Guard** — `cf_metadata._time_attrs` branches on dtype: a non-datetime axis gets `axis: T`, `units: d`, `long_name: elapsed simulation time`, and is explicitly *not* given `standard_name: time`. That protects any future producer that forgets the conversion, since a wrong CF claim is worse than none.

Both covered by new tests: `test_datetime_time_axis_gets_the_cf_time_attributes` and `test_numeric_time_axis_is_not_claimed_as_a_cf_time_coordinate`.

tools/jam_burden_report.py
```diff
@@ -60,26 +59,19 @@ def _wmean(da: xr.DataArray, weights) -> float:
...
-                          > float(ph.isel({iface_dim: -1}).mean()))
-    pf = ds["pressure_full"]
-    if "time" in pf.dims:
-        pf = pf.isel(time=0)
-    full_surface_first = (float(pf.isel(level=0).mean())
-                          > float(pf.isel(level=-1).mean()))
-    if half_surface_first != full_surface_first:
-        dp = np.flip(dp, axis=axis)
-    dims = tuple("level" if d == iface_dim else d for d in ph.dims)
+    dims = tuple("level" if d == "level_i" else d for d in ph.dims)
```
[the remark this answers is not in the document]
duncanwp: Declined — deliberate, not an oversight.

The orientation compensation was removed on purpose: it was the workaround for #710, and the maintainer's call is that the convention change is called out explicitly in the release notes rather than papered over at read time. Silently re-orienting old files is what let the two axes disagree unnoticed for as long as they did.

Your factual points are both correct, and neither changes that decision:

- Pre-#710 dinosaur files do pair each tracer level with the wrong Δp here. That is the accepted cost of not carrying a compatibility shim; those files need re-running.
- Pre-#710 pyses files raise on the `level_i` lookup instead of being handled. That is also accepted — this tool targets current output.

The docstring now states the scope plainly instead of implying the tool still copes with both orientations.

duncanwp:
@codex review

jcm/cf_metadata.py
```diff
@@ -0,0 +1,320 @@
...
+        if dim not in ds.coords:
+            continue
+        attrs: dict[str, Any] = dict(_SIGMA_COMMON_ATTRS)
+        attrs["long_name"] = (
+            f"nominal sigma (a/p0 + b) at {where}, surface-first "
+            "(index 0 is the surface)"
+        )
+        if has_ps and a_name in ds.coords and b_name in ds.coords:
+            attrs["formula_terms"] = (
+                f"ap: {a_name} b: {b_name} ps: surface_pressure"
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in cb2c3f9.

Correct — CF-1.11 requires `formula_terms` on any parametric vertical coordinate, so stamping `atmosphere_hybrid_sigma_pressure_coordinate` on a dynamics-only or observer file (no `surface_pressure`) advertised a coordinate a checker rejects. The parametric `standard_name` is now emitted together with `formula_terms` or not at all; the axis keeps `units`/`positive`/`axis`/`long_name`, which is a legal CF vertical coordinate on its own. `test_formula_terms_omitted_without_surface_pressure` now asserts both are absent together.

duncanwp:
@codex review

jcm/dycore/pyses/dycore.py
```diff
@@ -1063,4 +1065,13 @@ def add_field(name, arr):
 
         for key, value in (additional_coords or {}).items():
             ds.coords[key] = value
-        return ds
+
+        # ``add_field`` already emitted the data surface-first, so only the
+        # coordinates and CF metadata are left to attach.
+        return cf_metadata.finalize_output(
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in b16baf0e.

Right — the non-modal branch returned before the merge ever ran. The apply loop is now a module-level helper used by both branches, and the delegated dataset goes through it before returning. Term-declared names (radiation.*, cloud/convection) are disjoint from `cf_metadata._VARIABLE_ATTRS`' names, so applying them after the backend's `finalize_output` leaves the documented CSV < term < cf_metadata precedence intact — noted in a comment. Covered by a new delegation test in `jcm/dycore/protocol_test.py` using a fake non-modal dycore.

jcm/observers.py
```diff
@@ -456,7 +474,15 @@ def to_dataset(self, samples: dict, t0_days: float,
         ds = xr.Dataset(data_vars, coords=coords)
         ds.attrs["observer"] = self.name
         ds.attrs["vertical"] = self.vertical
-        return ds
+        # Data is already surface-first above, so only the coordinates and CF
+        # attributes are attached here. ``a_half`` is absent only if
+        # ``cache_grid`` was never called, in which case there is no vertical
+        # table to describe.
+        a_half = getattr(self, "_a_half", None)
+        return cf_metadata.finalize_output(
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in b16baf0e.

`latitude`/`longitude` (and the `altitude`/`pressure` target coordinates) are now in `_COORD_ATTRS` with `standard_name`, degree/SI units and `long_name` — deliberately without `axis`, which CF restricts to true coordinate variables; these are 2-D auxiliary coordinates. `test_observation_datasets` now asserts the attributes land in both observer datasets.

duncanwp:
@codex review

jcm/cf_metadata.py
```diff
@@ -0,0 +1,385 @@
...
+    claimed to be a CF time coordinate.
+
+    Backends should emit datetime64 (``PysesCamSEDycore.to_xarray`` and
+    ``ModelPredictions._trajectory_dataset`` both do); this is the guard for
+    anything that does not.
+    """
+    if np.issubdtype(np.asarray(time_coord.values).dtype, np.datetime64):
+        return dict(_COORD_ATTRS["time"])
+    return {"axis": "T", "units": "d",
+            "long_name": "elapsed simulation time"}
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in the follow-up commit.

Agreed — `axis = "T"` is itself the CF time-axis marker, so the fallback was making the same over-claim through a different attribute. The numeric branch now emits only `units: "d"` + `long_name: "elapsed simulation time"`; `standard_name`/`axis` are reserved for a datetime64 coordinate where xarray's encoding supplies the reference units. Test updated to assert both are absent on a numeric axis.

duncanwp:
@codex review

duncanwp:
@codex review

docs/source/getting_started.rst
```diff
@@ -509,6 +512,28 @@ The model output is a :py:class:`Predictions` object containing the model state
...
+Output files carry two vertical axes: ``level`` (``nlev`` layer mid-levels —
+temperature, tracers, ``pressure_full``) and ``level_i`` (``nlev+1``
+interfaces — ``pressure_half``, radiative fluxes). **Both run surface-first**,
+so index 0 is the level nearest the ground and ``level[k]`` sits between
+``level_i[k]`` and ``level_i[k+1]``. Mixing the two is therefore safe:
+
+.. code-block:: python
+
+   dp = -ds['pressure_half'].diff('level_i').rename(level_i='level')
+   burden = (ds['qc'] * dp / 9.81).sum('level')     # kg/m^2
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed — and verified by executing the snippet, which did indeed return exactly 0.0 (the inner join on two disjoint `level` label sets matches nothing).

The documented example now assigns the mid-level coordinate after the rename (`.assign_coords(level=ds['level'])`) with a comment explaining why, and the fixed snippet is verified to reproduce the hand-computed burden. Grep confirms no code shared the pattern: `jam_burden_report._layer_dp` constructs its Δp without coordinate labels, so multiplication propagates the tracer's coords and was never affected.

Good catch — a docs example that silently returns zero is precisely the failure class this PR exists to remove.

docs/source/getting_started.rst
```diff
@@ -499,8 +499,11 @@ The model output is a :py:class:`Predictions` object containing the model state
    # Print variables
    print(ds.data_vars)
 
-   # Plot surface temperature evolution
-   ds['temperature'].isel(level=7).mean(dim='lon').plot()
+   # Plot surface temperature evolution. Select by coordinate value, not by
+   # index: both vertical axes are surface-first (``level`` ~ 1 at the ground,
+   # ~1e-5 at the model top), but say so through ``sel`` rather than baking an
+   # index into the analysis.
+   ds['temperature'].sel(level=1.0, method='nearest').mean(dim='lon').plot()
```
duncanwp: No, I actually disagree with this. If you're picking a level and specifically want the surface level, this is confusing (especially as python is zero indexed, and you need to do a nearest neighbor lookup). Just do `isel(level=0)`.  Unless there's an `altitude=0` I feel integer level selection is cleanest.

docs/source/design/output_vertical_conventions.md
```diff
@@ -0,0 +1,128 @@
...
+The two axes used to be flipped independently, which meant they could — and
+did — disagree. `ModelPredictions._trajectory_dataset` reversed `level` inline
+and left `level_i` in the TOA-first frame it came out of the physics in, so
+`pressure_full` and `pressure_half` in the *same file* ran in opposite
+directions in every column ([#710]). Nothing raised: a heating rate computed
+from the saved fluxes and compared against the saved temperature, or a
+mass-weighted tracer burden, simply came out vertically reversed with a
+plausible magnitude.
+
+The failure mode is not "someone forgot to flip an array" — it is that the
```
duncanwp: Remove this paragraph. It's uncecasary

docs/source/design/output_vertical_conventions.md
```diff
@@ -0,0 +1,128 @@
...
+"down"` whichever way it is stored. The storage order is legible from the
+coordinate values themselves (they descend from ~1 towards 0) and is stated in
+`long_name` for anyone skimming `ncdump -h`.
+
+`positive` is deliberately **not** set on `pressure_full` / `pressure_half` /
+`height_*`: CF defines it for vertical coordinate variables, and those are data
+variables. Their orientation is that of the axis they sit on, and both axes now
+agree.
+
+## Both backends, one name
```
duncanwp: Drop this subseciion. It's already noted in release notes and doesn't need a logging here

jcm/data/test/echam_t63l47/generate_default_stats.py
```diff
@@ -93,7 +93,18 @@ def _block_until_ready(predictions):
 
 
 def _load_spinup_state():
-    """Load the saved spun-up nodal PhysicsState as the initial condition."""
+    """Load the saved spun-up nodal PhysicsState as the initial condition.
+
+    ``spinup_state.nc`` is written by ``ModelPredictions.to_xarray()`` and is
```
duncanwp: This is unnecessary and confusing. What convention does this *function* expect?

docs/source/getting_started.rst
duncanwp: Done in 2e60857 — example is now `isel(level=0)` with a one-line comment that index 0 is the surface. Agreed: with the convention guaranteed and stated in the file, the coordinate-value lookup was the confusing form.

docs/source/design/output_vertical_conventions.md
duncanwp: Removed in 2e60857.

docs/source/design/output_vertical_conventions.md
duncanwp: Dropped in 2e60857 — the rename note lives in the release notes only.

jcm/data/test/echam_t63l47/generate_default_stats.py
duncanwp: Simplified in 2e60857 — the docstring now states only the function's own contract: it returns the top-first physics frame, and the loader owns file orientation.

Issue #640, opened 2026-08-14
api: promote the science out of runners.py — runners becomes a thin Hydra adapter


duncanwp:
Part 2 of the documentation/API restructure (with #639, #641; same design smell as #638's preset-vs-factory drift).

## Problem

`jcm/runners.py` has grown a second API with science trapped behind Hydra `DictConfig` signatures. Functionality that is genuinely part of the modelling library is reachable only through the CLI layer:

- **Initial-condition science**: `inject_jw_profile` (Jablonowski–Williamson + humidity injection and ps rebalance), `inject_balanced_isothermal_profile`, `inject_era5_state`. A notebook user wanting a JW start has no supported path.
- **Stable time-step resolution** (physics/resolution-aware default Δt).
- **Sponge and nudging composition** (`maybe_add_sponge`, `maybe_add_nudging`, `_nudging_inv_tau`).
- **Yearly-forcing plumbing**: `_expand_years` + per-product coverage clamping (#633) — anyone assembling `ForcingData` from yearly files by hand in Python re-implements it.

## What this is NOT

The notebooks should **not** be ported to import from runners. Audit result: every notebook already uses the clean public API (`Model`, `get_coords`, `TerrainData.from_*`, `ForcingData.from_file`, `speedy_physics`/`echam_physics`) and none touch runners — coupling them to `DictConfig` composition would be a regression. The direction of travel is the opposite: promote the library-worthy pieces OUT of runners.

## Principle

**`runners.py` contains zero science.** It parses config and calls the library. Concretely:

- [ ] Extract the profile injectors to a `jcm.initialization` (or similar) module taking `(model | coords, ...)` — runners keeps one-line adapters.
- [ ] Move the stable-Δt resolution onto `Model` (or a helper beside it) and document it in the Python quick start.
- [ ] Give year-expansion/coverage-clamp a public home on the forcing side (`ForcingData.from_yearly_files(...)`?) so the CLI and Python paths share it.
- [ ] Sponge/nudging: expose as ordinary physics-term helpers composable via `ComposablePhysics` (mostly true already — finish and document).
- [ ] Sweep the remainder of runners for any other science; the end state is that a diff of runners never needs a scientific reviewer.

Update the notebooks only where a promoted API genuinely simplifies them (e.g. a JW-start demo), never to route through Hydra.

duncanwp:
Design smells surfaced by the #644 dedupe pass (Duncan asked these be recorded while keeping `tools/` lean):

[...]

PR #760, opened 2026-09-04
params: wire cld_frac_min, delete the dead Hydra-settable fields, split the 1M ccraut overload (#674, PR 4 of 4)

### LINKED ISSUE #674: Physics parameters: dead, misleading, and mistuned Hydra-settable fields
Found by a re-audit of the July 2026 ECHAM+RRTMGP review against `dev` @ `f6d1bcd`. Parent: #614.

## Why this is a correctness issue, not hygiene

`runners.py:221-226` builds every Parameters struct as `base.__class__(**{**base.__dict__, **dict(overrides)})`, so **any** field of these structs is settable from Hydra — and `config/physics/echam.yaml:11` advertises exactly that. Verified against the real constructor path:

```
override rhcrit=0.5:   ACCEPTED, stored=0.5
override cmfctop=0.9:  ACCEPTED, stored=0.9
override dt_conv=1800: ACCEPTED, stored=1800.0
unknown field -> TypeError: unexpected keyword argument 'not_a_field'
```

A **misspelled** field raises. A **real-but-dead** field is silently accepted and stored. That is the only silent failure mode in the whole override system, and it lands exactly on the dead set. A calibration sweep over any field below runs to completion, produces zero sensitivity, and reports "the model is insensitive to X". Six of them are advertised in the built Sphinx docs (`docs/source/echam_physics.rst`).

## Wire up (the value is hardcoded next to the dead field)

| Field | Evidence |
|---|---|
| `ConvectionParameters.rhcrit` | `tiedtke_nordeng.py:564` hardcodes `sigmoid((rh_env - 0.90)/config.smooth_rh)` — **0.90 is exactly `rhcrit`'s default**, and the sigmoid *width* beside it IS read from config |
| `CloudParameters.t_ice` (Sundqvist) | 238.15 hardcoded at `sundqvist.py:187`, `:211`, `:441`; `:180` even comments "Linear interpolation between t_ice and tmelt" one line above a literal |
| `RadiationParameters.cld_frac_min` | `mcica.py:44-47` `in_cloud_path(..., eps=1.0e-3)` — 1e-3 is exactly the default — and no caller passes `eps` (`rrtmgp.py:581,584`; `grey_two_stream/radiation_scheme.py:385,388`) |

## Delete (no hardcoded counterpart; nothing to wire)

- **`ConvectionParameters`** — `dt_conv` (+ its stale FIXME at `types.py:85` and its docs row), `epsilon` (a numerical guard, and the scheme already uses three different literals: 1e-12, 1e-10, 1e-7), `rlcrit` (whose own comment "Critical relative humidity" is wrong — 8e-4 is not an RH).
- **1M `MicrophysicsParameters` — 14 of 33 fields.** Never read: `cthomi`, `csecfrl`, `vt_ice`, `vt_snow_a/b`, `vt_rain_a/b`, `dt_sedi`. Formerly read only by code deleted in #569, now with **zero references of any kind**: `ccollec`, `ccollei`, `tau_melt`, `tau_freeze`, `cevaprain`, `cevapsnow`. Six of these are in the Sphinx docs and must be stripped there too. (`cn0s`, `crhosno`, `ceffmin`, `ceffmax` are now live — do not delete.)
- **`CloudParameters`** — `ceffmin`, `ceffmax` (Sundqvist computes no effective radius at all; these are orphaned duplicates of the live `MicrophysicsParameters` pair).
- **`CloudParams2M`** — `cap` only. The other 17 fields the review listed are now gone or genuinely wired via `cloud_utils.py`.
- **`AerosolParameters`** — `nplumes`, `nfeatures` (shape metadata read only by tests; derive from `plume_lat.shape[0]` / `ftr_weight.shape[0]`, or mark `pytree_node=False` so `jax.grad` stops tracing them). `background_aod` is now live — do not delete.
- **`RadiationParameters`** — `n_sw_bands`, `n_lw_bands` (**cannot** be wired: the code must use the module constants `N_SW_BANDS`/`N_LW_BANDS` because they need to be Python ints for static shapes, so a traced jnp copy is unusable by construction), `flux_epsilon`, `cld_tau_min`.
- **`ConvectionParameters.cmfctop`** — decided in the updraft entrainment/detrainment issue; delete unless the ECHAM overshoot layer is wired there.

## Two overloaded/mistuned fields

**`ccraut` means two different things in 1M** (`echam_1m.py:110-120`). `default(ccraut=15.0)` is the *Beheng rate prefactor*; in KK2000 mode the same field is a qc *threshold* (~1e-5). Since #556 put `ccraut` inside the value as `sigmoid((qc - ccraut)/smooth_ccraut)`, selecting `autoconversion_scheme="kk2000"` without also overriding `ccraut` gives `sigmoid((qc - 15.0)/5e-5) ≡ 0` for every physical qc — **autoconversion silently disabled.** Split into `ccraut_beheng` and `ccraut_kk_threshold`, or validate at construction.

**`ccsaut = 95` and `ccraut = 15` in 2M are ECHAM's un-overridden defaults** (`lohmann_2m_params.py:143, 146`). They are literally `mo_echam_cloud_params.f90:59, 68` — correctly ported — but `mo_activ.f90:391-408` re-tunes them to **900** and **10.6** for `nlev==47 .AND. nn==63 .AND. ncd_activ==2 .AND. cdnc_min_fixed==40`, which is exactly jcm's configuration. Cold-cloud snow autoconversion is therefore ~9.5× weaker than the HAM tuning (consistent with the ne30 cirrus-canopy symptom pattern) and warm-rain autoconversion ~40 % too fast.

**These two are recorded as tuning targets, not defects** — they are internally consistent, and both are already differentiable leaves, so they are legitimate calibration parameters rather than things to hard-pin. Decide deliberately whether jcm targets HAM's T63L47 tuning. Note `cdnc_min_fixed = 40` is what *selects* that tuning branch, so the three move together. Confirmed **not** CAM-derived: a CAM-faithful KK2000 port would carry `micro_mg_autocon_fact = 0.01` and `nd_exp = −1.1`, not jcm's original-KK `−1.79` and integrated-sink form.

## Fix

Prune or wire per the tables, strip the dead rows from `docs/source/echam_physics.rst`, and — independently of the pruning — add a validation pass in `_build_term` that warns when an override names a field the scheme never reads, so the class cannot silently return.

## Refs

Maintainability review C.2, plus review findings B.1.12 (`ccraut`), 2.14 (`ccollec`/`ccollei`), and the `mo_activ` tuning check.

duncanwp:
Closes #674. **STACKED on #759** (base is `feat/640-jam-macsp-separation`) — review after it; the diff shown here is only the 4 parameter-pass commits. Refs #640 (final PR of the campaign).

## What

- **Wire**: `RadiationParameters.cld_frac_min` now reaches `mcica.in_cloud_path(eps=...)` at all four call sites (RRTMGP ×2, grey ×2). Behavior unchanged at the 1e-3 default; an override now actually moves the masking (tested through the grey path). The NN emulator's call deliberately stays at the default — the eps is part of its trained feature/label contract.
- **Delete ~27 confirmed-dead fields** (each verified unread on this base): `ConvectionParameters.{dt_conv, epsilon, rlcrit, cmfctop, rhcrit}` (`rhcrit`'s 0.90 target became `cu_midlev_rh` in #690, so it moved from the issue's wire-up list to deletion); 14 dead 1M `MicrophysicsParameters` fields (`cthomi, csecfrl, vt_*, dt_sedi, ccollec, ccollei, tau_melt, tau_freeze, cevaprain, cevapsnow` — the same-named *2M* fields are live and untouched); Sundqvist `ceffmin/ceffmax` (orphaned duplicates; the 1M pair is live); `AerosolParameters.{nplumes, nfeatures}` (tests now derive from array shapes); `RadiationParameters.{n_sw_bands, n_lw_bands, flux_epsilon, cld_tau_min}` (the band counts *cannot* be wired — static shapes require the module constants). With #747's audible-override warning in the base, a config naming any of these now fails loudly instead of silently returning a flat sensitivity.
- **Split the 1M `ccraut` overload**: `ccraut` keeps its name and 15.0 default as the Beheng rate prefactor; new differentiable `ccraut_kk_threshold` (1e-5 kg/kg, per the struct's own prior guidance) feeds the KK2000 sigmoid. Selecting `autoconversion_scheme="kk2000"` at defaults previously computed `sigmoid((qc−15)/5e-5) ≡ 0` — autoconversion silently off; now it produces physical rates (regression-tested at qc = 1e-4..1e-3), and Beheng is bit-identical at defaults.
- **Docs**: every deleted field's row stripped from `docs/source/echam_physics.rst`; a `ccraut_kk_threshold` row added.
- **Recorded, deliberately not changed**: the 2M `ccsaut=95`/`ccraut=15` defaults vs ECHAM-HAM's T63L47 retune (900/10.6, `mo_activ.f90:369-410`, selected under `ncd_activ=2` + `cdnc_min_fixed=40` — verified against the Fortran). They are internally consistent, differentiable calibration targets; the comment at the definitions now names the retune and its selection condition so the decision is discoverable where the values live.

## Verification

`ruff` clean; targeted parameter/cloud/radiation suites **418 passed** on this tip over the rebased base; the full fast+slow gates ran green on the base branch (#759's numbers). A hidden test dependency the issue's audit missed was fixed rather than left to rot: `aerosol_radiation_test` read `n_sw_bands` inside a `try/except: return`, which the deletion would have turned into a silent no-op — it now asserts against the module constants.

### DIFF
--- a/jcm/physics/aerosol/macv2_sp_params.py
+++ b/jcm/physics/aerosol/macv2_sp_params.py
@@ -21,10 +21,11 @@ class AerosolParameters:
     background aerosol.
     """
     
-    # Number of plumes and features
-    nplumes: int
-    nfeatures: int
-    
+    # Plume count and feature count are not stored as fields: they are the
+    # leading axis lengths of the arrays below (``plume_lat.shape[0]`` and
+    # ``ftr_weight.shape[0]``), so a stored copy could only ever drift out
+    # of sync with the real array shapes (#674).
+
     # Plume center locations [degrees]
     plume_lat: jnp.ndarray        # (nplumes,) latitude of plume centers
     plume_lon: jnp.ndarray        # (nplumes,) longitude of plume centers
@@ -87,9 +88,6 @@ class AerosolParameters:
         Longitudes use the file's 0-360 convention, matching the
         dinosaur-derived column longitudes cached by the term.
         """
-        nplumes = 9
-        nfeatures = 2
-
         # Plume centers [degrees N / degrees E, 0-360].
         plume_lat = jnp.array(
             [49.4, 40.1, 30.0, 23.3, 3.5, -10.3, -1.0, -3.5, -20.0])
@@ -154,8 +152,6 @@ class AerosolParameters:
         ]).T
 
         return cls(
-            nplumes=nplumes,
-            nfeatures=nfeatures,
             plume_lat=plume_lat,
             plume_lon=plume_lon,
             beta_a=beta_a,
@@ -193,8 +189,6 @@ class AerosolParameters:
         as_arr = lambda name: jnp.asarray(ds[name].values)
         as_arr_T = lambda name: jnp.asarray(ds[name].values.T)
         return cls(
-            nplumes=int(ds.sizes["plume_number"]),
-            nfeatures=int(ds.sizes["plume_feature"]),
             plume_lat=as_arr("plume_lat"),
             plume_lon=as_arr("plume_lon"),
             beta_a=as_arr("beta_a"),
--- a/jcm/physics/aerosol/macv2_sp_params.py
+++ b/jcm/physics/aerosol/macv2_sp_params.py
@@ -21,11 +21,6 @@ class AerosolParameters:
     background aerosol.
     """
     
-    # Plume count and feature count are not stored as fields: they are the
-    # leading axis lengths of the arrays below (``plume_lat.shape[0]`` and
-    # ``ftr_weight.shape[0]``), so a stored copy could only ever drift out
-    # of sync with the real array shapes (#674).
-
     # Plume center locations [degrees]
     plume_lat: jnp.ndarray        # (nplumes,) latitude of plume centers
     plume_lon: jnp.ndarray        # (nplumes,) longitude of plume centers
--- a/jcm/physics/clouds/echam_1m.py
+++ b/jcm/physics/clouds/echam_1m.py
@@ -30,11 +30,21 @@ from jcm.physics.clouds.cloud_utils import eff_liquid_droplet_radius
 class MicrophysicsParameters:
     """Configuration parameters for cloud microphysics"""
     
-    # Autoconversion parameters
-    ccraut: float        # Critical cloud water for autoconversion (kg/kg)
+    # Autoconversion parameters. ``ccraut`` and ``ccraut_kk_threshold`` are
+    # separate fields because they are physically different quantities: one
+    # field serving as "rate prefactor (Beheng) OR qc threshold (KK2000)"
+    # meant selecting KK2000 without overriding it fed the Beheng 15.0 into
+    # the threshold sigmoid — sigmoid((qc - 15)/5e-5) ≡ 0 for any physical
+    # qc, silently disabling autoconversion (#674).
+    ccraut: float        # Beheng (1994) autoconversion rate prefactor (-);
+                         # ECHAM mo_echam_cloud_params ``ccraut`` (15.0).
+                         # Read only in Beheng mode.
+    ccraut_kk_threshold: float  # KK2000 in-cloud qc threshold (kg/kg) above
+                         # which autoconversion fires. Read only in KK2000
+                         # mode.
     smooth_ccraut: float # Sigmoid width of the KK2000 qc threshold (kg/kg);
-                         # only read in KK2000 mode — Beheng uses ccraut as
-                         # a rate coefficient (already smooth)
+                         # only read in KK2000 mode — Beheng's rate form is
+                         # already smooth
     ccracl: float        # Accretion coefficient (cloud to rain)
     cauloc: float        # ECHAM ``zrac2`` local-rain accretion enhancement.
                          # 0.0 is the ECHAM6.3 default (zrac2 disabled); raise to
@@ -91,20 +101,19 @@ class MicrophysicsParameters:
                          # write-back (mo_cloud.f90:1280, #687)
 
     # Autoconversion scheme selector (int flag — JAX won't trace strings).
-    # 0 = Beheng (1994) implicit form (default; robust at large dt).
+    # 0 = Beheng (1994) implicit form (default; robust at large dt),
+    #     controlled by ``ccraut``.
     # 1 = Khairoutdinov & Kogan (2000) explicit form (good fit for 2M
-    #     microphysics with prognostic Nc).
-    # ``ccraut`` is interpreted differently by each scheme: in Beheng
-    # it's the rate prefactor (default 15.0); in KK2000 it's the qc
-    # threshold above which autoconversion fires (a small g/kg-scale
-    # value is appropriate, e.g. 1e-5).
+    #     microphysics with prognostic Nc), controlled by
+    #     ``ccraut_kk_threshold`` / ``smooth_ccraut``.
     autoconversion_scheme: int
 
     SCHEME_BEHENG = 0
     SCHEME_KK2000 = 1
 
     @classmethod
-    def default(cls, ccraut=15.0, smooth_ccraut=5e-5,
+    def default(cls, ccraut=15.0, ccraut_kk_threshold=1.0e-5,
+                smooth_ccraut=5e-5,
                 ccracl=6.0, cauloc=0.0, clmin=0.0, clmax=0.5,
                  ceffmin=10.0, ceffmax=150.0, cn0s=3.0e6,
                  crhosno=100.0, ccsaut=95.0, ccsacl=0.1,
@@ -128,6 +137,7 @@ class MicrophysicsParameters:
 
         return cls(
             ccraut=jnp.array(ccraut),
+            ccraut_kk_threshold=jnp.array(ccraut_kk_threshold),
             smooth_ccraut=jnp.array(smooth_ccraut),
             ccracl=jnp.array(ccracl),
             cauloc=jnp.array(cauloc),
@@ -274,7 +284,8 @@ def autoconversion_kk2000(
 
         P_aut = 1350 * qc^2.47 * Nc_cm3^(-1.79)   [kg/kg/s, qc in kg/kg]
 
-    Activates above the ``ccraut`` threshold. KK2000 was the original
+    Activates above the ``ccraut_kk_threshold`` in-cloud qc threshold.
+    KK2000 was the original
     1M default and remains a good fit for 2M microphysics where the
     droplet number ``Nc`` is a prognostic variable. In the 1M context
     with prescribed ``Nc`` and large dt, the explicit form can produce
@@ -290,7 +301,8 @@ def autoconversion_kk2000(
         droplet_number: Cloud droplet number concentration (1/m³)
         dt: Time step (s) — unused (explicit rate); kept in signature
             for parity with ``autoconversion_beheng``.
-        config: Microphysics configuration (uses ccraut, epsilon)
+        config: Microphysics configuration (uses ccraut_kk_threshold,
+            smooth_ccraut, epsilon)
 
     Returns:
         Grid-mean autoconversion rate (kg/kg/s)
@@ -309,17 +321,20 @@ def autoconversion_kk2000(
     # previous code fed qc in g/m³ (×ρ×1000 ≈ ×1200) into the 2.47 power
     # and then applied a spurious g/m³→kg/kg back-conversion — a net
     # ~2.6e4× overestimate (review finding 1.5; non-default branch).
-    # Smooth threshold (maintainability review B.2.5): with the hard
-    # ``where(qc > ccraut, rate, 0)`` the threshold appears only in the
-    # inequality, so d(rate)/d(ccraut) is identically zero — ccraut was
-    # calibratable only in Beheng mode. The sigmoid ramp puts it in the
+    # Smooth threshold (maintainability review B.2.5): with a hard
+    # ``where(qc > threshold, rate, 0)`` the threshold appears only in the
+    # inequality, so d(rate)/d(threshold) is identically zero — the
+    # threshold was not calibratable. The sigmoid ramp puts it in the
     # value; width -> 0 recovers the hard gate. The qc power base is
     # double-where-guarded so the ramp's sub-threshold tail cannot
     # differentiate a negative/zero base (0**x cotangent class).
+    # ``ccraut_kk_threshold`` (~1e-5 kg/kg) is the KK2000-specific qc
+    # threshold — NOT the Beheng ``ccraut`` prefactor, which would push
+    # the sigmoid to 0 for any physical qc (#674).
     has_qc = qc_in_cloud > 0.0
     qc_safe = jnp.where(has_qc, qc_in_cloud, 1.0)
     ramp = jax.nn.sigmoid(
-        (qc_in_cloud - config.ccraut) / config.smooth_ccraut
+        (qc_in_cloud - config.ccraut_kk_threshold) / config.smooth_ccraut
     )
     autoconv_rate = jnp.where(
         has_qc,
--- a/jcm/physics/clouds/echam_1m.py
+++ b/jcm/physics/clouds/echam_1m.py
@@ -17,6 +17,8 @@ Based on the ECHAM6/ICON ``mo_cloud.f90`` single-moment branch
 (Lohmann and Roeckner, 1996).
 """
 
+import math
+
 import jax
 import jax.numpy as jnp
 from typing import NamedTuple, Tuple, Optional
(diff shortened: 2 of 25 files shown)


jcm/physics/radiation/rrtmgp.py
```diff
@@ -608,10 +608,12 @@ def radiation_scheme_rrtmgp(
     # the high end of realistic in-cloud water, so genuine clouds are untouched
     # and only the pathological inflation is clipped.
     cloud_water_in_cloud = jnp.minimum(
-        in_cloud_path(cloud_water, cloud_fraction), _MAX_IN_CLOUD_CONDENSATE
+        in_cloud_path(cloud_water, cloud_fraction, eps=parameters.cld_frac_min),
+        _MAX_IN_CLOUD_CONDENSATE,
     )
     cloud_ice_in_cloud = jnp.minimum(
-        in_cloud_path(cloud_ice, cloud_fraction), _MAX_IN_CLOUD_CONDENSATE
+        in_cloud_path(cloud_ice, cloud_fraction, eps=parameters.cld_frac_min),
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in ef98bcc2. Added `effective_cloud_fraction(cf, eps) = where(cf > 2*eps, cf, 0)` and fed it to the McICA sampler + cover in BOTH the RRTMGP and grey paths, so a cell whose in-cloud condensate `in_cloud_path` already zeroed (cf <= 2*cld_frac_min) is treated as clear everywhere. Reference verdict: FAITHFUL — ECHAM mo_psrad_interface.f90:232's WHERE zeros ziwgkg_vr/zlwgkg_vr AND clears the icldlyr cloud flag on the identical `cld_frc_vr > 2*EPSILON` test, tying condensate and cloud flag together; ECHAM's EPSILON is machine epsilon so it only bit empty cells, but here eps is physical (1e-3) so the tie is made explicit. Marginal-layer quantification (column test): a cf=1.5e-3-only column went from max-random cover 1.5e-3 to 0; the 1M/Sundqvist diagnostic snaps cf<0.01 to 0 so only 2M can populate (0, 2e-3]. NN-emulator deliberately excluded (learned features).

jcm/physics/clouds/echam_1m.py
```diff
@@ -348,17 +321,20 @@ def autoconversion_kk2000(
...
     # differentiate a negative/zero base (0**x cotangent class).
+    # ``ccraut_kk_threshold`` (~1e-5 kg/kg) is the KK2000-specific qc
+    # threshold — NOT the Beheng ``ccraut`` prefactor, which would push
+    # the sigmoid to 0 for any physical qc (#674).
     has_qc = qc_in_cloud > 0.0
     qc_safe = jnp.where(has_qc, qc_in_cloud, 1.0)
     ramp = jax.nn.sigmoid(
-        (qc_in_cloud - config.ccraut) / config.smooth_ccraut
+        (qc_in_cloud - config.ccraut_kk_threshold) / config.smooth_ccraut
     )
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in 5d733a0f. Added a construction-time guard in `MicrophysicsParameters.default` (the sole construction hook; host-side floats, unlike a __post_init__ that would run on every traced replace): when autoconversion_scheme resolves to kk2000, ccraut differs from the Beheng default (15.0), and ccraut_kk_threshold is still at its 1e-5 default, it raises a migration error naming ccraut_kk_threshold. Raising is safe — ccraut (Beheng prefactor) is UNUSED in the kk2000 branch, so a Beheng override alongside kk2000 is meaningless. Both defaults are now module constants shared by the signature and the guard. Tests: legacy ccraut-under-kk2000 raises; new-field override constructs; ccraut-under-Beheng untouched.

duncanwp:
@codex review

jcm/physics/clouds/echam_1m.py
```diff
@@ -151,8 +145,33 @@ def default(cls, ccraut=15.0, smooth_ccraut=5e-5,
...
+        # because ``ccraut`` remains a live Beheng field, but the KK2000 branch
+        # now reads ``ccraut_kk_threshold`` — so the override would be silently
+        # ignored and the threshold would stay at its 1e-5 default. Raising is
+        # safe: ``ccraut`` (the Beheng prefactor) is UNUSED in the KK2000
+        # branch, so a deliberate Beheng-prefactor override alongside kk2000
+        # is meaningless. A Beheng-mode ``ccraut`` override is untouched.
+        if (autoconversion_scheme == cls.SCHEME_KK2000
+                and not math.isclose(ccraut, _BEHENG_CCRAUT_DEFAULT)
+                and math.isclose(ccraut_kk_threshold,
+                                 _KK2000_QC_THRESHOLD_DEFAULT)):
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in f01d90c1. Moved the cross-field check into a plain-Python `MicrophysicsParameters.validate()` shared by both construction doors: `default()` calls it, and `_build_term` invokes the opt-in `getattr(params, "validate", None)` hook on the post-override object so any Parameters class can guard both doors. Config-time / concrete-values-only (never under a jit trace); f32-tolerant `rel_tol`. Tests cover the Hydra-override door raising, the `default()` door still raising, and both doors accepting the migrated config.

duncanwp:
@codex review

jcm/physics/clouds/echam_1m.py
```diff
@@ -165,29 +160,65 @@ def default(cls, ccraut=15.0, smooth_ccraut=5e-5,
...
+        # ignored and the threshold would stay at its 1e-5 default. Raising is
+        # safe: ``ccraut`` (the Beheng prefactor) is UNUSED in the KK2000
+        # branch, so a deliberate Beheng-prefactor override alongside kk2000
+        # is meaningless. A Beheng-mode ``ccraut`` override is untouched.
+        # ``rel_tol`` accommodates the f32 round-trip: the fields are stored as
+        # ``jnp.array`` (float32), so a Python 1e-5 default reads back as
+        # 9.9999997e-6 — well outside ``math.isclose``'s 1e-9 default. 1e-6 is
+        # far tighter than any physically meaningful override yet forgiving of
+        # f32 precision, so "left at the default" is recognised.
+        if (int(self.autoconversion_scheme) == self.SCHEME_KK2000
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in ff534352. Root cause: the "beheng"/"kk2000" aliases were only mapped to ints inside default(), so _build_term's direct __class__(**...) reconstruction of a Hydra override left the field a raw string — validate()'s int("beheng") then crashed (and the lax.cond dispatch compared string-vs-int, silently wrong).

Fix makes the STORED field canonical rather than adding tolerant readers: a _normalize_scheme() classmethod (maps aliases, passes ints/traced leaves through) runs in __post_init__, so BOTH construction doors yield the canonical int; default() no longer normalizes inline; validate() and the dispatch are unchanged.

FAMILY: VDiffParameters.surface_layer_scheme ("businger_dyer"/"echam_louis") is the same class — no validate() so no crash, but a raw string override silently mis-selected the turbulence_coefficients lax.cond branch. Fixed identically. No other alias-bearing Parameters fields found.

Tests: both alias sets construct+validate through _build_term; int forms unchanged; the legacy-ccraut migration guard fires under both the string and int kk2000 forms.

duncanwp:
@codex review

jcm/physics/aerosol/macv2_sp_params.py
```diff
@@ -21,10 +21,11 @@ class AerosolParameters:
     background aerosol.
     """
     
-    # Number of plumes and features
-    nplumes: int
-    nfeatures: int
-    
+    # Plume count and feature count are not stored as fields: they are the
```
duncanwp: Delete this comment

duncanwp: Done in 986f0765.

jcm/physics/clouds/lohmann_2m_params.py
```diff
@@ -154,6 +154,14 @@ def default(
         cn0s: float = 3e6,
         crhoi: float = 500.0,
         crhosno: float = 100.0,
+        # Recorded tuning target (#674, #641): these defaults are the generic
```
duncanwp: Delete this comment and update the parameters. JAM with ARG is the primary route for the 2m scheme. Update the macv2sp+2m config to specify these particular params instead

jcm/physics/convection/tiedtke_nordeng/types.py
```diff
@@ -15,10 +15,7 @@
 @tree_math.struct
 class ConvectionParameters:
     """Configuration parameters for Tiedtke-Nordeng convection scheme"""
-    
-    # Time stepping
-    dt_conv: float           # Convection timestep (s)
```
duncanwp, line 20: Glad to see this go. I seem to remember some complicated wiring in the Model that checks this is close to the model (dycore) timestep? Would be good to be able to remove that too

duncanwp: Investigated — nothing to remove, and here's why. `dt_conv` never had any consistency wiring: `git log -S dt_conv -- jcm/model.py jcm/physics_interface.py` is empty, and the field had no readers anywhere in `jcm/` (it was a pure `ConvectionParameters` field with a stale `# FIXME: validate dt_conv`). So the deletion is self-contained.

The "checks this is close to the dycore timestep" machinery you're remembering is the **generic** `Physics.stable_time_step_minutes` API, and it's unrelated to `dt_conv`:
- `Model._resolve_time_step_minutes` (jcm/model.py:640-685) — reconciles the model step against the dycore's `dt_seconds` (raises on an explicit-`time_step` vs explicit-`dycore` mismatch, else adopts the dycore step), and on the coords path consults the physics' stable-step limit.
- `ComposablePhysics.stable_time_step_minutes` (jcm/physics/composable_physics.py:179) takes the min over terms; `PhysicsTerm` base returns `None` (jcm/physics/physics_term.py:210).
- The **only** term that actually overrides it is `SpeedySurfaceFlux.stable_time_step_minutes` (jcm/physics/speedy/speedy_terms.py:418) — the forward-Euler stability limit of SPEEDY's explicit surface drag in the thin bottom sigma layer. That's live and load-bearing, so it stays.

Tiedtke convection never implemented `stable_time_step_minutes` (it returns the `None` base default), so removing `dt_conv` touches none of this. No change needed here.

jcm/physics/clouds/lohmann_2m_params.py
duncanwp: Done in 59f7d6e3.

**Fortran values** (verified in `/data/.../echam6.3.0-ham2.3-moz1.0.r7492/src/mo_activ.f90`, `activ_initialize`): under `IF (lcdnc_progn)` → `nn==63` (T63) → `CASE(2)` (AR&G activation) → `cdnc_min_fixed CASE(40)`:
```
ccsaut = 900._dp
ccraut = 10.6_dp
```
Identical for `nlev==31` and `nlev==47`. So the retune is ccsaut 95→900, ccraut 15→10.6.

**Defaults flipped** in `lohmann_2m_params.py::CloudParams2M.default` to 900 / 10.6, and the change-narrative comment replaced with a present-state one citing the Fortran and its selection condition (lcdnc_progn + ncd_activ==2 + cdnc_min_fixed==40, the JAM+ARG route per #759). JAM and every `echam_physics(cloud_scheme="2m")` build get `CloudParams2M.default()`, so they now carry 900 / 10.6 — **default JAM/2M runs change answers, by design**; release-validation will pick it up on its next matrix run.

**Base values pinned** (ccsaut: 95.0, ccraut: 15.0) with an explanatory comment in the three MACv2-SP (SPA-activation) presets — `echam-rrtmgp-2m.yaml`, `echam-rrtmgp-2m-cosp.yaml`, `echam-emulated-2m.yaml` — since SPA is not the AR&G branch the retune is conditioned on. Verified by instantiating each preset (95/15) and the factory route (900/10.6).

**Sweep:** no source docs pinned the 2M values (only the code definition site, now updated) — the `echam_physics.rst` ccraut rows are the separate 1M `MicrophysicsParameters.ccraut` (Beheng/kk2000), untouched. No test asserted the 2M default value. One test (`TestPrecipFluxProfiles2M::test_bottom_row_equals_surface_fluxes`) used `.default()` and asserted a hard rain/snow `>= 0`; under the new arithmetic a physically-zero level lands at ~-9e-16 (f32 roundoff, ~12 orders below the ~1e-3 signal), so I relaxed it to the same 1e-12 tolerance its equality checks already use. Full fast suite green (2363 passed).

duncanwp: Slashed to two lines in 09a5cbed.

Issue #356, opened 2026-01-04
Simple NN Bias correction paramaterization


duncanwp:
Add a simple NN-based (state) bias correction parameterization which we can train online to make the SPEEDY physics look more like e.g. ERA5. This could start off being a simple one-column scheme, but might also be easy to include non-local terms c.f. Troy's work.

PR #764, opened 2026-09-07
feat(bias-correction): online-trained NN correction term for SPEEDY (#356)

JOhnsonKC201:
## Describe your changes

A small neural network that runs as an ordinary `PhysicsTerm` on top of SPEEDY and removes part of the model's systematic error from inside the running simulation, rather than post-processing the output. It reads the model's own column state and needs no reanalysis at run time. This is the one-column scheme from #356; the non-local version is not in this PR.

**What the network is.** A per-column MLP with 32 inputs at T31L8 (the temperature, humidity and wind profiles) plus one context input (daily-mean top-of-atmosphere insolation), three hidden layers of 256 with tanh, 148,512 parameters shared across every column. It corrects temperature and specific humidity.

**How it is trained.** Four stages, each fixing a failure of the one before:

1. A supervised fit to the nudging tendency of a run nudged toward ERA5. Converges quickly but optimises a proxy: a term that scores well here is unstable in a free run.
2. Backpropagation through 6 h to 48 h rollouts of the full model, which is what makes a term stable.
3. A fine-tune against a 30-day free-run climatology, because a short rollout is not the same objective as a faithful long-run climate.
4. A post-hoc taper that fades the near-surface temperature correction to zero over sigma 0.7 to 1.0. Without it the correction improves the mid-troposphere and makes the surface worse.

**Result.** Trained on 1995-2015 and scored on 2016-2022, seven years absent from training, free-running with no nudging and no reanalysis input. Cos-latitude weighted RMS against ERA5, with plain SPEEDY over the same period in the first column:

| Metric | plain SPEEDY | with the term, seed 0 | same recipe, seed 7 |
|---|---|---|---|
| Near-surface T, annual | 3.64 K | 3.29 K | 3.23 K |
| Near-surface T, DJF | 3.60 K | 3.32 K | 3.38 K |
| Near-surface T, JJA | 4.17 K | 3.87 K | 3.85 K |
| T at 500 hPa | 4.73 K | 2.53 K | 3.70 K |
| Specific humidity | 1.46 g/kg | 0.79 g/kg | 1.00 g/kg |

**Three things a reviewer should know before reading the table as a headline.**

1. These are single training runs. The shipped recipe reproduced 5 of 5 at a second seed, but the same architecture without the context input, reseeded three times, gave 5, 3 and 5 metrics improved, with spreads up to 0.89 K at the surface and 2.79 K at 500 hPa. The improvement over plain SPEEDY is far larger than that spread and is not in question; a ranking of one corrected term against another on a 7-year window is not supported, and the docs page says so next to the numbers.
2. Every score was produced on the SPEEDY physics as of 748413a (July). The surface flux and vertical diffusion schemes changed on `dev` after that. On this branch, rebased onto current `dev`, the term loads, composes and runs (two model days on CPU, finite, checked here), but I have not re-scored it against the current plain SPEEDY. I will run plain and the term through the evaluator on the current physics on the lab GPU and post that table in this PR.
3. What fixed winter was capacity, not the context feature. At 8,352 parameters no configuration beat plain SPEEDY in DJF, whether it was given a land mask, a seasonal input, or twenty-one years of data instead of one. At 148k parameters every configuration did, with or without a context input. The insolation term ships because it is the only one that scored 5 of 5 in two seeds and on both the in-sample and the held-out years. I had written up the opposite conclusion in July (winter as a structural limit of a per-column correction); the docs keep the retraction.

**What did not work.** The docs page has a table of twelve things tried and rejected, each with the number that killed it: a latitude taper, training with the taper active, the land mask and the seasonal input at 8k, a loss weighted toward 500 hPa, 21 years at 8k, retraining at T63, sea ice at 148k, ensembling trained terms, selecting the network on offline error, and seasonal climatology windows (a first run improved DJF and the paired three-seed repeat flipped sign). It also lists what has not been tested.

### Changes outside the package

- `jcm/nudging.py` gains humidity relaxation. `NudgingTarget.specific_humidity` and `NudgingConfig.inv_tau_humidity` both default to `None`, meaning not relaxed, so existing targets and configs keep working with their old argument counts and `nudging_tendency` returns a zero humidity tendency unless both are supplied. `NudgingConfig.temp_humidity` builds the profile the training uses. The target's humidity is in g/kg (the `PhysicsState` convention), and the docstring says so because raw ERA5 is kg/kg and the wrong unit nudges toward one thousandth of the intended value without any error.
- `jcm/data/bc/interpolate_ozone.py`: `_vertical_interp_log_p` is now `vertical_interp_log_p`. The ERA5 loader reuses it for temperature, humidity and winds, and importing an underscore name across packages couples to a private detail.
- `.gitignore`: trained terms under `jcm/data/bias_correction/` are ignored except the shipped one, so an experiment artifact cannot enter the repository by accident; the drivers' output directories are ignored too.

### What ships and what does not

Ships: the package (`jcm/physics/bias_correction/`, 9 modules with co-located tests), one trained term (`online_term_t31_big_insol_vt.npz`, 584 KB, self-contained), `TERMS_README.md` with the comparison table and the measured noise floors, the drivers under `tools/bias_correction/` with a README so the term is reproducible from this checkout (following the `tools/release_validation/` precedent), and `docs/source/bias_correction.rst`.

Does not ship: the other 37 trained terms, the two early notebooks, and the exploratory sweep scripts. They stay in my research repository. `TERMS_README.md` lists the comparison terms with their scores and the chain regenerates any of them.

### Usage

```python
from jcm.model import Model
from jcm.physics.bias_correction import load_bias_correction
from jcm.physics.speedy.speedy_coords import get_speedy_coords
from jcm.physics.speedy.speedy_terms import speedy_physics

coords = get_speedy_coords(layers=8, spectral_truncation=31)
term = load_bias_correction("jcm/data/bias_correction/online_term_t31_big_insol_vt.npz")
model = Model(coords=coords, terrain=terrain, physics=speedy_physics() + term)
```

`requires` and `provides` are empty, so it composes anywhere in the stack; append it last. A freshly built term (`make_bias_correction(coords)`) is a zero-initialised no-op, so composing it changes nothing until it is trained.

### Testing

- Fast suite (`pytest -m "not slow"`) in a clean Linux environment built from `requirements.txt` plus the forked dinosaur, on this branch as rebased onto `dev` at eff6c443: 2514 passed, 45 skipped, 1 failed. The failure is `provenance_test::test_jcm_probed_with_git_state`, which asserts a git sha and ran in a scratch tree unpacked from an archive with no `.git`; it passes in a real checkout. On Windows the same suite gives 2432 passed with only the known Windows-only failures (rrtmgp's hard-coded `/tmp/netcdf` path, file locks, path separators).
- The layout audit (`jcm/physics/layout_audit_test.py`) now drives `NNBiasCorrection` with non-zero weights on both the nodal grid and a column-vectorised block and requires the tendencies to agree per column, so the broadcasting-native claim is checked on every CI run rather than asserted.
- Slow suite (`pytest -m slow`) in the same Linux environment: 90 passed, 27 skipped (the GPU-gated ones, as on CI), 0 failed, plus the new shipped-term test below.
- PR coverage gate (`.coveragerc-pr`, 80%): 80.35%. Two things got it there and both are in this PR as their own commit or test. The seven training, data and evaluation modules are added to the PR-only omit list, following the file's existing rule for `runners.py` and `utils.py`: the fast suite drives them at 96 to 100% and the slow suite never enters them, so without the omit the gate read 78% for a reason unrelated to what the slow suite tests. The term itself stays in the denominator, and a new slow test loads the shipped `.npz`, checks its configuration (insolation context, taper edges, 148,512 parameters), composes it onto SPEEDY at T31 with the packaged boundary conditions and runs half a model day, which took the term's slow coverage from 53% to 72% and also guards the artifact file.
- `ruff check .` is clean with the pinned 0.15.17.
- The docs build; the page renders with its tables, and the only Sphinx warnings are the pre-existing ones from the pyses dycore docstrings.
- Gradients of the tendency and the rollout loss are checked with `check_vjp` and `check_jvp` against finite differences in float64, not only asserted finite: online training backpropagates through hundreds of these calls, and a subtly wrong gradient still descends, just to the wrong place.
- Coverage on that Linux run: 92% for the whole package against the 90% push gate. The new modules are at 96 to 100% except `era5_cache.py` at 63%, whose uncovered lines are the year-long nudged model runs that build the cache; the tests around it cover completion marking, resumability, metadata validation and the pairing, not the model run itself. `jcm/nudging.py` is at 97%.
- The shipped term was loaded, composed onto `speedy_physics()` and run for two model days at T31 on CPU on this branch: finite, and it changes the trajectory.

Anyone running the suite needs the forked dinosaur that `dev` already requires; four tests construct a `Model`.

#356


duncanwp:
Thanks KC! I'm going to close this since it doesn't belong on the main JCM repo. If you need specific changes / additions to the main API then make that as a separate PI here. 

Please recreate it against https://github.com/climate-analytics-lab/jcm-experiments. If you don't have access just let me know.

JOhnsonKC201:
Thanks, that makes sense. I don't have access to jcm-experiments (it 404s for me), could you add me?

Plan once I can see the layout: the two small changes the main API needs (optional humidity relaxation in `jcm/nudging.py`, and making `vertical_interp_log_p` public in `interpolate_ozone.py`) go up as one short PR here against dev. Everything else (the package, the trained term, the drivers, the docs page) moves to jcm-experiments. I'll fold the re-score on current dev physics into that PR rather than posting it here.

duncanwp:
Great, sounds good. You should have an invite now!

Issue #768, opened 2026-09-09
Suspicious dust emissions


duncanwp:
@ajosyula1 noticed suspicious looking emissions maps diagnosed from his JAM run. Dig into and check we're getting a sensible burden

PR #776, opened 2026-09-09
JAM removal: Slinn/CAM below-cloud scavenging, wet-particle settling, and an operator-split removal chain (#722)

### LINKED ISSUE #722: Sea-salt burden ~4x low with τ = 0.23 d: the deficit is removal-side — emissions audited faithful vs HAMMOZ Gong
## Finding (2026-08 emission-parameter audit)

jcm's sea-salt source (`jam/emissions/seasalt.py`) is a faithful port of HAMMOZ's `seasalt_emissions_gong` (`mo_ham_m7_emi_seasalt.f90:721-862`, nseasalt=6): identical 300-bin spectrum constants (0.1–10 µm dry, `rm = 1.814·rd`, `u^3.41`, both Gong branch polynomials), same sea-ice/land masking semantics, mass/number internally consistent by construction. Deviations found, roughly cancelling:

- **−12%** emitted mass from the MAM4 sea-salt density 1900 kg/m³ vs HAM's 2165 (`species.py:35`; deliberate MAM4-provenance choice — deserves a code comment).
- **+35–40%** from feeding the **lowest-model-level wind (~30 m)** instead of a diagnosed u10 (`seasalt.py:159`; same shortcut in `dms.py:94`) — filed separately as its own fidelity gap.
- Accum/coarse split at ~0.67 µm dry (from MAM4 `dgnum_lo/hi`) vs HAM's fixed 1.0 µm: moves only 2.6% of mass; inflates coarse *number* 1.49× at smaller mean size (2.30 vs 2.61 µm volume-mean) — if anything this *slows* coarse settling.
- Net vs HAM's actual default scheme (LSCE tables, nseasalt=2): jcm emits 0.8–0.9× at typical winds. **Emission cannot explain a 4× burden deficit.**

## The actual gap

Burden 1.6–1.9 mg/m² (AeroCom ~5–13) *combined with* lifetime 0.23 d (obs 0.4–1 d) says the **sink is too strong**, not the source too weak. Suspects, in order:

1. **Wet scavenging of the coarse/accum soluble modes** — including the #708 ledger's fraction=1 marker semantics for fully-converting cells and the below-cloud impaction coefficients for the coarse mode.
2. **Coarse-mode sedimentation/dry deposition** post-#660 (the moment-weighted settling is *correct* but its interaction with the emission size split above changes the effective sink).
3. Possible double-counting between sedimentation and dry deposition for the coarse mode.

A useful first diagnostic: per-process sink partition for ss (wet vs dry vs sedi) from the flux ledgers over a spun-up window, compared against HAM's published partition (~30% wet / ~70% dry+sedi for sea salt).

Optional source-side upgrades (legitimate, but not the fix): SST-dependent schemes (HAM's nseasalt=7 Long/Sofiev or 8 Gong+SST) would raise tropical emission ~1.1–1.5×.

Context: 2-yr closed-budget run on PR #720; lifetime table in the #660 thread.

duncanwp:
Closes #722, #781. Refs #660, #658, #812.

The sea-salt burden is ~4x low with a lifetime of 0.23 d (obs 0.4–1 d) because the
**sink is too strong**, not the source too weak — the emission audit on #722 cleared
`jam/emissions/seasalt.py` as a faithful Gong/HAMMOZ port. This PR fixes the four
removal-side defects that audit found, each against CAM MAM4
(ESCOMP/CAM `cam_development`).

Commits are separable and can be reviewed in order.

---

## 1. Below-cloud impaction: the Slinn integral, not an r² extrapolation

`below_cloud_rate` was

```
Λ = below_coeff(1e-4) · R[mm/h] · (1 - cf) · (r_wet / 0.1 µm)²
```

— a coefficient tuned at 0.1 µm, extrapolated as r² with **no upper bound, no
Greenfield minimum and no saturation**. Per mm/h that gives coarse sea salt at 80 % RH
Λ = 3.2e-2 s⁻¹ against the Slinn/CAM 1.4e-4 (**226x**), coarse dust 84x, sea salt at
99 % RH 1700x. At 1 mm/h the coarse mode loses **100 % per 1800 s step**. That is why
#660's (correct) settling fix could not show up in the coarse mode.

**Replacement — CAM's treatment.** `wetdepa_v2`'s below-cloud term reduces to

```
odds  = min(1, precab/cldv · Λ₁ · Δt)
srcs2 = sol_factb · cldv · odds · q / Δt        ⟹   Λ = sol_factb · Λ₁(D_wet) · R
```

(the swept volume `cldv` cancels against the in-precip-area rain rate inside `odds`),
with `R` the precip flux in kg m⁻² s⁻¹ ≡ mm s⁻¹ and `Λ₁` in 1/mm. `Λ₁` is
`calc_1_impact_rate`'s double sum over a raindrop spectrum (50 bins, 0.005–0.25 cm,
N ∝ exp(-r/0.027 cm), renormalised to 1 mm/h) and the mode's lognormal (51 bins about
the volume-median radius), weighted by Slinn's collection efficiency

```
E_brown     = 4(1 + 0.4 Re^½ Sc^⅓)/(Re·Sc)
E_intercept = 4χ(χ + (1 + 2·60·χ)/(1 + 60/Re^½)),   χ = a/r
E_impact    = ((St − S*)/(St − S* + ⅔))^{3/2},      S* = (1.2 + ln(1+Re)/12)/(1 + ln(1+Re))
E           = min(E_brown + E_intercept + E_impact, 1)
```

integrated against **number** and **volume** weights separately — CAM's
`scavcoefnv(:,:,1)`/`(:,:,2)`, applied to the number and mass mixing ratios
respectively, which jcm now does too. `E ≤ 1` makes Λ₁ saturate at the rain's
geometric sweep-out rate, and the Brownian/impaction competition reproduces the
Greenfield gap. Neither survives an r² law.

**Tabulated, as CAM does.** The 50×51 sum is ~1e11 flop/step if evaluated per cell, so
`impaction.build_impaction_table` tabulates it once per mode over CAM's growth-ratio
grid (20 nodes, `ln(1.25)` apart in `D_wet/D_dry`, at 273.16 K / 750 hPa / the mode's
first-species density — `modal_aero_bcscavcoef_init`) and `bcscavcoef` looks it up by
differentiable log-linear interpolation, clamped below and linearly extrapolated
above, exactly like `modal_aero_bcscavcoef_get`. The runtime path is JAX,
differentiable in the wet radius, and broadcasting-native.

**Verified against CAM itself.** `calc_1_impact_rate` was extracted and compiled
standalone; the port reproduces it to 8 significant figures at 8 (D, σ, ρ) points
(`impaction_test.CAM_REFERENCE`), e.g. accumulation at 0.11 µm → (6.8273473378e-04,
8.1447204069e-04) 1/mm and coarse at 2 µm → (4.0869856241e-01, 1.7861883989e+00).

**`sol_factb`** replaces `below_coeff`/`below_radius_ref` as the differentiable knob,
at CAM's `sol_factb_interstitial` namelist default of **0.1**. Cloud-borne aerosol
keeps 0 structurally, as in CAM. *Decision to flag:* CAM's fallback when the namelist
leaves it unset is the mode's mass-weighted hygroscopicity — for sea salt that is
κ = 1.16, i.e. a "solubility" > 1, and no supported CAM configuration takes that path,
so it is not ported. Say the word if you want it as an option.

Net effect for coarse sea salt at 80 % RH, 1 mm/h: Λ falls ~470x (coefficient 47x,
`sol_factb` 10x), i.e. 11 % removed per step instead of 100 %.

## 2. Settle the wet particle, not a dry particle at the wet radius

`_jam_state.rho` from the κ-Köhler placeholder core was the **dry** material density
while `stokes_velocity` and the Slinn quasi-laminar resistance evaluate it at the
**wet** radius. CAM passes `wetdens` from `modal_aero_wateruptake`; the MAM4-JAX core
already publishes it, only the placeholder disagreed. It now returns

```
ρ_wet = (ρ_dry + (g³ − 1)·ρ_water)/g³,      g = r_wet/r_dry
```

the same expression CAM uses when it builds the impaction table. Measured for coarse
sea salt: 1900 → 1160 kg/m³ at 80 % RH (settling **1.64x** too fast before) and
1900 → 1008 at 99 % (**1.89x**); dust at 99 % RH is 2.16x.

*Scope note:* production (`jam_microphysics=mam4_jax`) already used `wetdens`, so this
one changes only the placeholder core — which is what `tools/release_validation/scm_check.py`
and most tests run. It does not move the T63 validation numbers below.

## 3. The removal chain now operator-splits

`ComposablePhysics` hands every term the step-start state and sums the tendencies.
Sedimentation (CFL cap), dry deposition and wet scavenging (implicit `1 − exp(−Λ·Δt)`)
each bound their own removal at 100 % of what they see — but the **sum was unbounded**:
a raining marine surface cell removed 0.35 + 0.29 + 1.00 = **164 %** of its coarse sea
salt, leaving the dycore to absorb a negative tracer.

Fixed by operator splitting, the treatment ECHAM and CAM use: each removal term reads
the state the previous ones left, reconstructed from the running tendency the driver
already publishes as `_tendency_run` (the mechanism the AeroCom diagnostics use).
Removing a fraction of what remains can never exceed the whole.

**Why not a joint limiter** that rescales the three tendencies to fit: a limiter
changes the answer only in the cells where it binds and leaves every other cell
double-counting the aerosol each process sees, and it needs cross-term communication
the driver has no place for. Splitting is what the reference models do everywhere. Its
cost is order dependence — settle, deposit, scavenge — which is both the composed order
and the physical one.

## 4. `dry_*` now includes dry deposition

`dry_<species>` omitted `SlinnDryDeposition` entirely (~45 % of the coarse-mode dry
sink; only sedimentation and wet scavenging called `accumulate_deposition_fluxes`), and
`flux_diagnostic`'s comment asserted the term did not exist. Dry deposition now
publishes into `dry_*`, cloud-borne removals included, so `dry_* + wet_*` is exactly the
mass the chain removed — asserted in `removal_split_test` against the composed chain's
actual interstitial + cloud-borne state change (to 1e-5 relative), not against the ledger
itself. An independent instrumented run on `dev` measured the dust Slinn sink at
1195 Tg/yr against 550 in `dry_du`, i.e. the ledger captured **46 %** of the dry sink;
it is now complete by construction. In the A/B runs `dry_ss` goes 277 → 974 Tg/yr and
`dry_du` 874 → 2778 Tg/yr, which mixes this ledger fix with the physics changes above.

---

## Tests

- `wetdep/impaction_test.py` — the 8 CAM reference points to 8 s.f.; the Greenfield
  minimum lands in 0.05–0.5 µm; Λ₁ saturates (50 µm/10 µm < 2x where r² would give 25x)
  and is monotone above the gap; table nodes reproduce the integral and interior points
  are bracketed; low-end clamp; `check_vjp`/`check_jvp` against numerics; column and
  (kx, ncols) block agree.
- `removal_split_test.py` — the composed chain never removes more than the available
  mass; the step-start reads *do* overdraw (124 % in the extreme cell) while sequential
  reads do not; `dry_* + wet_*` equals the chain's actual interstitial + cloud-borne
  mass change per species (a *state* assertion, not a ledger-against-itself check);
  `dry_*` now exceeds sedimentation alone; single column and block agree.
- `microphysics/placeholder_test.py` — ρ_wet matches the mixture formula at four RHs,
  reproduces the 1.64x/1.89x sea-salt ratios, stays bounded between water and the dry
  material, and reduces to ρ_dry in dry air.

## Validation — 30-day T63L47 A/B

`+configuration=ma-t63-l47`, same seed and dates, one A100, `origin/dev` (f5c8b35) vs
this branch. Global means over **days 10–30**; lifetime = burden / total deposition.
Both runs NaN-free throughout (6/6 chunks, verified from the output files).
Re-run twice after review: dropping the `1 - cf` weighting and then extending the
operator split to the coupled sulfur/cloud-borne transfers each moved sea salt by
under 2 %.

| species | burden mg/m² | | lifetime d | | wet fraction | | deposition Tg/yr | | target |
|---|---|---|---|---|---|---|---|---|---|
| | before | after | before | after | before | after | before | after | |
| **sea salt** | 1.46 | **3.78** | 0.242 | **0.485** | 0.75 | **0.33** | 1122 | 1452 | 5–15 mg/m², 0.4–1 d |
| dust | 12.80 | 18.53 | 0.766 | 0.786 | 0.72 | 0.38 | 3114 | 4395 | 10–40 mg/m², 3–5 d |
| so4 | 2.70 | 2.98 | 1.77 | 1.83 | 1.00 | 0.93 | 285 | 304 | |
| bc | 0.19 | 0.21 | 4.61 | 4.27 | 0.99 | 0.81 | 7.5 | 9.0 | |

Deposition split, Tg/yr:

| species | dry before → after | wet before → after |
|---|---|---|
| sea salt | 277 → **971** | 845 → 480 |
| dust | 874 → **2720** | 2240 → 1675 |
| so4 | 1.2 → 20.9 | 284 → 283 |
| bc | 0.04 → 1.7 | 7.5 → 7.2 |

**Sea salt is the target and it moves the right way**: burden **2.6x** (1.46 → 3.78
mg/m²) and lifetime **2.0x** (0.24 → 0.49 d), landing inside the observed 0.4–1 d band.
The pathway partition is the cleaner check #722 asked for — HAM publishes ~30 % wet /
~70 % dry+sedi for sea salt, and the run goes from **75 % wet** to **32 % wet**, i.e.
from wet-dominated to the observed partition. The burden is still short of 5–15 mg/m²,
consistent with this run's sea-salt source being 1453 Tg/yr against an AeroCom 3000–10000.

**Sulfate and BC barely move** (+10 %, +11 % burden), which is the expected null result:
the accumulation mode sits in the Greenfield gap where below-cloud impaction was never
the dominant sink. This PR therefore does not make #658 worse.

**Dust burden rises 45 % but its lifetime does not** (0.766 → 0.786 d) — deposition
rises with the burden. Dust is dominated by the emission-side defects in #768
(erodibility mask, mode split, emitted number), not by this.

Confounders, as expected: convective scavenging is dead on `dev` (#690 regression, being
fixed on a parallel branch) and a baseline mass leak is under separate diagnosis, so
absolute lifetimes are not expected to reach the targets from this PR alone.

Reproduce: `/glade/derecho/scratch/duncanwp/jam_runs/fix722_{before,after}`.

## Gates

`ruff check .` clean; PBS `develop` job: fast gate **2440 passed, coverage 95.23 %**
(≥90 required), slow gate **coverage 83.59 %** (≥80), `GATES PASSED`. One gotcha worth
recording: `.claude/skills/jcm-local-ci/scripts/local_ci.sh` does not put the
semi-Lagrangian dinosaur on `PYTHONPATH`, so on a host whose venv has the released
`dinosaur` every model-building test fails with the `_require_semi_lagrangian` error;
the gates above were run with `PYTHONPATH=~/dinosaur-sl:$REPO`.

## 5. Aerosol tracers were also outside the interface's non-negativity guard

`verify_tendencies` caps a sink at `-q/dt`, but `_NON_NEGATIVE_TRACERS` listed water,
cloud and GHG species and no aerosol or gas tracer — so nothing bounded the removal sum
at the interface either. Measured on `dev`: 8 of 18432 columns overdraw, inventing
0.029 Tg/yr of dust. `has_non_negative_tendency` now matches the JAM name families
(`m_`/`mc_`/`n_`/`nc_`/`g_`), which are built per population and cannot be listed.

The entry clip in `verify_state` is deliberately **not** extended: it would hide the
advection ringing on aerosol tracers from the #713 mass-budget gauge, which reads the
same verified state, and every removal term already floors its own reads at zero. This
is belt-and-braces behind the operator splitting, which is what actually bounds the sum.

That exposed a second, latent defect in the same helper: `_cap_negative_tend` returned
`-value/dt` whenever the step would end below zero, which for a tracer arriving
*negative* is a **positive** tendency — the guard would have invented mass to lift a
ringing negative up to zero. It now caps at `-max(value, 0)/dt`: drain to zero, never
fill to zero. Bit-identical for every field that was already covered (they are
entry-clipped, so `value >= 0`).

## 6. Convective washout acts in the updraft footprint (#781)

Folded in at review. Below-cloud impaction by convective rain removed aerosol from the
whole grid-mean interstitial population, through the cloud depth and below its base, on
top of the in-plume sink `ConvectiveTracerTransport` applies inside the updraft — the
same air removed twice under two area assumptions (#781). HAMMOZ partitions the
convective call by the carrier's own cover: `prep_wetdep_hydro` takes the precipitating
fraction to be the updraft area `f_cu = M_u/(ρ·w_u)`, `w_u = 2 m/s`, on the `pmfu` that
`cuflx` has tapered linearly in pressure below the cloud base (squared for mid-level
convection), and `ham_wetdep` removes `q_ambient·f_cu·(1 − exp(−Λ·Δt))` per layer with
`Λ` at the grid-mean flux.

`conv_precip_cover` rebuilds that footprint from `ConvectionData.mass_flux_up` — the
sub-cloud taper from the layer masses, because the published profile is the plume alone
and the tracer transport needs it that way; exact against ECHAM's `zzp` — and
`conv_below_cloud_rate` applies it to the removed fraction and converts back to the rate
the batched exponential update composes, so a step takes at most `f_cu` of a layer
however hard it rains. `w_u` is the new differentiable
`WetDepParameters.conv_updraft_velocity`, floored at 0.01 m/s in the division.

*Decisions to flag:*
- **The references disagree by the factor `f_cu`.** CAM's `wetdepa_v2` rescales the rain
  rate to the precipitating area so the area cancels; HAMMOZ evaluates `Λ` at the
  grid-mean flux, so its convective washout is `f_cu` (a few per cent) times CAM's. This
  item takes HAMMOZ's form for the convective carrier, as asked on #781; the
  CAM-consistent variant (`R/f_cu` inside the exponential, still capped at `f_cu` per
  step) is a one-line change in `conv_below_cloud_rate`. ECHAM's own sub-cloud rain
  evaporation uses the in-shaft intensity with this same cover, which argues for the
  variant.
- HAMMOZ's cloud-free gating of below-cloud scavenging (`paclc < 1e-10`) is not ported:
  the stratiform carrier follows CAM (no cover), and gating one carrier on the other's
  cover is the coupling #781 removed.
- The environment density stands in for HAMMOZ's updraft density (`zrhou = p/(rd·ptu)`;
  `ConvectionData` publishes no `tu`): `f_cu` is low by `(T_u − T_env)/T_env`, under 2 %.
- ECHAM under `lham` uses the same updraft area as the footprint of sub-cloud convective
  rain evaporation; jcm's `flux_tendencies.py` still carries the non-HAM `zcucov = 0.05`.
  Changing it moves the convective moisture budget, so it is #812 rather than this PR.

No run accompanies this item: it cuts the convective below-cloud sink by roughly `f_cu`,
and the 30-day A/B above predates both it and #775's restoration of the convective
carrier. Tests pin the cover (0.1 in-cloud, 0.05 / 0.025 sub-cloud for deep / mid-level
convection), the removed-fraction cap, the scaling of the washout with the mass flux and
its vanishing without a plume, flux-linearity under a uniform footprint, and a strictly
negative gradient through `w_u`.

## Documentation

`docs/source/design/jam_aerosol_removal.md` (new, in the toctree) records the reference
formulation for each pathway, the tabulation decision, the splitting decision and the
gaps that remain.

## Out of scope, deliberately

The convective below-cloud washout precipitation-flux profile was fixed in #775 (merged
here); the in-plume / washout coupling it left open became #781 and is item 6 above. Emissions (#768/#723) are also separate.


duncanwp (commented):
## Independent review (Claude, substituting for the quota-capped Codex run)

I diffed the port line-by-line against `ESCOMP/CAM cam_development`
`src/chemistry/modal_aero/aero_model.F90` (fetched fresh) and re-ran the new tests
serially on a Derecho login node (`JAX_PLATFORMS=cpu`,
`PYTHONPATH=$HOME/dinosaur-sl:<worktree>`): `impaction_test` + `placeholder_test` +
`removal_split_test` + `physics_interface_test` = **36 passed in 27 s**.

**The Slinn port is faithful.** Against `calc_1_impact_rate` (`aero_model.F90:1727-1962`)
every detail matches: the 50-bin 0.005–0.25 cm spectrum with `exp(-r/2.7e-2)`, the
five-branch `vfallstp` and the `sqrt(1.204e-3/rhoair)` density correction, the
`1.333333` in `precipsum` vs the `1.3333` in `yvolaerosv` (CAM really does use both,
and the port keeps them distinct), `xg0 = log(dg/2)`, `xg3 = xg0 + 3s²`,
`dx = max(0.2s, 0.01)`, `na = 1 + nint((xhi-xlo)/dx)`, the Fuchs correction, `taurelax`,
`schmidt`, `stokes`, all three efficiency terms, `min(E,1)`, the
`rainsweepout = xnum·4πr²·vfall` kernel (CAM's own factor of 4 — good that it was
carried rather than "corrected"), and the `×3600`. `build_impaction_table` matches
`modal_aero_bcscavcoef_init:1352-1455` including 273.16 K / 0.75e6 dyn cm⁻²,
`specdens_amode(1,mode)`, `dlndg = log(1.25)`, `nimptblgrow_mind/maxd = -7/12`, and
the `rhowetaero = rhodryaero` override. `_interp_log_table` reproduces
`modal_aero_bcscavcoef_get:1656-1724`'s clamp-below / linear-extrapolate-above.
`bcscavcoef`'s `2·r_wet / table.dgnum` is the right ratio: `population.ModalMode.dgnum`
is the dry *diameter* and `placeholder` publishes a number-median wet *radius*.
The wet-density algebra is also right — ρ_wet = (ρ_dry + (g³−1)ρ_w)/g³ is exactly
`modal_aero_wateruptake`'s mass-weighted mixture.

Findings, ranked:

### 1. `_cap_negative_tend` now silently discards **positive** tendencies on a negative tracer — blocker

`jcm/physics_interface.py:430-431`:

```python
capped = jnp.where(next_value < 0, -jnp.maximum(value, 0.0) / time_step, tend)
```

The branch is `next_value < 0` only — it does not ask whether `tend` is a sink. When
`value < 0` (which aerosol now routinely is, since it is deliberately excluded from
the `verify_state` entry clip) and `tend > 0` but too small to lift the tracer back to
zero in one step, the guard fires and sets the **source** to zero. Verified on this
branch:

```
state -1e-18, tendency +1e-22  ->  0.0        # source destroyed
state -1e-18, tendency +1e-15  ->  1.0e-15    # survives only because next_value >= 0
```

So in every ringing-negative aerosol cell, emissions, sulfur chemistry and wet-dep
re-injection are switched off until advection happens to bring the cell positive
again. That is the opposite of the stated intent ("drain to zero, never fill to zero")
— it also drains to zero the things that were not sinks.

Suggested fix, one line, provably identical to the old behaviour for `value >= 0`
(`max(tend, -value/dt) == -value/dt ⟺ value + dt·tend < 0`):

```python
capped = jnp.maximum(tend, -jnp.maximum(value, 0.0) / time_step)
```

and extend `test_negative_aerosol_tracer_is_not_topped_up` with the positive-tendency
case — it currently only exercises a negative one, which is why this passed.

### 2. The operator split silently no-ops on the 3D physics path — should-fix

`removal_split.split_view` falls back to the step-start view when `_tendency_run` is
absent (`removal_split.py:37-39`), and `_tendency_run` is published **only** in
`composable_physics._compute_tendencies_columns` (`composable_physics.py:357`) — never
in `_compute_tendencies_3d` (`:254-281`), which does publish `_dt_seconds`. JAM is
`vectorize_columns=True` only today (`cloud_borne_store.py:157` says so explicitly),
so this is latent rather than live — but a silent fallback that restores the exact
164 % overdraw this PR exists to remove is the wrong failure mode. Either publish
`_tendency_run` on the 3D path too (two lines, and it is genuinely useful there for
the same diagnostic reason), or make the absence loud for a JAM composition.

### 3. `split_view` folds in **every** upstream term, not just the removal chain — should-fix

`_tendency_run` is the running sum over terms `[0, i)`, and in `jam_aerosol_physics`
the terms before `StokesSedimentation` (`jam_terms.py:346`) are emissions
(`:238-258`), `ConvectiveTracerTransport` (`:301`), sulfur chemistry, the microphysics
core and `ArgActivation` (`:340`). So sedimentation now settles the sea salt and dust
emitted *this* step and the sulfate formed *this* step, minus whatever convection just
exported out of the layer; dry deposition and scavenging inherit that too.

That is a defensible *full* operator split and it may well be what you want — but it
is materially more than what `removal_split.py`'s docstring, `jam_aerosol_removal.md`
("each removal term reads the state the previous ones left") and the PR body describe,
and `removal_split_test._run_chain` only ever injects removal tendencies, so nothing
covers the wider behaviour. Please either scope the reconstruction to the removal
categories (`aerosol_sedimentation` / `aerosol_drydep` / `aerosol_wetdep`) or state and
test the wider split. A reviewer reading the doc would not expect emissions in there.

### 4. The `(1 - cf)` factor has no CAM counterpart, and now stacks with `sol_factb` — should-fix (doc)

In `wetdepa_v2` the swept volume cancels exactly, so CAM's grid-mean below-cloud rate
is `sol_factb·Λ₁·precab` with **no** clear-sky weighting — there is no `cldv` left to
"put `1 - cf` in place of". `jam_aerosol_removal.md:29-35` states the deviation and
gives the right justification (the in-cloud pathway is already keyed to `cf`, so the
pair partitions the box once), and I agree with the choice. But it is a *second*
independent reduction relative to CAM on top of `sol_factb = 0.1`, and a fully
overcast raining layer gets exactly zero impaction. Given the whole PR is calibrating
a sink that was 226x too strong, the doc should say what net bias against CAM this is
expected to leave, so the next person tuning it knows both factors are in play.

### 5. `_RHO_WATER = 1000.0` duplicates `jcm.constants.rhow` — nit

`jcm/physics/aerosol/jam/microphysics/placeholder.py:29`. CLAUDE.md makes
`jcm/constants.py` the single source of truth and `aerosol/jam/activation/arg.py`
already pulls `rhow` from it. Use `c.rhow` (and drop the hard-coded `RHO_WATER` in
`placeholder_test.py` in favour of the same constant, so the test cannot pass against
a stale literal).

### 6. Prefix matching in `has_non_negative_tendency` is broader than the JAM name space — nit

`physics_interface.py:344` `("m_", "mc_", "n_", "nc_", "g_")` is a bare `startswith`
over every tracer name a term returns. It is correct for today's tracers, but `n_` and
`g_` are two characters and any future tracer that happens to start with them gets
silently positivity-capped — a hard bug to find, because the symptom is a tendency
quietly clipped in one cell. The layout module already knows the real names; deriving
the predicate from `ModalAerosolSpec`/`tracer_layout` (or at least anchoring a regex
like `^m_[a-z0-9]+_[a-z0-9]+$`) removes the guesswork.

### 7. Comment style: campaign numbers and prior behaviour in code comments — nit

`removal_split.py:1-16` ("removed 164 % of its coarse sea salt"),
`placeholder.py:115-121` ("1.6x for coarse sea salt at 80 % RH, 1.9x at 99 %"),
`drydep_term.py:186-189` ("it is ~45 % of the coarse-mode dry sink"), and the 14-line
block at `physics_interface.py:331-345`. CLAUDE.md asks comments to describe the
current state and the *why*, 1–3 lines; measured magnitudes and descriptions of the
old behaviour go stale and are already captured in the (excellent) commit messages and
in `jam_aerosol_removal.md`. E.g. `placeholder.py` needs only *"ρ_wet = (ρ_dry +
(g³−1)ρ_w)/g³ — the settling and deposition velocities use the wet radius, so they
need the density of that same particle (CAM's `wetdens`)."*

### 8. "`dry_* + wet_*` equals the chain's total mass change exactly" is not quite true — nit

`jam_aerosol_removal.md:136`. Each term accumulates its own ledger *before*
`verify_tendencies`, so whenever the new interface cap binds — or any other term moves
an aerosol tracer — the identity breaks. `removal_split_test` asserts it against the
composed chain in isolation, which is right, but the doc should say "up to the
interface guard". (The ledger work itself is a clear improvement — `dry_*` omitting
`SlinnDryDeposition` was a real gap and folding cloud-borne removals in explicitly is
the right call.)

### 9. Two harmless deviations from `modal_aero_bcscavcoef_get`, worth a word — nit

(a) CAM short-circuits `0.99 <= dumdgratio <= 1.01` to node 0 exactly
(`aero_model.F90:1681-1684`); `_interp_log_table` always interpolates. Numerically
indistinguishable, but the port is otherwise node-for-node, so a one-line note stops
the next reader hunting for it. (b) CAM gates the lookup on `isprx`; jcm relies on
`Λ ∝ R` being zero without precip, which is equivalent and cleaner — also worth
saying, since it explains why there is no mask.

### 10. Cross-PR: collides with #775 — should-flag

#775 rewrites the same `below_cloud_rate` signature and the same `below_conv` call
site in `wetdep_term.py` (per-level convective `precip_flux` instead of the broadcast
surface value). Whichever lands second needs a real merge, and
`jam_aerosol_removal.md`'s Known-gaps bullet *"the convective below-cloud pathway
still drives every sub-cloud level with the surface convective precip flux"* should be
deleted at that point.

---

JAX hygiene looks clean throughout: no Python branching on traced values, the table
lookup's `maximum`/`clip` give zero (not NaN) gradients at the clamps, `check_vjp` /
`check_jvp` are exercised on a size scale rather than the metre-scale radius (the right
call for a piecewise-linear table), `sol_factb` stays a differentiable leaf, and the
`(kx,)` / `(kx, ncols)` agreement is covered in three separate places. Finding 1 is
the only one I would call blocking.

duncanwp:
Thank you — this caught a real blocker. All ten items answered; code fixes are in `7144b5bf`.

**1. `_cap_negative_tend` zeroed positive tendencies — Confirmed and fixed in `7144b5bf`.** Reproduced exactly as reported. The branch tested only "does the step end below zero", so on a negative tracer any source too small to lift it back was replaced by the drain rate. Now `capped = jnp.maximum(tend, -jnp.maximum(value, 0.0) / time_step)`, which is your suggestion and is provably identical to the old form for `value >= 0` (`max(tend, -v/dt) == -v/dt ⟺ v + dt·tend < 0`). `test_source_on_a_negative_tracer_survives` pins the positive-tendency-on-negative-state case.

**2. `_tendency_run` missing on the 3D host — Confirmed and fixed in `7144b5bf`.** `_compute_tendencies_3d` now publishes it with the same "sum over terms [0, i)" contract, and `TestTendencyRunHostParity` runs the same two-term composition through both hosts and asserts the second term saw the same running tendency.

**3. The split folds in every upstream term — Confirmed; documented and tested rather than narrowed.** It is the full sequential split and that is what the removal terms want (aerosol emitted or formed this step is there to be removed; aerosol convection has exported is not), so I have made the description match the behaviour instead of the reverse: `removal_split`'s docstring and `jam_aerosol_removal.md` now name emissions, convective transport, chemistry, the core and activation explicitly, and `test_upstream_sources_are_visible_to_the_removal_chain` injects a source ahead of the chain and asserts the chain removes more.

**4. `(1 - cf)` has no CAM counterpart — Confirmed and removed in `7144b5bf`.** You are right that `cldv` cancels exactly, and on re-reading, my justification did not survive: with the explicit cloud-borne phase the interstitial in-cloud pathway is zero (`form_num = form_mass = zeros`), exactly as CAM's interstitial `sol_facti = 0` — so there is nothing to partition the box against and the weighting was a second, unjustified reduction. `below_cloud_rate` is now `sol_factb · Λ₁ · R` on the grid-mean interstitial tracer, with `sol_factb` the only reduction. This strengthens the sink relative to the numbers currently in the PR body, so the 30-day A/B is re-running (PBS 7359407) and I will repost the table.

**5. `_RHO_WATER` duplicates `c.rhow` — Confirmed and fixed in `7144b5bf`**, in both the module and the test.

**6. Bare prefix matching — Confirmed and fixed in `7144b5bf`.** Replaced with an anchored pattern `^(?:mc?_[a-z0-9]+_[a-z0-9]+|nc?_[a-z0-9]+|g_[a-z0-9]+)$`, so an unrelated future `n_*`/`g_*` tracer is not silently capped.

**7. Comment style — Confirmed and fixed in `7144b5bf`.** Campaign numbers and prior-behaviour prose are out of `removal_split.py`, `placeholder.py` (now the one-line form you suggested), `drydep_term.py` and `physics_interface.py`; they remain in the commit messages and the design doc.

**8. "equals the chain's mass change exactly" — Confirmed and fixed in `7144b5bf`.** The doc now says "up to the interface guard", noting each term records its ledger before `verify_tendencies` sees the summed tendency.

**9. The two `bcscavcoef_get` departures — Confirmed and documented in `7144b5bf`**: the `0.99–1.01` short-circuit (numerically indistinguishable) and the absent `isprx` mask (equivalent, because `Λ ∝ R`).

**10. Collision with #775 — acknowledged.** #775 rewrites the same `below_cloud_rate` signature and the same `below_conv` call site. I have kept my diff there to the coefficient computation and the now-removed `cloud_fraction` argument. Whichever lands second gets a real merge, and the "surface convective precip flux" known-gaps bullet in `jam_aerosol_removal.md` goes at that point. Happy to rebase onto `dev` after #775 if you land it first.

Targeted suites re-run after the changes: `physics_interface_test`, `composable_physics_test`, `removal_split_test`, `impaction_test`, `wetdep_test`, `placeholder_test`, `drydep_test`, `sedi_test` — **137 passed**. Full gates re-running as PBS 7359408.

duncanwp:
Follow-up to the review response in `7144b5bf`.

**Gates, on the tree with all review fixes** (PBS `develop`, `PYTHONPATH=~/dinosaur-sl:$REPO`):
```
Required test coverage of 90% reached. Total coverage: 95.23%   FAST_EXIT=0
Required test coverage of 80% reached. Total coverage: 83.59%   SLOW_EXIT=0
GATES PASSED
```
`ruff check .` clean.

**Validation re-run** (required because dropping the `1 - cf` weighting strengthens the sink — finding 4). Same seed/dates, days 10–30; the PR body table is updated. Removing the weighting moved sea salt by **under 2 %**, so the conclusion is unchanged:

| species | burden mg/m² | lifetime d | wet fraction | deposition Tg/yr |
|---|---|---|---|---|
| **sea salt** | 1.46 → **3.82** | 0.242 → **0.496** | 0.75 → **0.32** | 1122 → 1433 |
| dust | 12.80 → 18.95 | 0.766 → 0.785 | 0.72 → 0.37 | 3114 → 4499 |
| so4 | 2.70 → 3.08 | 1.77 → 1.90 | 1.00 → 0.93 | 285 → 302 |
| bc | 0.19 → 0.21 | 4.61 → 4.39 | 0.99 → 0.81 | 7.5 → 9.0 |

Sea-salt lifetime is inside the observed 0.4–1 d band and the pathway partition is now 32 % wet / 68 % dry+sedi against HAM's ~30/70 — the diagnostic #722 asked for. so4/bc are essentially unchanged, as expected for a mode sitting in the Greenfield gap, so this does not worsen #658.

That the `1 - cf` removal barely moved the run is itself worth recording: in the cells that dominate the sea-salt sink, cloud fraction is low where the precip flux is large, so the two reductions were not stacking as hard as the algebra suggests. The change is still correct on reference grounds and is what the doc now describes.

duncanwp:
GitHub CI is green on `7144b5bf`: `build (3.11)` **pass** (both workflow runs, ~53 min each), `lint` **pass**, `claude-review` **pass**. Earlier in this session I described the build check as stuck — it was simply slow, and it has now completed.

That, with the PBS gates (`GATES PASSED`, fast 95.23 % / slow 83.59 %), the re-run 30-day A/B in the PR body, and every review finding answered above, is the full set of checks this PR needs. Ready for human review.

Two things for the reviewer that are not blockers:
- `.claude/skills/jcm-local-ci/scripts/local_ci.sh` does not put the semi-Lagrangian dinosaur on `PYTHONPATH`, so on a host whose venv carries released `dinosaur` its PBS gate fails wholesale on `_require_semi_lagrangian`. Worth fixing separately; the gates above were run with `PYTHONPATH=~/dinosaur-sl:$REPO`.
- #775 rewrites the same `below_cloud_rate` signature and `below_conv` call site. Happy to rebase onto `dev` if it lands first; the "surface convective precip flux" known-gaps bullet in `jam_aerosol_removal.md` should be deleted at that point.

duncanwp:
@codex review

duncanwp:
@codex review

jcm/physics_interface.py
```diff
@@ -414,7 +436,7 @@ def _cap_negative_tend(value, tend):
     clipped_tracer_tends = {
         name: (
             _cap_negative_tend(state.tracers[name], tend)
-            if name in _NON_NEGATIVE_TRACERS and name in state.tracers
+            if has_non_negative_tendency(name) and name in state.tracers
             else tend
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in `aa3b0fce`.

You are right, and it is a regression my own change introduced: before this PR the aerosol/gas tendency cap did not exist, so an oversubscribed `g_so2` simply went negative — wrong, but sulfur-conserving. Adding a per-tracer cap turned that into a silent source, because the cap shrinks the sink and leaves the paired product alone.

Verified reachable on both paths you name:
- `chemistry/sulfur_gas.py::sulfur_tendencies` consumes `g_so2` from step-start `state.tracers` (bounded by `so2 + so2_from_dms`), and `chemistry/aqueous.py::AqueousSulfur` independently consumed the same step-start `g_so2` via `tracer_view` and credited `so4`/`mc_so4` at the S-conserving ratio. Two independent sinks on one reservoir.
- `cloud_borne.py::CloudBorneExchange` read step-start interstitial (i.e. *before* the sedimentation and dry-deposition tendencies in the same step) and credited the carry immediately.

Fixed by the mechanism this PR already uses rather than a new limiter: both now read `removal_split.split_view`, so aqueous oxidation sees the SO2 gas-phase chemistry has already consumed, and the exchange transfers interstitial mass that survived removal. `SulfurGasChemistry` precedes `AqueousSulfur` in `jam_terms.py::jam_aerosol_physics` (pre-core vs post-core), and `CloudBorneExchange` follows both removal terms, so the ordering the fix relies on is the composed one. Each sink is individually bounded by what it sees, so sequentially they cannot oversubscribe — which also means the interface cap is no longer load-bearing for conservation, only a last resort.

I preferred this to rescaling both sides of each transfer: a limiter would have to know every producer/consumer pairing (sulfur, activation transfer, resuspension, aqueous mode distribution) and would only act where it binds, leaving every other cell reading a stale reservoir. `jcm/physics/aerosol/jam` fast suite: 253 passed. Full gates re-running.

duncanwp:
Final verification on `aa3b0fce` (docs commit + the Codex P1 fix).

**Gates:** `GATES PASSED` — fast **2457 passed, 95.26 %**; slow **105 passed, 83.59 %**; no failures. `ruff check .` clean. `pytest docs/science_pointers_test.py` 9 passed.

**30-day A/B, re-run on this tree** (6/6 chunks NaN-free, verified from the output files). Extending the operator split to the coupled sulfur and cloud-borne transfers moved every species by **under 1.5 %** — the fix closes a conservation hole without disturbing the result:

| species | burden mg/m² | lifetime d | wet fraction | deposition Tg/yr |
|---|---|---|---|---|
| **sea salt** | 1.46 → **3.78** | 0.242 → **0.485** | 0.75 → **0.33** | 1122 → 1452 |
| dust | 12.80 → 18.53 | 0.766 → 0.786 | 0.72 → 0.38 | 3114 → 4395 |
| so4 | 2.70 → 2.98 | 1.77 → 1.83 | 1.00 → 0.93 | 285 → 304 |
| bc | 0.19 → 0.21 | 4.61 → 4.27 | 0.99 → 0.81 | 7.5 → 9.0 |

Sea-salt lifetime sits inside the observed 0.4–1 d band and the pathway split is 33 % wet / 67 % dry+sedimentation against HAM's published ~30/70. Sulfate and BC are essentially unchanged, as expected for a mode in the Greenfield gap, so #658 is not made worse.

**Also in this round:** `docs/source/science/aerosol.md` gains an aerosol-removal subsection, per the living-document rule #771 added to CLAUDE.md after this branch started (I merged `dev` first — the science tree and its guard test did not exist at my merge base). Headings are stable and the convective in-plume subsection is untouched for #775.

Ready for review.

duncanwp:
@codex review

jcm/physics_interface.py
```diff
@@ -397,8 +415,12 @@ def _cap_negative_tend(value, tend):
...
         # silently detaching those cells from any parameter being
         # calibrated. Positivity in the forward pass is unaffected.
-        next_value = value + time_step * tend
-        capped = jnp.where(next_value < 0, -value / time_step, tend)
+        # Floor the tendency at the drain rate that empties the tracer and
+        # no further. ``max(tend, -max(value,0)/dt)`` equals the plain
+        # ``-value/dt`` cap wherever ``value >= 0``, leaves any source
+        # untouched, and on a tracer that arrives negative (aerosol is not
+        # entry-clipped) stops the sink rather than inventing mass.
+        capped = jnp.maximum(tend, -jnp.maximum(value, 0.0) / time_step)
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in `29fb4b95`.

Verified: `vertical_diffusion/tracer_diffusion.py` and `convection/tracer_transport.py` both read step-start `state.tracers` and both return column-conserving redistributions, and `jam_terms.py::jam_aerosol_physics` runs them back to back in `pre_core`, before any split-view consumer. Your arithmetic is exactly right: the summed tendency leaves the column total exact and the donor cell negative, and the cap then clamps the donor while both receiving cells keep their gain.

This is the third instance of one class, and the third makes the general point: **a per-cell positivity cap cannot be applied to a tendency that carries a conservative redistribution.** The interface sees only the summed tendency — it cannot tell a sink from the donor half of a transfer — so no pairing-aware limiter can live there. I fixed the first two instances by sequencing the *coupled* terms, but transport is genuinely parallel physics and sequencing it would change the answer, not just the bookkeeping.

So the cap is removed for aerosol and gas tracers entirely. Its only purpose was to bound the summed removal, and the operator split already does that where the removal is produced — each sink sees what its predecessors left, so the sum cannot exceed the tracer. Residual ringing negatives go back to being the dycore's positive-definite tracer filter's job, as they were before this PR. Net effect on conservation: strictly better than either the pre-PR state or the capped state.

The test that asserted the cap is replaced by the invariant it was breaking — `test_conservative_transport_of_aerosol_survives_the_interface` feeds your [0,1] / [1,0] / [0.083,0.917] example through `verify_tendencies` for `m_ss_cor`, `n_cor` and `g_so2` and asserts the column sum is untouched — plus `test_water_tracers_are_still_capped` pinning that the cap still applies to the pure-sink fields it was written for. `jcm/physics_interface_test.py` 14 passed; `docs/science_pointers_test.py` 9 passed. Both design and science docs updated to state that no cap is applied and why. Full gates re-running from a clean worktree.

duncanwp:
@codex review

duncanwp (commented):
**Independent review of the post-Codex delta (Claude on Opus, substituting for the quota-capped run), tree `29fb4b95`**

Read the P1 thread, `git show 29fb4b95`, the split chain and both dycore backends. Tests run serially on CPU: `physics_interface_test.py` + `removal_split_test.py` + `wetdep/` + `microphysics/placeholder_test.py` + `docs/science_pointers_test.py` = **71 passed**; `jcm/physics/aerosol/jam -m "not slow"` = **306 passed, 8 deselected**.

### The core argument holds

Confirmed independently. `TracerVerticalDiffusion` (`tracer_diffusion.py:151-174`) and `ConvectiveTracerTransport` (`tracer_transport.py:355-412`) are composed back to back in `pre_core` (`jam_terms.py:279-306`, `318-325`), both read step-start `state.tracers`, and both return column-conserving redistributions. The interface sees only the sum and structurally cannot tell a donor from a sink, so no pairing-aware limiter can live there. Removing the cap is the right call, not a descope.

One piece of supporting evidence the reply and the design doc both miss, and which strengthens the case: with the cap in place `split_view` (`removal_split.py:44-50`) reconstructed the working copy from tendencies the interface would *later* clamp, so the split view was itself wrong wherever the cap fired. Removing the cap is what makes the operator split exact. Worth a line in `jam_aerosol_removal.md`.

"Strictly better than pre-PR" holds for the removal sum (pre-PR had neither cap nor split; each sink now sees only what remains, and every consumer floors its read at 0 — `sedi_term.py:225`, `drydep_term.py:161`, `wetdep_term.py:524/532/549`, `cloud_borne.py:200-201`). I could not construct a JAM path that drives a tracer negative post-delta where pre-PR would not: the split only ever *reduces* what a sink sees. The transport donor overdraw is unchanged from pre-PR — but see F2 on what happens to it afterwards.

---

### F1 — P2 — `jcm/physics_interface.py:389-403` (with `:328-339`, `physics_interface_test.py:229`): the rule the docstring now states is violated by the set the cap is kept for

The docstring's new sentence is the load-bearing generalisation of this PR: the cap "is only valid for a field whose tendency is a pure sink plus sources — never for one carrying a conservative redistribution". `_NON_NEGATIVE_TRACERS` retains `specific_humidity`, `qc` and `qi`, and those tendencies *do* carry a conservative vertical redistribution: `TTETKEVerticalDiffusion` declares `qc`/`qi` as required tracers (`tte_tke/vertical_diffusion.py:608-612`) and emits `qc_tend`/`qi_tend` from the implicit diffusion solve (`:769-770`, `:815`) — a donor/receiver pair, exactly like tracer diffusion — and its `specific_humidity` tendency likewise. Summed with the cloud/convection sinks, which are also computed from the step-start state, a donor layer can be overdrawn; `_cap_negative_tend` then clamps the donor and leaves the receiving layers' gain, creating water mass. Same failure class, same mechanism, in the model's most load-bearing field.

`test_water_tracers_are_still_capped` now pins that behaviour as correct-by-design without establishing that q/qc/qi are pure-sink, so the PR leaves a stated invariant its own retained set does not satisfy. Per CLAUDE.md's "same failure class rolls into the PR", I would expect either the same treatment or an explicitly recorded decision — at minimum narrow the docstring to say *which* fields are believed pure-sink and name the vdiff redistribution as the known exception, rather than asserting a rule the code does not honour.

### F2 — P2 — `docs/source/science/aerosol.md:284` (and the commit message): the "dycore positive-definite tracer filter" does not exist

This is the sentence the whole justification rests on, and I cannot find the mechanism it names on either backend:

- `jcm/dycore/dinosaur/dycore.py:738-785` — `_fix_nodal_tracer_mass` is a *global proportional mass fixer*; its own docstring says it "must never manufacture sign". It rescales by one factor per step, clipped to `[2/3, 1.5]`; it does not clip negatives.
- The SL quasi-monotone limiter (`dinosaur/semi_lagrangian.py:577`, `:627`) clips against the *bracketing cell values*, so it prevents new extrema from interpolation and faithfully preserves an existing negative.
- `jcm/dycore/pyses/dycore.py:819-827` — the production JAM backend applies `maximum(..., 0)` to `water_vapor` only; `passive_forcing` and the passive tracers pass through untouched.
- `verify_state` deliberately excludes aerosol, and the repo's own tests say the capability is absent: `jcm/physics/echam/echam_column_moisture_test.py:29` and `:151` — "would need positive-definite tracer advection in the dinosaur dycore".

So after this delta a JAM aerosol negative has **no** downstream corrector anywhere. That may well still be the right call — it is the pre-PR behaviour and it keeps the #713 gauge honest — but the docs should say that plainly ("nothing removes it; it stays visible to the #713 budget gauge by design") instead of deferring to a filter that isn't there. Same wording needed in the commit message and in the design doc's neighbouring claim.

### F3 — P3 — `jcm/physics/aerosol/jam/chemistry/aqueous.py:274-286`: the `simple` scheme does not floor its split-view read

Every other split-view consumer floors at 0, with the comment that a removal rate applied to a negative *injects* mass. `_aqueous_so4` floors `so2`/`so4` internally, but the `simple` branch passes `so2=so2` straight through, so a negative `g_so2` gives `dso4 < 0` → a positive `so2_rate` and negative sulfate production. It is S-conserving, so not a mass source, but it is a sign-flipped source/sink pair — and it matters more after this delta precisely because the split view now folds in the transport donor overdraw the delta accepts and no longer caps. One line: `so2=jnp.maximum(so2, 0.0)`, matching the branch beside it.

### F4 — P3 — `jcm/physics/aerosol/jam/removal_split.py:12-19`: the docstring overstates the chain

It claims the reconstruction folds in every term already run, "the microphysics core" included, but the core reads `tracer_view` (step-start) at `microphysics/mam4_jax.py:424`, as does `ice_nucleation/ice_term.py:78-80`. The largest tracer-transforming term is outside the chain it heads. I could not construct a positivity failure from it (the core's internal processes are conservative, and its `g_h2so4` consumer is fed by a preceding *source*), so this is accuracy rather than correctness: either move the core onto `split_view` or narrow the docstring to the removal chain.

### F5 — P3 — dead code: clean, with one leftover

`import re`, `_JAM_TRACER_RE` and the four superseded tests are all gone, and `_NON_NEGATIVE_TRACERS` was never extended by this PR (net diff against `3cb28a5b` confirms), so there is nothing to unwind there. `has_non_negative_tendency` is now a one-line alias for frozenset membership, referenced only at `physics_interface.py:441` and from two docstrings. Keeping it as the named predicate is defensible, but if it stays it should say it is deliberately *just* the membership test, so a future reader does not re-add a name family without re-reading the redistribution argument.

---

Nothing here blocks the direction of the delta, which I think is correct. F1 and F2 are the two I would want resolved before merge: one is a stated invariant the code does not satisfy, the other a documented mechanism that does not exist.

duncanwp:
Thank you — this is the review the delta needed. All five addressed in `acd1fc39`; **F2 is partly refuted**.

> **F1** — the rule the docstring now states is violated by the set the cap is kept for

**Confirmed and fixed.** Verified: `TTETKEVerticalDiffusion.required_tracers` declares `qc`/`qi` and the term emits `qc_tendency`/`qi_tendency` (and a `specific_humidity` tendency) from the implicit diffusion solve — a donor/receiver pair, exactly as you say. So the retained set does not satisfy the rule I asserted, and the test pinned that as correct-by-design without establishing it.

Decided on the evidence rather than by symmetry: the cap **stays** for water, because the downstream moist physics divides by and takes logs of `q` and `verify_state`'s companion entry clip exists for the same reason (PR #458). Removing or reshaping it is a moist-physics change needing TOA/precip/cloud validation — the "different subsystem, own validation campaign" carve-out in CLAUDE.md — not something to fold into an aerosol-removal PR. What I have changed is the honesty: the docstring now says the cap is sound only for a pure sink-plus-sources field, that the retained water fields **do not** satisfy it because vdiff redistributes them, and that it is kept for a named reason; the test is renamed `test_water_tracers_are_capped_despite_the_redistribution` and its docstring says it pins behaviour, not correctness; the design doc agrees. Tracked in **#806** with the measurement a fix would need.

> **F2** — the "dycore positive-definite tracer filter" does not exist

**Partly refuted, and the doc fixed anyway.** You are right that nothing corrects a physics-side negative *within* physics, and right about `_fix_nodal_tracer_mass`, the SL limiter and the pySES water-vapour-only floor. But the filter I meant does exist: `jcm/filters.py::MassConservingPositivity`, a column-mass-conserving hole-filler (clip negatives, rescale the surviving positive part), applied by the dinosaur core at the dynamics→physics projection (`dycore/dinosaur/dycore.py`, `tracer_filter`) and enabled **by default for JAM runs** — `diffusion.tracer_positivity: auto` resolves to on when `aerosol_module == "jam"` (`runners.py::build_tracer_filter`).

So the accurate statement, now in both docs: a negative left by the parallel transport terms is not removed in physics; it persists through the dynamics step and is cleaned on the way back **in** at the next boundary on the dinosaur backend, while under pySES (no `tracer_filter` wiring) or with the filter switched off it persists and stays visible to the mass-budget gauge. It is a boundary guard, not positivity-preserving transport. I have **not** linked #521 — it is closed, and the science register's rule is that an issue number means a live gap.

> **F3** — the `simple` branch does not floor its split-view read

**Confirmed and fixed**: `so2=jnp.maximum(so2, 0.0)`, matching the branch beside it. Your reasoning for why it matters more after this delta is right.

> **F4** — the `removal_split` docstring overstates the chain

**Confirmed and fixed.** Narrowed to the terms that return tracer *tendencies*, and it now says explicitly that the MAM4 core and ice nucleation read `tracer_view` and are outside the split. I took the docstring route rather than moving the core, for the reason you give — it is accuracy, not correctness.

> **F5** — `has_non_negative_tendency` is now a bare alias

**Confirmed and fixed**: it now says it is deliberately just the membership test, with a warning not to re-add a name family without re-reading the redistribution argument.

Also folded in your supporting point, which I had missed and which does strengthen the case: with the cap in place `split_view` reconstructed the working copy from tendencies the interface would later clamp, so **removing the cap is what makes the operator split exact**. That is now in `jam_aerosol_removal.md`.

Tests after the changes: `docs/science_pointers_test.py` + `physics_interface_test.py` + `aerosol/jam/chemistry` = 59 passed. Full gates re-running from the clean detached worktree at `acd1fc39`.

duncanwp:
@codex review

duncanwp:
Note for the 30-day T63L47 table in this PR: that run is a **cold start from `init=jw`**, so its circulation is still spinning up and the numbers are a before/after contrast on identical unspun states, not a climatology — they should not be read against observational totals.

Recorded while validating #804 (Tegen dust), where the same cold-start setup produced a global dust emission 35x below expectation purely because the near-surface winds had not spun up (mean 10 m wind 3.19 m/s over the source regions against a 7.6 m/s saltation onset). Threshold-gated quantities collapse on such a state where smooth ones only degrade, so it is worth knowing which kind a given table reports.

jcm/physics/aerosol/jam/wetdep/impaction.py
```diff
@@ -0,0 +1,232 @@
...
+    r, xnum, vfall = _rain_spectrum(rho_air)
+    a, fnum, fvol = _aerosol_spectrum(dg_cgs, geom_std_dev)
+
+    reynolds = r * vfall / kin_visc                     # (nr,)
+    sqrt_re = np.sqrt(reynolds)
+
+    chi = a[None, :] / r[:, None]                       # (nr, na)
+
+    dum = freepath / a
+    fuchs = 1.0 + 1.246 * dum + 0.42 * dum * np.exp(-0.87 / dum)
```
duncanwp: Lots of magic number here. where do they come from?

jcm/physics/aerosol/jam/wetdep/impaction.py
```diff
@@ -0,0 +1,232 @@
...
+_RAIN_R_LO_CM = 0.005
+_RAIN_R_HI_CM = 0.250
+_RAIN_DR_CM = 0.005
+#: Ratio of water to air dynamic viscosity used by Slinn's interception term.
+_MU_WATER_AIR = 60.0
+#: CAM's tabulation reference state (``modal_aero_bcscavcoef_init``).
+_REF_TEMPERATURE_K = 273.16
+_REF_PRESSURE_PA = 0.75e5
+#: cgs constants matching CAM ``mo_constants`` (boltz_cgs, rgas_cgs).
+_BOLTZ_CGS = 1.38065e-16
```
duncanwp: Don't do this. You're explicitly asked to use the shared constants module- use that

jcm/physics/aerosol/jam/wetdep/impaction.py
```diff
@@ -0,0 +1,232 @@
...
+import jax.numpy as jnp
+import numpy as np
+
+# Raindrop spectrum: radii 0.005–0.250 cm in 0.005 cm steps, number
+# concentration ∝ exp(-r/0.027 cm), renormalised to 1 mm/h.
+_RAIN_R_LO_CM = 0.005
+_RAIN_R_HI_CM = 0.250
+_RAIN_DR_CM = 0.005
+#: Ratio of water to air dynamic viscosity used by Slinn's interception term.
+_MU_WATER_AIR = 60.0
```
duncanwp: This sounds like a pretty uncertain parameter - make it a differentiable parameter in the physics term? I want at least a couple of knobs on impaction as it's inherently very uncertain

jcm/physics/aerosol/jam/wetdep/impaction.py
duncanwp: Fixed in `97d47857`. You're right — I had no business redefining them.

`_BOLTZ_CGS`, `_RGAS_CGS` and the hard-coded `28.966` g/mol air molar mass are gone. They now come from the shared module by attribute access with an explicit conversion at use, so `set_constants` is honoured:

```python
import jcm.constants as c
_BOLTZ_CGS = c.ak * 1.0e7          # erg/K   (1 J = 1e7 erg)
_RGAS_CGS  = c.r_universal * 1.0e7 # erg/K/mol
_M_AIR_CGS = c.m_air * 1.0e3       # g/mol   (1 kg/mol = 1e3 g/mol)
```

Nothing was missing from the module — `ak` and `r_universal` are base fields and `m_air` is already derived as `r_universal/rd`, so no addition is needed.

**Effect, since you asked for the number.** jcm's values differ from CAM `mo_constants`:

| | jcm | CAM | rel. diff |
|---|---|---|---|
| Boltzmann | 1.3806504e-23 | 1.38065e-23 | 2.9e-7 |
| gas constant | 8.314462 | 8.3144724 | 1.3e-6 |
| air molar mass | 28.96621 g/mol | 28.966 g/mol | 7.4e-6 |

That moves the tabulated coefficient by **<1e-4 relative** — physically nil, but it does break an eight-significant-figure pin. Rather than just loosening the tolerance and losing the strongest check in the file, I made the cgs values arguments defaulting to the shared constants: `test_matches_cam_calc_1_impact_rate` passes CAM's own values and **keeps `places=8`**, so the *algorithm* is still pinned exactly; a new `test_shared_constants_move_the_coefficient_negligibly` runs the same eight cases with jcm's constants and bounds the difference at 1e-4. Production takes the shared-constants path.

jcm/physics/aerosol/jam/wetdep/impaction.py
duncanwp: Fixed in `97d47857`. Every one of them is now a named module constant with its source, so the block can be checked without opening the Fortran.

**Sutherland's law** for the dynamic viscosity of air, `μ = μ₀·(T_S₁/(T + T_S₂))·(T/T_S₃)^{3/2}` [poise]:
`_VISC_MU0_CGS` 1.8325e-4, `_VISC_TS1_K` 416.16, `_VISC_TS2_K` 120.0, `_VISC_TS3_K` 296.16.

**Fuchs slip correction**, `C = 1 + 1.246·Kn + 0.42·Kn·exp(-0.87/Kn)` (Fuchs 1964; the Davies coefficients CAM and Seinfeld & Pandis both use): `_FUCHS_A/B/C`. The mean free path `λ = _MFP_COEF_CGS/c_air` keeps CAM's 2.8052e-10 with its units spelled out (cm·mol/cm³).

**Slinn (1983) collection efficiency**:
- Brownian `E = 4(1 + a·Re^½·Sc^b)/(Re·Sc)` → `_BROWN_A` 0.4, `_BROWN_SC_EXP` 1/3.
- critical Stokes `S* = (c₁ + ln(1+Re)/c₂)/(1 + ln(1+Re))` → `_SSTAR_C1` 1.2, `_SSTAR_C2` 12.
- inertial impaction `E = ((St − S*)/(St − S* + c))^p` above S* → `_IMPACT_C` 2/3, `_IMPACT_P` 1.5.

**Raindrop spectrum**: `_RAIN_R0_CM` 0.027 (exponential Marshall-Palmer-like `N(r) ∝ exp(-r/r₀)`), and the five-band terminal-fall-speed fit is now a table of `(d_max, A, B)` triples (`_VFALL_BANDS`/`_VFALL_TAIL`) rather than a nested `where`, with `_VFALL_RHO_STP_CGS` naming the 1.204e-3 g/cm³ the fit is referenced to and `v ∝ √(ρ₀/ρ)` written out.

Each carries a one-line comment giving the formula and the source, per CLAUDE.md — no history prose.

jcm/physics/aerosol/jam/wetdep/impaction.py
duncanwp: Agreed, and done in `97d47857` — you now have **three** live knobs on impaction. I measured both designs before choosing rather than bolting on a multiplier.

**Measured (CPU, 4 MAM4 modes, 20 growth nodes):** rebuilding the whole Slinn integral in JAX costs **2.19 s compile, 1.26 ms per call, 1.34 MB temporaries**. Per step that is negligible in flops — the table is 20 nodes × 4 modes = 80 evaluations against 385k cells × 4 modes if it were done per cell, ~19,000× less work — but it is ~2,000 small kernels a step, which is launch-bound on GPU and wasteful for what it computes.

**What I did instead, because the measurement showed something better.** Only *two scalars* are actually traced. So the table is split rather than wholesale-JAXed: everything knob-independent — drop and aerosol geometry, Reynolds/Schmidt/Stokes, the Brownian efficiency, the inertial efficiency at unit scale, the sweep-out and size weights — is built once per mode in **float64 NumPy** (`build_impaction_table`), and only the knob-dependent completion runs in JAX inside the step (`table_log_coefficients`):

```
E = min(E_brown + 4χ(χ + (1 + 2μχ)/(1 + μ/√Re)) + s·E_impact, 1)
```

That is ~10 elementwise ops on a (20, 50, 51) array per mode — cheaper than the full JAX build, cheaper than it would be to re-derive, and it gives the knobs a gradient **into the collection physics itself**, not a post-hoc factor on the answer. It also keeps the float64 reference integral intact for the CAM-parity test.

**The knobs**, on `WetDepParameters` (the term already holds it via `nnx.Param`), defaults exactly at today's values so **results are unchanged**:

| knob | default | what it is |
|---|---|---|
| `sol_factb` | 0.1 | below-cloud solubility factor (CAM `sol_factb_interstitial`) |
| `mu_water_air` | 60.0 | water/air viscosity ratio in Slinn's interception term |
| `impact_scale` | 1.0 | multiplier on the inertial-impaction efficiency |

Interception and inertial impaction are the two least-constrained mechanisms in the integral, so that is where I put the handles.

**Gradient checks**: `test_both_knobs_carry_a_gradient` (finite, non-zero for both), `test_knob_gradients_match_central_differences` (autodiff vs a central difference in the knobs' own units, to 5 %), and `test_defaults_reproduce_the_reference_integral` (the traced kernel equals the float64 integral at every growth node to 4 decimals).

One thing worth flagging, because it surprised me and I initially wrote the test wrong: **`mu_water_air` is not monotone.** In `(1 + 2μχ)/(1 + μ/√Re)` the sign of ∂/∂μ is that of `2χ − 1/√Re`, so raising μ *lowers* collection for particles small enough that `χ < 1/(2√Re)` and raises it for larger ones. `impact_scale` is monotone. Both the test and the design doc now say so.

duncanwp:
@codex review

jcm/physics/aerosol/jam/wetdep/impaction.py
```diff
@@ -0,0 +1,383 @@
...
+#: CAM's tabulation reference state (``modal_aero_bcscavcoef_init``).
+_REF_TEMPERATURE_K = 273.16
+_REF_PRESSURE_PA = 0.75e5
+
+# cgs values of the shared constants, converted at use: 1 J = 1e7 erg,
+# 1 kg/mol = 1e3 g/mol. CAM's own ``mo_constants`` differ from these in the
+# 6th-7th significant digit (see ``impaction_test``).
+_BOLTZ_CGS = c.ak * 1.0e7                    # erg/K
+_RGAS_CGS = c.r_universal * 1.0e7            # erg/K/mol
+_M_AIR_CGS = c.m_air * 1.0e3                 # g/mol
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in `4ed73672`. Good catch — I used attribute access but bound it at **module** level, which reintroduces exactly the failure CLAUDE.md's rule exists to prevent, just one scope out.

`_cgs_constants()` now resolves them on every call:

```python
def _cgs_constants(boltz_cgs=None, rgas_cgs=None, m_air_cgs=None):
    return (c.ak * 1.0e7 if boltz_cgs is None else boltz_cgs,
            c.r_universal * 1.0e7 if rgas_cgs is None else rgas_cgs,
            c.m_air * 1.0e3 if m_air_cgs is None else m_air_cgs)
```

Both call sites go through it — `impaction_scavenging_rates` (the reference integral) and `_kernel_at` (the static per-mode kernel `build_impaction_table` stacks), so a table built after an override uses the overridden air properties. Explicit values still win, which is how the CAM-parity test passes CAM's own `mo_constants` and keeps its eight-significant-figure pin.

Two tests pin it, since the previous version would have passed anything weaker: `test_set_constants_after_import_is_honoured` overrides `r_universal` by 2 % after import, asserts the volume coefficient moves by more than 1e-4 relative, restores, and asserts it returns to 10 decimals; `test_table_build_also_reads_the_live_constants` asserts the stacked `e_brown` kernel changes under the same override. Both restore the defaults in `tearDown` so they cannot leak into other tests.

docs/source/science/aerosol.md
```diff
@@ -416,6 +416,131 @@ overlap guards, the region vector, the snow and saturation cut-offs, the
...
+so the two collection knobs below stay differentiable. ``sol_factb`` is a differentiable
+parameter at CAM's namelist default 0.1 for interstitial aerosol and structurally
+zero for the cloud-borne phase, which is in-droplet by definition. Two further
+differentiable knobs sit on the collection integral itself — ``mu_water_air``
+(Slinn's water/air viscosity ratio, 60) and ``impact_scale`` (a multiplier on
+the inertial-impaction efficiency, 1) — because impaction is the least
+constrained part of the scheme; both default to CAM as written; no cloud
+weighting is applied, because the swept precipitating volume cancels against the
+in-precip-area rain rate and the stratiform in-cloud pathway acts on the
+cloud-borne tracers rather than the interstitial ones.
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in `4ed73672` — the page did contradict itself, and that is on my merge: I kept both sides' subsections without reconciling the one claim they disagreed about.

**One correction to the suggested remedy.** I revised the limitation rather than removing it, because #781's gap is not settled by this change. Its concern is that the *convective* carrier has no area of its own while `ConvectiveTracerTransport` already removes the same air in-plume — "the aerosol inside the plume is being removed twice under two different area assumptions". Dropping `clear_fraction` removes the *stratiform* weighting; it does not give the convective carrier a cover, so the double-exposure is unchanged (if anything it is now unconditional rather than `≈1` only in low-stratiform columns). #781 is open and, on my reading of its body, still live.

The text now reads:

> WetScavenging applies **no cloud cover** to either below-cloud carrier: CAM's swept precipitating volume cancels against the in-precip-area rain rate, so below_cloud_rate acts on the grid-mean interstitial mixing ratio. The convective carrier therefore still has no area of its own, and the full grid-mean aerosol is exposed to convective impaction through the depth of the convective cloud as well as below its base, on top of the in-plume sink — the same air removed twice under two different area assumptions. The references partition by the cover belonging to the carrier. Tracked in #781.

That matches `below_cloud_rate`'s signature (no `cloud_fraction` argument) and agrees with the removal subsection lower down. `grep clear_fraction docs/source/science/aerosol.md` is now empty.

If you read #781 as closed by this, say so and I will drop the bullet and the citation instead — but I did not want to retire an open issue on my own reading.

duncanwp:
@codex review

Issue #791, opened 2026-09-10
JAM optics: NeuralMie neural-network bulk-optics backend (sphere + core-shell), opt-in


aliakherati:
### Summary

Proposal: add **NeuralMie** (Geiss & Ma, GMD 2024,
[doi:10.5194/gmd-2024-30](https://doi.org/10.5194/gmd-2024-30)) as an opt-in
aerosol-optics backend for `jam/optics`, alongside the existing
Bohren–Huffman LUT. It replaces the 8-node Gauss–Hermite quadrature with a
single neural-network forward pass that predicts the *mode-integrated* bulk
optics directly, and it adds a core-shell capability the current volume-mixing
cannot represent.

The default would stay `mie_lut`, so nothing changes unless asked for.

Upstream: [`pnnl/NEURALMIE`](https://github.com/pnnl/NEURALMIE), BSD-2-Clause.
JAX port (inference + a NumPy Mie ground truth + validation):
[`reflective-org/neuralmie-jax`](https://github.com/reflective-org/neuralmie-jax).

### What it is

Two small pretrained MLPs, `swish` hidden layers, linear 3-output head:

| network | inputs | architecture | parameters |
|---|---|---|---|
| `sphere` | 4 | 4 → 69×4 → 3 | 15,045 |
| `coreshell` | 7 | 7 → 112×4 → 3 | 39,203 |

Inputs are `(2πr_g/λ, σ_g, m_r, m_i)` plus, for core-shell, the core index and
`f = r_core/r_total`. The output is `ln(ke_ρ·λ)`, `logit(ssa)`, `logit(g)`,
where `ke_ρ = 0.75·∫Qe·pdf·r² dlnr / ∫pdf·r³ dlnr` is the extinction
cross-section per unit particle volume.

### The seam

`optics_term.py:294-322` currently runs, per mode and band, an 8-node
Gauss–Hermite quadrature over the lognormal in `ln r` inside a `lax.scan`,
calling `interp_mie` per node. NeuralMie replaces that whole construct — the
quadrature *and* the LUT lookup — with one forward pass.

Worth being explicit, since it is the obvious wrong guess: NeuralMie is **not**
a drop-in for `interp_mie`'s per-particle `(x, mr, mi) → (q_ext, ssa, g)`
signature. It predicts the mode-integrated result, so it sits one level up.

### The two backends compute the same quantity

This is what makes a direct A/B a correctness test rather than a
[...]

Issue #790, opened 2026-09-10
JAM optics: hygroscopic water volume inconsistent with the third moment defining r_dry (understated by exp(4.5 ln²σ))


aliakherati:
### Summary

`JamOpticsTerm`'s hygroscopic water volume is inconsistent with the third
moment that `PlaceholderMicrophysics` used to define `r_dry`, so `v_water` is
understated by `exp(4.5 ln²σ_g)` — a factor of **2.70** for the Aitken mode
(σ=1.6) and **4.73** for accumulation / coarse / primary-carbon (σ=1.8).

This affects the current default optics path, via the volume-mixed refractive
index.

### Evidence

`placeholder.py:92-101` states the relation it inverts:

```python
# Number-median dry diameter from total volume of a log-normal mode:
#   V = N (π/6) Dg³ exp(9/2 ln²σ)   →   Dg = (V / (N k))^(1/3)
ln_sigma = jnp.log(mode.geom_std_dev)
k = (jnp.pi / 6.0) * jnp.exp(4.5 * ln_sigma ** 2)
dg_cubed = total_vol / (number * k)
...
r_dry = 0.5 * dg
```

Since `Dg = 2·r_dry`, that relation gives

```
N·(4/3)π·r_dry³  =  N·(π/6)·Dg³  =  total_vol / exp(4.5 ln²σ)
```

i.e. `N·(4/3)π·r_dry³` is deliberately *not* the mode's dry volume — it is the
dry volume divided by `exp(4.5 ln²σ)`.

But `optics_term.py:283-285` builds the water volume from exactly that product:

```python
v_water = aer.number[i] * _FOUR_THIRDS_PI * jnp.maximum(
    r_wet ** 3 - aer.r_dry[i] ** 3, 0.0
)
```

Because `r_wet = r_dry · growth` (`placeholder.py:107`), this is

```
v_water = [total_vol / exp(4.5 ln²σ)] · (growth³ − 1)
```

whereas the third-moment-consistent value is `vol_dry · (growth³ − 1)`. The
`vol_dry` accumulated just above at `optics_term.py:274-282` *is* the correct
third moment (`Σ mass/density`), so the two are mixed inconsistently in
`vol_tot`.

### Consequence

`vol_tot` sets the volume-mixed refractive index at `optics_term.py:291-293`,
so the water fraction of each mode is understated:

| σ_g | r_wet/r_dry | correct water fraction | as computed | ratio |
|---|---|---|---|---|
[...]

PR #792, opened 2026-09-10
JAM optics: opt-in NeuralMie neural-network bulk-optics backend (sphere + core-shell)

aliakherati:
## Describe your changes

Adds **NeuralMie** (Geiss & Ma, GMD 2024, [doi:10.5194/gmd-2024-30](https://doi.org/10.5194/gmd-2024-30)) as an **opt-in** aerosol-optics backend for `jam/optics`, selected with `jam_optics_backend` or `physics=echam-jam-neuralmie`. The default stays `mie_lut` and the existing quadrature body is byte-identical under the new branch, so **default answers are untouched** (verified: the 36 pre-existing optics tests pass unchanged).

Instead of an 8-node Gauss–Hermite quadrature over a 64×24×24 trilinear Bohren–Huffman table, one forward pass through a pretrained emulator (15,045 / 39,203 parameters) predicts the mode-integrated bulk optics directly.

### Why this approach

**The two backends compute the same integral**, which is what makes the A/B a correctness test rather than a characterisation test. With `⟨·⟩` the number-weighted lognormal moment and `n_A` the column number per area:

```
LUT       : τ = n_A·sec·π r_g²,  sec = ⟨Qe·(r/r_g)²⟩   = π·n_A·⟨Qe r²⟩
NeuralMie : τ = ke_ρ·V_A,        V_A = n_A·(4/3)π⟨r³⟩  = π·n_A·⟨Qe r²⟩
```

The LUT discretises it with 8 nodes over a coarse table; NeuralMie reproduces the 1024-point quadrature it was trained on. Forcing the homogeneous network everywhere, the two agree to **0.4% on ssa and 0.2% on g**; the residual ~5% on optical depth is the LUT's own discretisation error — its interpolation tolerance in `mie_test.py` is 15% on `q_ext` off-grid.

**Vendored rather than added as a dependency.** ~350 lines under BSD-2-Clause, which is inbound-compatible with Apache-2.0, so unlike `mam4-jax`'s GPL it needs no optional-extra quarantine — and code behind an extra counts as uncovered in CI anyway. A VCS requirement is what `requirements.txt` already warns against. **Happy to switch to a dependency if you'd prefer** — see the question in #791; that choice is yours and doesn't affect `neuralmie.py` itself.

Refs #791 (proposal and design discussion). Related: #790, the water-volume inconsistency this surfaced.

---

## Two physics changes — please read before approving

This is not a pure numerical swap.

**1. Core-shell mixing for BC-bearing modes.** BC as a concentric core, everything else including aerosol water as the coating, `f = clip((V_bc/V_tot)^(1/3), 0, 0.98)`. This is a modelling *choice*: MAM4/E3SM and ECHAM-HAM both volume-mix.

The direction is easy to invert, and **I initially got it wrong in #791 and corrected it there.** Averaging the refractive index smears BC's large imaginary part over the whole particle, which *over*-absorbs relative to confining it to a core. So these modes become **less** absorbing. Verified against exact TAMie + 1024-point quadrature at 550 nm (r_g = 100 nm, σ_g = 1.8, BC in sulfate), ssa core-shell vs volume-mixed:

| V_bc/V | core-shell | volume-mixed |
|---|---|---|
| 0.008 | 0.9676 | 0.9661 |
| 0.064 | 0.8182 | 0.7887 |
| 0.125 | 0.7001 | 0.6710 |
| 0.512 | 0.4323 | 0.4585 |

The sign reverses only near `V_bc/V ≈ 0.5`, far above realistic loadings. The "coating enhances BC absorption 1.2–2×" result compares coated BC to *bare* BC, which is not this comparison. **This moves ERFari and this PR does not retune for it.**

Applying core-shell to **primary carbon** is the weakest part: `_PCARBON_SPECIES` is `soluble=False`, "hydrophobic until it ages", exactly where a concentric coated sphere is least physical, and it is the mode most likely to reach the `f` cap. I applied it to all BC-bearing modes for uniformity; restricting to `soluble=True` is a one-line change if you prefer it.

**2. Number-free hygroscopic water volume.** `vol_dry·((r_wet/r_dry)³ − 1)` instead of `N·(4/3)π·(r_wet³ − r_dry³)`. `PlaceholderMicrophysics` defines `r_dry` from `V = N·(π/6)·Dg³·exp(4.5 ln²σ)`, so `N·(4/3)π·r_dry³` is the dry volume *divided by* that factor — the original understates water by 2.70× (Aitken) to 4.73× (σ=1.8 modes). Applied **only on the NeuralMie path** here; the default path is #790 because fixing it there moves default answers and deserves its own validation.

## How it was verified

Local gate, `ruff==0.15.17` clean, and `jcm/physics/aerosol` + `jcm/physics/echam`: **404 passed, 2 skipped**. The optics directory goes 36 → 75 tests.

**Coverage could not be measured locally.** `pytest --cov` aborts with `SIGABRT` from absl inside jaxlib 0.11.1 on this machine (macOS arm64) — it reproduces on untouched modules such as `mie_test.py`, so it is environmental, not from this branch. **CI will have to confirm the gate.** I covered the new branches deliberately instead: both backends, both Rayleigh arms, the `shell_ok=False` pure-core fallback (driven by zeroing non-BC tracers, since the MAM4 fixture never reaches it), domain clipping, and both config validation raises.

### Measured cost

CPU only — **this is not the T63L47 GPU number the workflow asks for**, and I don't have the device. `_compute_fields`, 14 SW + 16 LW bands, MAM4 4 modes:

| cells | backend | peak scratch | time |
|---|---|---|---|
| 37,600 | `mie_lut` | 234 MiB | 164 ms |
| 37,600 | `neuralmie` | **133 MiB** | 422 ms |
| 150,400 | `mie_lut` | 936 MiB | 709 ms |
| 150,400 | `neuralmie` | **530 MiB** | 1597 ms |

So on CPU it is **~2.3× slower but uses ~43% less memory**. I would not extrapolate the timing to GPU: this trades 8 gather-bound `map_coordinates` lookups for a dense matmul, which is the access pattern GPUs prefer, but the direction is genuinely unknown until someone runs `jcm-benchmark`.

**A memory regression I caught and fixed before shipping**, worth flagging since it would have bitten at production resolution: `vmap`ping 30 bands materialises `(band, mode, cell, 69)` activations per layer — **13.7 GiB** of scratch at 150k cells, extrapolating past an 80 GB device at T63L47. `jax.checkpoint` does not help, because this is the *forward* peak rather than rematerialisation for the backward pass. Running bands through `lax.map` brings it to 530 MiB for an identical result — the same reasoning that already puts the Gauss–Hermite nodes in a `lax.scan`.

## What I'd look at closely

1. **The primary-carbon decision** above — the one place I'd most expect you to disagree.
2. **`g = sigmoid(·) ∈ (0,1)`**, so this backend cannot represent back-scattering, unlike the LUT path clipped to `[-1, 1]`. Bounds how tightly the backends can agree.
3. **Per-species apportionment is no longer exact** under core-shell: the weights are justified by volume-mixing into one effective index, which no longer holds. It remains a defensible apportionment — and the only one an internally-mixed model can report — but the claim of exactness is gone. Noted in the design doc.
4. **The accuracy figures cap the size parameter at 100** to keep the Mie reference affordable, so they do not characterise the large-`x` corner of the emulator's domain (which reaches ~1570). Treat the maxima as lower bounds.
5. Whether the diagnostic pass switching backends with the radiative optics is what you want. I made it so deliberately — branching at `_band_optics` instead would let `od550aer` describe a different optical model from the one radiation used, and the existing closure tests would not catch it because both sides would stay internally consistent.


aliakherati:
**CI needs a maintainer to approve the workflows.** Both `Run Unit Tests via Pytest` and `Claude Code Review` are sitting at `action_required` — GitHub gates workflows on a first-time contributor's fork PR — so no checks have reported and none will until someone clicks *Approve and run workflows*. Nothing to fix on my side; flagging it so the empty check list isn't read as a failure.

Note also that `run_linter.yaml` triggers on `[push]` only, so ruff won't run for a fork PR at all even after approval. It is clean locally against the CI-pinned `ruff==0.15.17`.

What I ran locally, for reference until CI reports:

- `ruff check .` with `ruff==0.15.17` — clean
- `JAX_PLATFORMS=cpu pytest jcm/physics/aerosol jcm/physics/echam -m "not slow"` — **404 passed, 2 skipped**
- `jcm/physics/aerosol/jam/optics` goes 36 → 75 tests; the 36 pre-existing ones pass unchanged, and the LUT quadrature body is byte-identical under the new branch (verified with `diff` modulo indentation)

**The one gate I could not clear locally is coverage.** `pytest --cov` aborts with `SIGABRT` from absl inside jaxlib 0.11.1 on this machine (macOS arm64), and it reproduces on untouched modules such as `mie_test.py`, so it is environmental rather than from this branch. I covered the new branches deliberately instead — both backends, both Rayleigh arms, the `shell_ok=False` pure-core fallback (driven by zeroing non-BC tracers, since the MAM4 fixture never reaches it), domain clipping, and both config-validation raises — but I can't put a number on it, and I would rather say so than imply I cleared a gate I didn't.

duncanwp:
@codex review

duncanwp:
Thank you for this - it looks great . 

I think, rather than reproducing the original code directly in JCM though, I'd rather make a local interface only that makes it easy to drop in neural_mie from the upstream package. There are already a couple of small issues in the code that would need fixing both here and there, and they will inevitably drift. Depending on it isn't the answer either: it isn't on PyPI, and requirements.txt rules out VCS requirements. So neither vendor nor depend — a documented extension point, with the backend installed and attached by whoever wants it:

```
from jcm.physics.echam.echam_terms import echam_physics
from neuralmie_jax.jcm import NeuralMieOpticsTerm   # subclasses JamOpticsTerm

physics = echam_physics(aerosol_module="jam").replace(
    "aerosol_optics", NeuralMieOpticsTerm())
```

I've opened #835  with the jax-gcm half, with you as co-author. It takes the plumbing this PR had to invent and leaves out only what is NeuralMie's own:

_mode_optics(ModeOpticsInputs) -> (tau, tau_scat, tau_scat_g) — your per-mode branch, generalised. The record carries both normalisations, so a per-particle-volume backend gets vol_total and the col_factor you threaded through, and a cross-section backend gets num_per_area.
_map_bands — your lax.map finding, as a hook, since whether the band axis fits in memory is a property of the backend and not of the band loop. (One correction to the PR text: the two band maps agree to rounding, not identically — XLA associates a batched and a scanned axis differently, ~2 ulp in float32.)
_build_mie_lut — so a backend that never reads the table doesn't pay the 4 s build.
The base class keeps the mass gate, the SSA/asymmetry weighting, the AeroCom apportionment and the column diagnostics, so a backend implements one method.

The refactor is bit-for-bit identical on all 57 output fields of the default path, diagnostics on and off.

Two notes on the rest:

Your water-volume fix landed independently. [#790](https://github.com/climate-analytics-lab/jax-gcm/issues/790) was fixed in [#821](https://github.com/climate-analytics-lab/jax-gcm/issues/821) and is on dev now, for both paths rather than only the NeuralMie one — so that half of "two physics changes" is already yours and already in.

Of the four review issues, two are real and belong upstream, where I've filed them: [reflective-org/neuralmie-jax#4](https://github.com/reflective-org/neuralmie-jax/issues/4) (the reference truncates both integrals to the 99.9 %-of-number interval, which keeps only 93.7 % of the volume at σg = 1.8 — so ke_rho is a truncated-moment ratio and a consumer pairing it with a full third moment picks up −0.6 % to +9.8 % across σg 1.4–2.0; measured with the reference against itself) and [reflective-org/neuralmie-jax#5](https://github.com/reflective-org/neuralmie-jax/issues/5) (the README's f = (V_core/V_tot)**(1/3) has an infinite derivative at zero core volume, and a downstream where guard turns it into NaN — that is #793's root, and it wants a guarded helper upstream rather than a fix in each consumer). [#795](https://github.com/climate-analytics-lab/jax-gcm/issues/795) is moot once nothing is vendored. [#796](https://github.com/climate-analytics-lab/jax-gcm/issues/796) stands as a caveat for whoever writes the backend: an absorption claim needs AAOD, not SSA — I've carried that, the truncation caveat and the g ∈ (0,1) limitation into the seam's design note.

Issue #787, opened 2026-09-09
Dust emission's soil-moisture gate runs on SPEEDY's root-zone index, not ECHAM's ws/wsmx


duncanwp:
> **Re-scoped 2026-09-13.** This issue was first written against the CAM/CLM
> dust path in PR #783, which was superseded by the HAMMOZ/Tegen port (#802,
> merged as #804). `wet_threshold_factor` and `soil_water_gwc_scale` no longer
> exist. The underlying gap is unchanged and is now on an *unconditional* code
> path, so the issue stays open; the text below describes the code as merged.

## What is missing

`DustEmissions` uses soil moisture in two places, and both want ECHAM's
`ws/wsmx` — soil water relative to field capacity — which jcm does not carry.

**1. The saturation cut-off, unconditional in every preset**
(`jcm/physics/aerosol/jam/emissions/dust.py`, the `scale` expression):

```python
scale = jnp.where(mask & (wetness <= p.w0), 10.0 * (1.0 - snow) * pot, 0.0)
```

`w0 = 0.99` mirrors `mo_ham_dust.f90`'s `ws/wsmx > 0.99 → no flux`. What is
passed is `forcing.soilw_am`, SPEEDY's **vegetation-weighted root-zone
availability index** (`jcm/data/bc/compile.py`):

```python
rsw   = 1/(swcap + 3*(swcap - swwil))
soilw = min(1, rsw*(swl1 + veg*3*max(0, swl2 - swwil)))
```

**2. The Fécan et al. (1999) threshold correction**, behind
`DustParameters.fecan_moisture` (`k_dust_smst`). Enabled only by `ndust = 2`
(Cheng 2008); the shipped default `ndust = 4` leaves it off, so this path is
dormant today.

## Why it matters

Three mismatches, all pushing the same way — over-suppression of marginal
sources:

1. **It saturates at field capacity, not at porosity.** `soilw_am = 1` already
   means field capacity, so the index compresses the range the cut-off is
   defined against.
2. **It blends the 7-28 cm layer**, which does not dry on the timescale that
   governs saltation; the surface can be emitting while the root zone is not.
3. **It is already vegetation-weighted** (`veg = cvh + 0.8·cvl`), which
   double-counts against the potential-source field, whose `laieff` provenance
   carries vegetation seasonality of its own.

**Current impact is small but unquantified.** Because `soilw_am` is capped at
field capacity by construction it rarely approaches `0.99`, so the cut-off
[...]

PR #783, opened 2026-09-09
Dust source gating, CAM emission split/sizes, and a real 10 m emission wind (#768, #723)

duncanwp:
Closes #768, closes #723. Splits out #777 (the one gate that needs a new boundary field).

Five separable defects on the **emission** side of JAM — three in dust, one in the
emitted-number conversion for every primary species, one in the wind the
surface-flux schemes read. The removal side (#722/#776) is untouched, and both
validation arms sit on the same base (`origin/dev` @ `f5c8b354`) so they are not
confounded by it.

### #768.1 — the dust source map was ungated (`emissions/dust.py`)

`forcing.dust_source` is CAM's geomorphic basin factor `mbl_bsn_fct_geo`. CAM
(`dust_model.F90`) zeroes it below `soil_erod_threshold = 0.1`, applies **no upper
bound** (it runs to 5.7 in the big closed basins), and multiplies by CLM's
`dust_flux_in`, which already carries the land, snow, soil-moisture and vegetation
masks. jcm clipped it to `[0, 1]` and applied *no* mask — `terrain` was passed to
`__call__` and never used. Hence 9.8 % of global dust from south of 60 S and the
Amazon/Congo out-weighing the Sahara.

The term now applies, per column, the gates CLM applies (`DustEmisZender2003.F90`):

| factor | reference | field |
| --- | --- | --- |
| erodibility, zeroed below 0.1, unbounded above | CAM `soil_erod_threshold` | `forcing.dust_source` |
| land fraction | CLM runs on land columns | `terrain.fmask` |
| snow-free fraction | CLM `lnd_frc_mbl` | `forcing.snowc_am` |
| unfrozen fraction (ramp over the 2 K below `tmelt`) | CLM `liqfrac` | `forcing.stl_am` |
| `u*t · √(1 + 1.21·(100·(w − w_thr))^0.68)` | Fecan (1999) = CLM `frc_thr_wet_fct` | `forcing.soilw_am` |

The saltation flux itself already matched: jcm's `(1+r)(1−r²)` is White's
`(1−r)(1+r)²`, the same form CLM uses.

### #768.2 — mode split

`_MAM4_PRIMARY_EMISSION["du"]` was `(acc 0.1, cor 0.9)`; CAM's `dust_emis_sclfctr`
over the 0.1–1 µm and 1–10 µm bins is `(0.021, 0.979)` — **4.8× too much
accumulation-mode dust**. (CAM's third, 1.65e-5 Aitken share has no home in a
[...]

Issue #777, opened 2026-09-09
No vegetation/LAI boundary field: dust emission cannot apply CLM's bare-ground mobilization gate


duncanwp:
## What is missing

`DustEmissions` (`jcm/physics/aerosol/jam/emissions/dust.py`) now applies the
CAM/CLM surface gates it can build from the boundary conditions the model
carries — land fraction, snow cover, frozen ground, and the Fecan (1999)
soil-moisture increase of the threshold friction velocity. The one CLM gate it
cannot apply is the **vegetation** one:

```fortran
! CTSM DustEmisZender2003.F90
if (tlai_lu(l) < vai_mbl_thr) then          ! vai_mbl_thr = 0.3 m2/m2
   lnd_frc_mbl(p) = 1.0_r8 - tlai_lu(l)/vai_mbl_thr
else
   lnd_frc_mbl(p) = 0.0_r8
endif
lnd_frc_mbl(p) = lnd_frc_mbl(p) * (1.0_r8 - frac_sno(c))
```

CAM's `dust_flux_in` arrives from CLM already carrying this factor. In jcm
there is no LAI, VAI, PFT-cover or land-cover field on `ForcingData` or
`TerrainData` at all (`jcm/forcing.py:230-330`, `jcm/terrain.py:278-311`), so
vegetated cells are gated only by their soil moisture.

## Why it matters

CAM's geomorphic basin factor `mbl_bsn_fct_geo` — the erodibility map the
mirror bundle ships as `bundles/<grid>/dust.nc` — is a *topographic* measure
with no aridity information: the Amazon (7.2 % of the global source weight,
peak 4.40) and Congo (3.5 %, peak 2.16) basins carry larger values than the
Sahara or Taklamakan. In CAM they emit nothing because CLM's LAI gate zeroes
them; here the Fecan moisture factor only raises their threshold friction
velocity (~3.4x at 0.95 relative wetness), so a residual tropical-basin source
survives.

## Where the hooks already are

- `mobilization_fraction()` in `jcm/physics/aerosol/jam/emissions/dust.py` is
  the CLM `lnd_frc_mbl` analogue and is where a `(1 - VAI/0.3)` factor slots in.
- A monthly VAI product would also enable `physics.jam_dust_source=tegen_potential`
  with a genuine vegetation-seasonal potential-source map (the HAMMOZ
  `dust_potential_sources_T63.nc` layout `read_dust_source` already supports —
[...]

Issue #673, opened 2026-08-19
Undeclared diagnostics.get() reads silently return previous-step data


duncanwp:
Found by a re-audit of the July 2026 ECHAM+RRTMGP review against `dev` @ `f6d1bcd`. Parent: #614.

## What's wrong

`composable_physics.py:247` and `:286` seed each step's diagnostics dict with `{**prev_physics_data}` — the entire previous step's dict is visible to every term from the first term onward. `_validate_ordering` (`:707-722`) checks only *declared* `requires` against upstream `provides`. So any undeclared read silently returns t−1 data.

`requires_audit_test.py` was added since the review and is real hardening: it AST-walks each `PhysicsTerm.__call__` and asserts every bare `diagnostics["key"]` subscript is declared. **But its own docstring exempts the other form:**

> Reads via `diagnostics.get("key")` or behind an `if "key" in diagnostics` guard are optional-by-convention and **exempt**

and `physics_term.py:77-88` *mandates* that optional dependencies be written as `.get()`. **The enforcement mechanism channels every undeclared read into precisely the form nothing checks.** The hole is narrower than before but sharper.

## A live instance, currently shipping

`sundqvist.py:712-717`:
```python
conv = diagnostics.get("convection")
no_convection = (jnp.reshape(conv.ktype, (-1,)) == 0
                 if conv is not None and hasattr(conv, "ktype")
                 else jnp.ones_like(is_ocean, dtype=bool))
```
with the comment at `:702-705` asserting `ktype` comes from the convection diagnostics "when a convection term ran earlier in the step". In the default factory (`echam_terms.py:452-470`) `SundqvistCloudFraction` is index **4** and `TiedtkeConvection` is index **9** — convection runs *later*. The stratocumulus-enhancement gate therefore reads the **previous step's** `ktype` on every step, and the comment states the opposite. Nothing in CI catches it, because the read is a `.get()`.

## The complication a fix has to handle

A blanket "undeclared reads fail" would break the **legitimate** cross-step carry pattern, which uses the same syntax deliberately:
- `surface_physics.py:523` — `diagnostics.get("surface", SurfaceData.zeros(...))`
- `sundqvist.py:745` — `diagnostics.get("clouds", ...)`

The distinction the check needs is **carry-slot vs stale data read**, not declared vs undeclared.

## Fix

[...]

PR #803, opened 2026-09-11
Emit sea salt and DMS from a diagnosed 10 m wind (#723)

### LINKED ISSUE #723: Surface-flux emissions use lowest-model-level wind, not a diagnosed 10 m wind (+35–40% sea-salt flux at L47)
`jam/emissions/seasalt.py:159` and `jam/emissions/dms.py:94` compute the emission wind as `sqrt(u[-1]² + v[-1]²)` — the lowest full model level (~30 m at L47, σ≈0.996). The reference schemes are calibrated to **u10** (HAMMOZ passes `vphysc%velo10m`, `mo_vphysc.f90:54`). Over ocean u(30 m)/u(10 m) ≈ 1.1, and the Gong `u^3.41` dependence amplifies that to ≈ +35–40% sea-salt flux; DMS (Nightingale-type k ~ u²) gets ≈ +20%.

Fix: diagnose 10 m wind from the surface layer's stability functions — the ECHAM surface scheme (`surface/echam/`) already computes the transfer coefficients needed for the standard log-profile reduction; alternatively a neutral-log reduction with z0 from the surface tiles as a first version. Publish it as a diagnostic so all surface-flux parameterizations (sea salt, DMS, dust threshold winds if applicable) consume the same field.

Found during the sea-salt emission audit (see the sea-salt burden issue); at L47 this bias partially masks the removal-side deficit documented there.

duncanwp:
Closes #723. Split from #783, which is being divided: the dust source is redone
as a HAMMOZ/Tegen port under #802, and this is the 10 m emission wind and its
tooling.

## What was wrong

Gong (2003) sea salt (`u10**3.41`) and Nightingale (2000) DMS (`k_w ~ u10²`)
are calibrated to the **10 m** wind — HAMMOZ passes `vphysc%velo10m`
(`mo_vphysc.f90`) — but both read `state.u_wind[-1]`, the lowest full level,
~33 m at L47. Over the ocean that is ≈1.10× u10, which the exponents amplify to
+37-46 % on sea salt and +20-25 % on DMS.

## What this does

`TteTkeVerticalDiffusion` publishes the ECHAM5 `vdiff.f90` / ICON
`mo_surface_diag::nsurf_diag` reduction as `VerticalDiffusionData.wind_10m`:

```
bn   = ln(z1/z0m)                              neutral profile factor
bm   = bn·√(CM_n|U| / CM|U|)                   stability-corrected
zrat = 10/z1
red  = [ln(1 + (e^bn − 1)·zrat) + merge] / bm
merge = −(bn − bm)·zrat                        stable   (CM|U| < CM_n|U|)
      = −ln(1 + (e^(bn−bm) − 1)·zrat)          unstable
```

Three properties worth the reviewer's attention:

* **Computed inside each scheme's `lax.cond` branch**, from that scheme's own
  neutral drag. ECHAM-Louis and Businger-Dyer differ in roughness (state-carried
  vs a hard-coded table), in the bound on `z/z0`, and in whether the wind is
  `zepdu2`-floored; both branches now build their momentum drag *from* the same
  helper the reduction uses, so the pair cannot drift. A shared reference read
  stable at exact neutral over ocean (0.863 vs 0.900 — −13 % on the sea-salt
  wind term).
* **The `zepdu2` floor is respected**: the profile factor uses the
  `max(|U|, 1 m/s)` speed the coefficients were built from, or the calm-wind
  floor reads as a stability signal (0.9931 against the correct 0.9038 at
  |U| = 0.5 m/s). The resulting reduction multiplies the true wind.
* **The stable/unstable branch** is selected by `CM|U| < CM_n|U|` — exactly
  `Ri > 0` for both schemes, continuous at `Ri = 0` — so the Richardson number
  stays inside the coefficient solve.

Emissions run before vdiff in the ECHAM ordering, so the value is the previous
step's: `vertical_diffusion` is a **declared** cross-step carry slot
(`initial_carry_state`), the same one-step lag `DustEmissions` already takes for
`u*`, and the deliberate pattern #673 distinguishes from an undeclared stale
read. On step 1 of a cold start the slot is still zero-filled, so both schemes
fall back to the lowest level and publish a per-column `wind_10m_model_level`
flag that `ResetEmissionFluxes` zeroes each step. `check_health` reports the
chunk mean and fails only a **persistent** fallback (`chunk_idx > 0 and
frac > 0.5`): under interval averaging the legitimate bootstrap step and one
defective step mid-chunk give the same `1/N`, so the per-step invariant is
asserted in the unit tests instead.

The ECHAM surface term's own "10 m wind" diagnostic was the same defect
relabelled and now takes the diagnosed value.

## Validation

30 days at T63L47, `+configuration=ma-t63-l47`, identical dates. **These numbers
come from the combined branch (#783), which also changed dust**; the sea-salt
and DMS figures are unaffected by the dust half, and the dust numbers are
deliberately not quoted here.

| quantity | before | after |
| --- | ---: | ---: |
| sea-salt emission | 1130 Tg/yr | **768.9 Tg/yr** (−32 %) |
| DMS flux | 38.8 Tg/yr | **32.5 Tg/yr** (−17 %) |

Emission-wind provenance per chunk (`wind_10m_model_level`):

| chunk | 0 | 1 | 2 | 3 | 4 | 5 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| day | 5 | 10 | 15 | 20 | 25 | 30 |
| model-level | **0.208 %** | 0.000 % | 0.000 % | 0.000 % | 0.000 % | 0.000 % |

0.208 % is exactly 1/480 — one step of a 5-day chunk at dt = 15 min — and zero
thereafter: the bootstrap-only property measured end to end.

## Gates

`ruff check .` clean. Gates passed in a worktree detached at `ea069911`:

| gate | result | coverage |
| --- | --- | --- |
| fast (≥90 %) | 2453 passed, 28 skipped | **95.24 %** |
| slow (≥80 %) | 106 passed, 17 skipped | **84.36 %** |

`docs/science_pointers_test.py` green (9 passed).

## Also here

* **The emission-size mechanism** (`distributors.py`): `particle_mean_mass`
  takes an optional volume-mean diameter of the *emitted* distribution, since
  emitted number scales as D⁻³ and fresh particles are not at their class's
  equilibrium size. **Mechanism only** — every existing caller keeps the class
  geometry. #783 populated it for BC, POA and primary sulfate from the CESM
  emission files; that is a CAM number inside a HAMMOZ package, the same class
  of substitution the dust split rejects, so those species keep `dev`'s
  behaviour here and take HAM's own sizes (Stier et al. 2005) with the
  Tegen/HAM emissions port.
* **Delegated-timestep guards**: `run=pyses_year` sets `time_step: null`
  because the dycore owns the timestep, so anything in the per-chunk path that
  reads it as a number crashes after the integration and before the checkpoint.
  A fast-lane dinosaur test (13 s) and a slow end-to-end pySES test cover it;
  both were verified to fail on a reintroduced `float(None)`.
* Science-doc entries per the #771 rule: the 10 m diagnostic in
  `vertical_diffusion.md`, the sea-salt/DMS emission wind in `aerosol.md`. Dust
  entries belong to #802.

## Not here

Everything dust: the source map and its gating, the CAM erodibility product and
its mirror publication, the `jam_dust_source` option, the mode split and dust
emission sizes. #802 supersedes that work with a HAMMOZ/Tegen port on the
supplied HAMMOZ inputs. #777 and #787 stay open as artefacts of the superseded
approach.

---

### DIFF
--- a/docs/design/jam.md
+++ b/docs/design/jam.md
@@ -234,6 +234,87 @@ See `.claude/aerosol_emissions_plan.md` for the full design, the data-source
 investigation (the raw 0.5° gridded CEDS is ESGF-only; a self-hosted compressed
 mirror is a tracked follow-up), and the CESM adapter's documented approximations.
 
+## Natural emissions: the 10 m wind (#723)
+
+### The emission wind is 10 m, not the lowest model level
+
+Gong sea salt (`u10**3.41`) and Nightingale DMS (`k_w ~ u10**2`) are fitted to
+the 10 m wind; HAMMOZ passes `vphysc%velo10m`. Reading the lowest full level
+instead — ~33 m at L47 — inflates sea salt by 37-46 % and DMS by 20-25 %.
+
+`TteTkeVerticalDiffusion` now publishes `VerticalDiffusionData.wind_10m`, the
+ECHAM/ICON `nsurf_diag` reduction of the lowest-level wind along the same
+surface-layer profile that produced the drag:
+
+```
+bn  = ln(z1/z0m)                              neutral profile factor
+bm  = bn * sqrt(CM_n|U| / CM|U|)              stability-corrected
+red = [ln(1 + (e^bn - 1)*10/z1) + merge] / bm
+merge = -(bn - bm)*10/z1                      stable   (CM|U| < CM_n|U|)
+      = -ln(1 + (e^(bn-bm) - 1)*10/z1)        unstable
+```
+
+Building it from the per-tile `CM·|U|` the surface stress already uses means the
+10 m wind cannot drift from the momentum exchange, and the stable/unstable
+branch needs no separate Richardson number (the two branches meet continuously
+at `Ri = 0`, where `CM|U| = CM_n|U|`). The profile factor is built from the
+`zepdu2`-floored speed the coefficients themselves used (`max(|U|, 1 m/s)`),
+or ECHAM's calm-wind floor would be misread as a stability signal; the
+reduction it yields then multiplies the true wind.
+
+Emission terms read the result through `emissions/surface_wind.py`. They run
+*before* vdiff in the ECHAM ordering, so the value is one step old. This is
+the **declared** cross-step carry — `vertical_diffusion` is one of the slots
+`Physics.initial_carry_state` seeds — i.e. the deliberate pattern #673
+distinguishes from an accidental stale `.get()`, and the same lag the dust
+term's `u*` already carries. When #673's `carry_slots` check lands this read
+is one to declare explicitly.
+
+**The fallback to the lowest model level is bootstrap-only, by construction.**
+The 10 m wind *cannot* exist on step 1 of a cold start: emissions run first,
+and the `vertical_diffusion` carry slot is seeded zero-filled
+(`initial_carry_state`; verified — `surface_friction_velocity` and `wind_10m`
+are both exactly 0 there), so no surface layer has been diagnosed yet.
+(A resumed run carries a real 10 m wind and never takes the fallback; nor does
+a composition with no vdiff term, where there is no surface layer at all.)
+Because taking it silently would mean emitting 37-46 % too much sea salt, the
+emission terms publish a per-column flag `wind_10m_model_level` — 1 where the
+model level was used, 0 where the diagnosed wind was — zeroed every step with
+the other emission diagnostics. It is 1 on step 1 of a cold start and 0
+thereafter; `seasalt_test.py` (`test_fallback_is_taken_on_step_one_and_never_again`
+and its companions) asserts exactly that per step, against the zero-filled carry
+a cold start begins with and a populated one thereafter. The JAM integration
+test adds the end-to-end check that no column is still flagged after six steps.
+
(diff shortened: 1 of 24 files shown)


duncanwp:
@codex review

duncanwp:
Note for the 30-day T63L47 table in this PR (the -32 % sea salt / -17 % DMS numbers): that run is a **cold start from `init=jw`**, so its circulation is a spin-up, not a climatology. The numbers are a valid before/after contrast — both members start from the identical unspun state and differ only in the emission wind — but they are not an estimate of the model's sea-salt or DMS source, and should not be read against observational totals.

Surfaced while validating #804 (Tegen dust), where the same setup gave a global dust emission 35x below expectation: over the cells that pass the January vegetation gate the mean 10 m wind is 3.19 m/s against a 7.6 m/s saltation onset, because 30 days from JW has no spun-up Harmattan. Sea salt and DMS are smooth in `u10` so they degrade gracefully where dust, being threshold-gated, collapses — but the underlying wind field is the same cold one.

PR #821, opened 2026-09-15
JAM optics: third-moment-consistent hygroscopic water volume (#790)

### LINKED ISSUE #790: JAM optics: hygroscopic water volume inconsistent with the third moment defining r_dry (understated by exp(4.5 ln²σ))

duncanwp:
## What changed

`JamOpticsTerm` built each mode's hygroscopic water volume as
`N·(4/3)π·(r_wet³ − r_dry³)` and mixed it, in the volume-mixed refractive index,
with a dry volume summed as `Σ mass/density`. The two are not the same moment.

`r_dry` is the mode's **number-median** radius, which both microphysics cores
define through the third moment

```
V = N·(π/6)·Dg³·exp(4.5 ln²σ_g)        (placeholder.py:92-107; mam4_jax calcsize)
```

so `N·(4/3)π·r_dry³` = `V_dry / exp(4.5 ln²σ_g)` — deliberately *not* the mode's
dry volume. Mixing it with the third-moment `vol_dry` understated the water by

| σ_g | modes | factor |
|---|---|---|
| 1.6 | Aitken, primary carbon | 2.70 |
| 1.8 | accumulation, coarse | 4.73 |

**off a clip bound**. On one, the general factor is
`(dg_clip/dg_true)³ / exp(4.5 ln²σ_g)`, which moves either way: clipping DOWN
to `dgnum_hi` understates further (the A/B column's primary-carbon mode, ~19×),
while clipping UP to `dgnum_lo` — more number than the mass supports — can
**overstate** the water instead (the unit fixture's coarse mode, 11.2× too
much). Either way the volume-mixed index is wrong on the default optics path.

The water volume now comes from the dry volume and the growth factor:

```python
ratio = r_wet / jnp.maximum(r_dry_i, _MIN_DRY_RADIUS)
growth_cubed = jnp.where(r_dry_i > _MIN_DRY_RADIUS, ratio ** 3, 1.0)
v_water = vol_dry * jnp.maximum(growth_cubed - 1.0, 0.0)
```

Each core grows a mode by applying one ratio to the whole of it — the
placeholder core's κ-Köhler factor with the Kelvin term dropped, MAM4's
`wateruptake` Köhler solve (Kelvin kept) rescaling `dgncur_awet` from
`dgncur_a` — so **every** radius scales by the same `g = r_wet/r_dry` and the
wet third moment is exactly `g³` times the dry one. Four properties follow:

- It is **exact for the whole lognormal, for any σ_g**.
- Only the *ratio* of the radii enters, and the cores set `r_wet = r_dry·g`, so
  it **stays exact under `dg` clipping** — #790's stated caveat does not apply
  to this form.
- *Separately, and only off a clip bound*: `V_dry + V_w` is then the third
  moment of the same lognormal the Gauss–Hermite quadrature integrates the Mie
  efficiencies over, so mixing rule and size integral describe one particle
  population. Clipping breaks that second statement but not the first — the
  size integral follows the clipped radius while `V_dry` follows the mass, and
  the two part company by `(dg_clip/dg_true)³`.
- It carries **no modal number**, so it is identically zero on a mode with no
  dry material whatever ringing the number field holds. It is also the relation
  the placeholder core already applies to the wet density
  (`ρ_wet = (ρ_dry + (g³−1)ρ_w)/g³`), so the optics and the core describe the
  same particle.

A degenerate `r_dry` is **selected out**, not merely floored: the `jnp.where`
gives no growth, no water and no gradient there. The `_MIN_DRY_RADIUS` floor
stays inside the ratio because both arms of a `where` are evaluated and an
`inf` in the discarded arm returns as a NaN cotangent. Purely defensive —
neither in-repo core can produce `r_dry = 0` (the placeholder falls back to
`mode.dgnum`, MAM4-JAX takes `0.5·dgncur_a`).

Making the water proportional to `vol_dry` also means a mode with no dry
material now has `vol_tot` **exactly** zero, where before it inherited a
spurious number-driven water volume that kept it positive. That newly reaches
the mixing-rule and apportionment reciprocals, which divide by a floored
`vol_tot` — finite forward, but in reverse the local partial of
`1/max(v, 1e-30)` is `-1e60`, overflowing float32 to `-inf`, and the outer
select's zero cotangent makes it `0 × inf = NaN`. Substituting the denominator
*before* the division fixes it. Measured `d(od550_wat)/d(mass scale)` at an
empty cell: `0.0` on dev, `NaN` on this branch's first commit, `0.0` now.
Clean columns are most of a cold-start aerosol field and one NaN poisons the
whole reverse pass.

## Class sweep

No other site in `jcm` computes aerosol water or a wet volume from `N·r³`:

- `jam/emissions/seasalt.py:88` — `(4/3)π·rd³` is a per-bin particle mass inside
  a **sectional** Gong/Monahan emission integral, correct as written.
- `diagnostics/aerocom.py` — uses `r_dry` only with σ_g for a
  number-above-threshold (`erfc`); no volume.
- `radiation/cloud_optics.py` — `(4/3)π·r_eff³` cancels against its own
  `π·r_eff²`, giving the textbook `τ = 3·LWP/(2 ρ_w r_eff)`; not the same defect.

## Tests

`test_more_aerosol_more_aod` **had to change**, and its old form was the tell:
it doubled the mass tracers while holding `_jam_state` fixed. The per-particle
cross-section comes from `_jam_state` (number and wet radius), so mass enters
only through the refractive index, whose species ratios are scale-free — the
extinction could not respond at all. It passed only *because* the water fraction
was number-dependent and so shifted the index. It now re-diagnoses the modal
state from the doubled tracers with the placeholder core (new `_consistent_state`
helper), which is the state a core would actually produce.

New tests pin the properties the fix rests on:

- `test_refractive_index_independent_of_number` — AOD exactly linear in the modal
  number, SSA and asymmetry invariant.
- `test_water_depends_on_growth_ratio_not_on_number` — with the number/radius
  pairing made deliberately inconsistent, `r_wet == r_dry` adds exactly zero
  water *and* a 1.05 growth factor gives a water share of exactly `(g³−1)/g³`.
- `test_water_fraction_is_the_third_moment_growth` — the water share of each
  mode's extinction is exactly `(g³ − 1)/g³`, checked through the
  `od550_wat` / `od550_mode_*` AeroCom apportionment so it is exact and
  Mie-independent. Uses `center_modes=True` so every mode sits strictly inside
  its `dg` bounds, which the test asserts rather than assumes.
- `test_degenerate_dry_radius_gives_zero_water_and_zero_gradient` — `r_dry = 0`
  carrying real mass adds no water (dev: 0.698 of spurious water optical
  depth) and no gradient, and an all-empty cell keeps a finite gradient.

All four fail on `origin/dev` and pass here, checked by running this branch's
test module against the pristine dev tree.

`_consistent_state`, the helper that re-diagnoses `_jam_state` from the
fixture's tracers, documents that with `_setup`'s own numbers three of the four
modes are clipped (Aitken and primary carbon at `dgnum_hi`, coarse at
`dgnum_lo`), so its `r_dry` is the clipped radius and not the solution of the
third-moment relation. `center_modes=True` picks, per mode, the number that
puts `dg` exactly on `mode.dgnum` — strictly inside every bound — for the one
test that appeals to that relation.

Drive-by in the same file: the stray `if __name__ == "__main__": unittest.main()`
block sat *before* `OpticsDiagnosticsTest`, hiding that whole class from a direct
`python optics_term_test.py` run; moved to the end of the module.

## Provenance, verified against the sources

- **CAM** computes this same quantity per particle:
  `modal_aero_wateruptake.F90::modal_aero_wateruptake_sub` takes
  `wetvol = (4/3)π·wetrad³` and `wtrvol = wetvol − dryvol` (596-598), with
  `qaerwat = ρ_w·naer·wtrvol` (456). That is self-consistent *there* because
  CAM's `dryrad` is the dry **volume-mean** radius,
  `dryrad = (dryvol/pi43)**third` from the single-particle-mean `dryvol`
  (`modal_aero_calcsize.F90::modal_aero_calcdry`, 1565-1567), so `naer·wtrvol`
  telescopes to exactly `V_dry·(g³−1)` — this PR's form. CAM's *number-median*
  diameter is a different quantity, `dgncur_a = (drv/(dumfac·num))**third` with
  `dumfac = exp(4.5 ln²σ)·π/6` (`calcsize:549, 685`), and the mode is grown by
  one ratio applied to it, `dgncur_awet = dgncur_a*(wetrad/dryrad)`
  (`wateruptake:455`). **That is precisely why the per-particle shape could not
  be transplanted onto this code's `r_dry`**, which is number-median.
- **ECHAM-HAM** carries aerosol water as a per-mode tracer and volume-mixes it
  with the dry species before the optics lookup
  (`mo_ham_rad.f90::ham_rad_refrac_volume`, "Add aerosol water" at 277-298,
  summing `zv = mass/density` into the same `znrsum`/`znisum`/`zvsum`), and
  reports it as its own optics component (`ham_rad_diag` 1898-1906, apportioned
  by volume fraction at 1927-1933, written at 1943; the `TAU_COMP_WAT` stream
  element is `mo_ham_streams.f90:652`).

## SCM A/B

Release-validation JAM single column (`jcm.rce.jam_scavenging_column`, 47 levels,
2M clouds, grey radiation, AeroCom optics on, SST 302 K, RH 0.8), 6 days at
dt = 900 s, means over the last 3 days:

| diagnostic | dev | this PR | change |
|---|---|---|---|
| `od550aer` | 5.102e-3 | 4.679e-3 | −8.3% |
| `jam_optics.aod_550` (grey's sole SW band *is* 550 nm, so it tracks exactly) | 5.102e-3 | 4.679e-3 | −8.3% |
| `abs550aer` | 6.250e-5 | 5.918e-5 | −5.3% |
| `od550_wat` | 1.217e-3 | 2.381e-3 | ×1.96 |
| water share of `od550aer` | 0.239 | 0.509 | ×2.13 |
| `od550_so4` | 2.442e-3 | 1.009e-3 | −59% |
| `od550_poa` | 1.318e-3 | 1.240e-3 | −5.9% |
| AOD-weighted column SSA | 0.98773 | 0.98733 | −4.0e-4 |
| AOD-weighted column asymmetry | 0.6711 | 0.6787 | +0.0076 |
| boundary-layer SSA (p > 850 hPa) | 0.97843 | 0.97879 | +3.6e-4 |
| `ssa440aer` | 0.96861 | 0.96805 | −5.6e-4 |
| `ang4487aer` | 1.632 | 1.716 | +0.084 |

Directions, checked rather than assumed:

- **Water share up ×2.1**, and the per-mode shares match the analytic
  `(g³−1)/g³` to within ~1.5% — accumulation (unclipped, σ_g = 1.8)
  0.300 → 0.670 predicted against 0.312 → 0.679 diagnosed; primary carbon
  (clipped at `dgnum_hi`, σ_g = 1.6) 0.0021 → 0.0385 predicted against
  0.0020 → 0.0379 diagnosed. (The residual is because the column diagnostic is
  an AOD-weighted mix over levels whose RH is not exactly the 0.8 the analytic
  value assumes.) The clipped mode here is the ~19× under-water case; a mode
  clipped the other way, at `dgnum_lo`, would have been over-watered instead.
- **AOD down 8.3%.** The number and wet radius come from the core and are
  untouched, so the geometric cross-section `n·π·r_wet²` is identical; the AOD
  moves only through `Q_ext` at the lower mixed index (water n ≈ 1.33 against
  sulfate's ≈ 1.43–1.53). That is the "too refractive" bias being removed.
- **SSA barely moves** (−4e-4 at 550 nm, +4e-4 in the boundary layer), which is
  *not* the "clearly less absorbing" one might expect. Absorption does fall
  (−5.3%), because `k_eff = Σ V_s k_s / V_tot` is diluted by the extra water,
  but scattering falls slightly more: the lower real index costs more `Q_sca`
  than the dilution saves `Q_abs`. At fixed particle size the direct effect here
  is an extinction change, not an SSA change.
- **`od550_so4` −59%** is apportionment, not a change in sulfate: water's share
  of the accumulation mode's volume more than doubles, and extinction is split
  by volume fraction. The components still close on `od550aer` exactly
  (`test_species_apportionment_closes`).
- **Ångström +0.08**: 865 nm loses relatively more extinction than 440 nm as the
  real index drops, steepening the spectral slope.

## Verification

- `ruff check .` clean (whole repo).
- Full fast suite (`-m "not slow" -n 8`) run in this branch **and** in a pristine
  `origin/dev` worktree at the same base commit (figures from the first commit;
  the review round changed only `optics_term.py` and its test module, covered
  by the targeted re-run below):
  **150 failed / 2498 passed / 45 skipped / 7 errors** here against
  **150 failed / 2494 passed / 45 skipped / 7 errors** on dev. The 157
  failure-plus-error identifiers are an **identical set**; the four extra passes
  are the new tests.
- Those 157 are all environmental in this local checkout, which is not at
  requirements parity: 109 × `RuntimeError: the installed dinosaur has no
  semi-Lagrangian transport`, 36 × `RRTMGP.compute_heating_rate() got an
  unexpected keyword argument 'sfc_alb'` (older `jax-rrtmgp`), 5 ×
  `ModuleNotFoundError: No module named 'huggingface_hub'`, and 7
  `cf_metadata_test` collection errors downstream of the same. CI installs
  `pip install -e .` at parity and is the real gate for those, coverage included.
- `jcm/physics/aerosol` + `jcm/physics/diagnostics` in full (slow tests
  included), re-run against the pristine dev worktree after the review round:
  **19 failed / 480 passed / 23 skipped** here against **19 failed / 476
  passed / 23 skipped** on dev — identical failure sets, the four extra passes
  being the new tests. That is the subsystem the change touches, so nothing
  downstream pins the old numbers.
- This branch's optics test module run against the pristine dev tree: the four
  #790 tests fail there and pass here.
- `d(od550_wat)/d(mass scale)` at an empty cell: `0.0` on dev, `NaN` on this
  branch's first commit, `0.0` after a7bfbc02. The SCM A/B above reproduces
  bit-identically across that commit, so the reverse-mode fix is
  forward-neutral.
- `sphinx -b html` builds both edited pages without new warnings.
- The SCM A/B above was run once per worktree from the same probe, which lives
  in a scratchpad rather than the repo.

## A limitation this makes visible — #823

With the water number-free the mixed refractive index is scale-free in a mode's
masses, so **the core's `dg` diagnosis is the only burden-to-radiation
channel** — and it saturates on a clipped mode, whose AOD then does not respond
to its mass at all. Doubling every mass tracer at fixed `_jam_state` changes
the summed band AOD by **+3.273% on dev** and **exactly 0.000% here**. The dev
response was the artefact — the old water volume's number dependence moving the
index — not a physical mass channel, so this is a pre-existing gap that this PR
exposes rather than creates. Neither core adjusts number to bring a clipped
mode back inside its bounds the way MAM4's `calcsize` does. Recorded in the
science doc's "Status & known limitations" and filed as #823, which also sets
out what a fix would need and why it wants its own measurement campaign.

## Review round (a7bfbc02)

| finding | resolution |
|---|---|
| Clipping only ever makes the old deficit larger — false; clipping up to `dgnum_lo` over-waters (fixture coarse mode, 11.2×) | Confirmed and fixed in a7bfbc02 — corrected in the code comment, `aerosol.md`, the test docstring and this body |
| "`V_dry + V_w` is the third moment the quadrature integrates" conflated with the exactness of `V_w` | Confirmed and fixed in a7bfbc02 — the two are now split, the second qualified to the unclipped case |
| Optics now exactly mass-invariant at fixed `_jam_state`, so a clipped mode's AOD cannot respond to mass | Confirmed (+3.273% dev / 0.000% here); documented in the science doc and filed as #823 |
| `_consistent_state` docstring claims a third-moment `r_dry` while 3 of 4 modes clip | Confirmed and fixed in a7bfbc02 — truthful docstring plus an opt-in `center_modes` that avoids every clip, asserted in the test |
| `test_dry_particles_add_no_water` passes on dev, so its rationale is inert | Confirmed and fixed in a7bfbc02 — replaced with a growth-ratio test that fails on dev |
| `test_empty_mode_...` passes on dev; no in-repo core yields `r_dry = 0` | Confirmed and fixed in a7bfbc02 — reworded as explicitly defensive, and strengthened to a case that does fail on dev |
| Prefer a `where`-gate over the bare floor for a zero gradient on the degenerate branch | Confirmed and fixed in a7bfbc02 — measured `d/d vol_dry` 1e15 → 0; it also surfaced the clean-column NaN gradient above, fixed in the same commit |
| Gate comment: a total-volume gate "would be the same test" | Confirmed and fixed in a7bfbc02 — now "monotone in `vol_dry`, the same cells at a threshold differing by `g³`" |
| "Kelvin term dropped" describes only the placeholder core | Confirmed and fixed in a7bfbc02 — worded for both cores |
| CAM `modal_aero_wateruptake` is literally this PR's form | Verified on the CAM source and cited, though not as `wetvolmr = dryvolmr*(rwet/rdry)**3` — this version writes `wetvol = pi43*wetrad**3` then `wtrvol = wetvol - dryvol`, which telescopes to the same thing *because* `dryrad` is the volume-mean radius. That distinction is the sharper provenance and is what the doc now says. |
| The two ECHAM-HAM assertions are unbacked | Verified against the local `mo_ham_rad.f90` and cited by line |

## Related

PR #792 (opt-in NeuralMie bulk-optics backend) already carries this same
number-free water volume on its opt-in path, so the `if backend == ...` split
around the water volume there collapses to a single expression once this lands.
That branch floors `r_dry` with `_TINY` where this one selects on
`_MIN_DRY_RADIUS`, and it does not carry the clean-column reverse-mode fix,
since on its path `vol_tot` is only exactly zero once the water is number-free
everywhere.

Closes #790


duncanwp:
@codex review

Issue #782, opened 2026-09-09
Tiedtke trigger chain (#690) removes ~0.11 of low cloud and shifts TOA in ECHAM 1M/2M — unvalidated climate change (plus a second ~0.05 from post-#690 dev)


duncanwp:
> [!IMPORTANT]
> **Correction (bisect in progress).** The attribution below credits the whole −0.161 low-cloud change and +15.7 W/m² TOA shift to #690. A commit-level bisect shows that is **too much**. Measuring the squash's own code point (PR-tip `41ad1795` ≈ `66b2d758`) at 30 days gives low cloud **0.331**, not arm A's 0.276, so the change splits:
>
> | segment | low cloud | Δ |
> |---|---|---|
> | `1b7e7078` (pre-#690) | 0.437 | — |
> | `66b2d758` (post-#690) | **0.331** | **−0.106 (≈2/3)** |
> | `f5c8b354` (dev HEAD) | **0.276** | **−0.055 (≈1/3)** |
>
> #690 is the **majority** contributor, not the sole one; a second, separate contribution lives in the 19 merges between `66b2d758` and dev HEAD. Both bisects are running; a consolidated comment with the per-commit trajectory and the reading of each responsible change will follow. The A/B arms and the year-run table below are unchanged and remain correct as measured — only the *attribution of the whole change to #690* is being revised.

Found by the pre-release validation sweep of `origin/dev` @ `f5c8b354` (issue #638 matrix).

#690 ("the faithful ECHAM trigger chain + water-budget cluster", squash-merged as `66b2d758`) removes **0.14 of total cloud cover — 0.16 of it low cloud** — and moves TOA net by **+15.7 W/m²** in the ECHAM 1M/2M configurations. The change is real, large, and was not validated against a climatological reference in the PR. This issue is about **whether the new cloud state is right**, not about whether the port is faithful (per #775 the `cubase` port is faithful to `mo_cuinitialize.f90`).

## Attribution: a controlled three-arm A/B

30 days, T63L47, one A100 per arm, identical explicit override set (verified `diff`-identical to `+configuration=t63-echam-1m`), same deterministic `init=jw init.rh=0.0` state, same dates. Only one factor varies per arm. Versions below are read back from each run's own `jcm_prov_code` attrs, not from the launch flags.

| arm | jcm | jax-rrtmgp | cloud total | low | mid | high | TOA net W/m² | precip mm/day | rate d/hr |
[...]

Issue #682, opened 2026-08-19
Tiedtke convection: retune the trigger/closure against corrected (no longer inflated) CAPE


duncanwp:
Follow-up to #661. Parent: #614.

## What's wrong

#661 removed a cloud-base saturation adjustment that created water and over-warmed the plume's seed parcel. That bug inflated CAPE substantially — 49110 → 16270 J/kg on the `rce_integration_test` sounding, 19047 → 8547 on a plain tropical one, 5039 → 4287 on a hydrostatic 47-level column. The convective trigger and the mass-flux closure were tuned while that inflation was present, so they are now calibrated against a quantity that has moved.

The clearest symptom is the whole-model single-column RCE test (`jcm/rce_test.py::TestRceWholeModelTiedtke::test_whole_model_column_reaches_physical_time_mean_rce`). Measured over the last 40 days of the 80-day integration:

| | convection ON | ktype switches/day | `precip_conv` zero-frac | max temporal std |
|---|---|---|---|---|
| `dev` @ `f6d1bcd` | 99.1 % | 0.8 | 0.9 % | 9.9 K/day |
| with #661 | 9.4 % | 3.4 | 91.4 % | 28.6 K/day |

The column now parks **below** the 100 J/kg `trigger_cape` and flips on and off, which is the on/off extinction pathology the flicker bound was originally written to guard against. The bound was re-pinned from 16.5 to 32.0 in the #661 PR to track the measurement; it should come back down here.

## Why it matters, and why it is NOT a reason to revert #661

The global model is **not** showing this. A 3-day T63L47 ECHAM run (real orography and SST, NaN-free) gives:

| | `dev` | with #661 (full) |
|---|---|---|
| convective precip | 0.776 mm/day | 0.948 mm/day |
| large-scale precip | 0.275 mm/day | 0.125 mm/day |
| total precip | 1.052 mm/day | 1.073 mm/day |
| fraction of columns convecting | 0.448 | 0.445 |

So the trigger sensitivity bites on a *closed* single column sitting near its threshold, not across the globe. #661 itself says the fix should land before anything downstream of it is re-tuned; this issue is that re-tuning.

## Where to look

- `types.py` — `trigger_cape` (100 J/kg) and `smooth_trigger_j` (25 J/kg), the sigmoid trigger from #556.
[...]

Issue #661, opened 2026-08-19
Tiedtke convection: the cloud-base saturation adjustment creates water and over-warms the parcel


duncanwp:
Found by a re-audit of the July 2026 ECHAM+RRTMGP review against `dev` @ `f6d1bcd`. Parent: #614.

## What's wrong

`jcm/physics/convection/tiedtke_nordeng/updraft.py:187-201` performs the cloud-base saturation adjustment by hand instead of calling `cuadjtq`. It has no damping denominator and it re-saturates at the *warmed* temperature:

```python
qsat_at_cb = saturation_mixing_ratio(pressure[kbase], parcel_T_dry_at_cb)
excess     = jnp.maximum(surf_humid - qsat_at_cb, 0.0)
tu_cb      = parcel_T_dry_at_cb + (c.alhc / c.cpd) * excess
qu_cb      = saturation_mixing_ratio(pressure[kbase], tu_cb)     # > qsat_at_cb
```

so `qu_cb + lu_cb > surf_humid`: **water is created and the parcel is over-warmed.**

There is a second, independent source. `qu_cb` is assigned unconditionally, so when `excess == 0` — the parcel is *sub*saturated at `kbase`, the normal outcome of `find_cloud_base`'s discrete `argmax` at `tiedtke_nordeng.py:146-161` — the code still moistens the plume up to saturation, **with no latent-heat debit at all**.

## Evidence

Probed on a 47-level column (`JAX_PLATFORMS=cpu`), against the conserving routine that already exists in the same file (`updraft.py:46-114`):

| p / T_dry / q_in | this code | `updraft.saturation_adjustment` |
|---|---|---|
| 950 hPa, 295 K, 18 g/kg | ΔT **+1.66 K**, water **+10.3 %** | ΔT +0.45 K, water +0.000 % |
| 900 hPa, 290 K, 16 g/kg | ΔT **+6.54 K**, water **+42.2 %** | ΔT +2.01 K, water +0.000 % |
| 970 hPa, 300 K, 22 g/kg (**subsaturated**) | ΔT 0, water **+4.4 %** | ΔT 0, water +0.000 % |

On the coarser `rce_integration_test` sounding the error reaches +80 % water / +11.8 K at kbase 905 hPa, and +270 % / +36 K at 743 hPa — **it grows as the vertical grid coarsens.**

## Why it matters

[...]

PR #822, opened 2026-09-15
Release validation: score total cloud cover on ECHAM's max-random definition (#782)

duncanwp:
Addresses item 2 of #782's "Revised next steps": re-derive the release-validation `cloud_cover` gate on a defensible definition, move the band with it, and record what old numbers may and may not be compared with.

> **Review round.** The first revision of this PR got two things wrong that are now fixed and called out below in full: it left the band at 0.4–0.8 under a definition that reads ~0.13 higher, and it misdescribed the Fortran it was citing. A "Review round" table at the end maps every finding to its commit. The proposed band is stated as a proposal — the number is the maintainer's to set.

## The audit first: no code defect in the physics

Before changing the gate I re-read both halves of #782's decomposition against the ICON Fortran, since a gate change would be the wrong response to a real defect.

**#661/#690 cloud base is a faithful `cubase` port.** `find_cloud_base` (`jcm/physics/convection/tiedtke_nordeng/tiedtke_nordeng.py:155`) matches `cubase` (`mo_cuinitialize.f90:276-320`): the `klab` walk, `zlift = min(clip(thvsig·cbfac, cminbuoy, cmaxbuoy), 1)`, stopping at the LCL, and the moist buoyancy test. Its effect (−0.106 low cloud, +14.5 W/m² TOA) is radiatively real and belongs to the retune question in #682, not here.

**The #707 1M write-back is faithful too, and is bookkeeping.** `mo_cloud.f90` section 8.4 (ICON `atm_phy_echam/mo_cloud.f90:1129-1138`) reads

```fortran
zxlp1_d      = ccwmin - zxlp1                ! from the RAW liquid
zxip1_d      = ccwmin - zxip1                ! from the RAW ice
zxlp1        = FSEL(-zxlp1_d, zxlp1, 0)      ! zero a phase below ccwmin
zxip1        = FSEL(-zxip1_d, zxip1, 0)
zxlp1_d      = MAX(zxlp1_d, 0)               ! liquid only
paclc(jl,jk) = FSEL(-(zxlp1_d*zxip1_d), paclc(jl,jk), 0)
```

`FSEL(a,b,c)` is `a >= 0 ? b : c`, so the cover is cleared exactly when `zxlp1_d * zxip1_d > 0`. Both `_d` terms come from the **raw, pre-clamp** condensates and are *not* recomputed after the two `FSEL` clamps, so `zxip1_d` can be anywhere in `(0, ccwmin)`. What makes the test well defined is the `MAX` on the **liquid** term alone: it leaves `zxlp1_d ≥ 0`, so a positive product requires both `zxlp1_d > 0` and `zxip1_d > 0`, i.e. `qc < ccwmin` **and** `qi < ccwmin`. Without that `MAX`, a cell with *both* phases above the threshold would have two negative `_d` terms, a positive product, and its cover wrongly cleared. `jcm/physics/clouds/echam_1m.py:1408` implements the resulting test directly as `(qc_end < ccwmin) & (qi_end < ccwmin)`. Faithful. Its measured cost is −0.066 of low cloud for +0.15 W/m².

(The first revision gave a different derivation — "after the two phase clamps each condensate is either exactly 0 or at least `ccwmin`" — which is not what the Fortran computes, because the `_d` terms are never refreshed from the clamped values. The conclusion and jcm's implementation were right; the derivation was not. Corrected in `199643bb`/`5e8f5055` and in the design doc.)

So the actionable finding is the gate, which is what this PR changes.

## What the gate scored, and what it scores now

It scored `ds["clouds.cloud_fraction"].max("level")` — a **column maximum**, only a lower bound on total cover: it assumes every layer overlaps maximally, so two half-covered decks at opposite ends of a column read 0.5 where the sky is 0.75 covered. It is not the construction ECHAM reports, nor a total cover of the kind the satellite climatologies are quoted on, yet the 0.4–0.8 band's anchors were taken from exactly those.

It now scores **ECHAM's own `aclcov`** — maximum-random overlap, `mo_cloud.f90` section "10.2 Total cloud cover" (ICON `mo_cloud.f90:1165-1182`, ECHAM6 `mo_cloud.f90:1359-1383`) — through a new `jcm.analysis.total_cloud_cover`, including the `zxsec = 1 - 1e-12` guard that keeps an overcast layer's `1/(1 - paclc)` finite.

### Measured, on real output

Scored with this branch's `total_cloud_cover` on the shared dev workstation, area-weighted, reduction taken **after** the overlap product. The runs save 5-day means, so "last 40 saved frames" is the window `--last-n 40` picks out of a 5-day-chunked run — the settled ~200 days.

| run | code point | window | column max | **max-random** | random | offset |
|---|---|---|---|---|---|---|
| 2M year `v10_2m_prefix_year/echam-rrtmgp-2m_day*.nc` | dev @ 2026-07-04 | full year (73 frames) | 0.544 | **0.691** | 0.843 | +0.147 |
| " | " | settled (last 40) | 0.553 | **0.701** | 0.847 | +0.148 |
| 1M year `v7_1m_fullyear/echam-rrtmgp_day*.nc` | dev @ 2026-07-04 | full year | 0.555 | **0.679** | 0.815 | +0.124 |
| " | " | settled (last 40) | 0.563 | **0.691** | 0.823 | +0.128 |
| 2M+JAM 90-day arm `jam_scav_ab/abbase_260823_0100_day{30,60,90}.nc` | jcm `0ee92eaa`, **post-#707** | days 1–90 | 0.425 | **0.535** | 0.699 | +0.111 |
| " | " | days 61–90 | 0.417 | **0.527** | 0.703 | +0.110 |

Caveats, stated because they bound what the table can be used for:

- Both year runs are **pre-#690/#707/#710**; their directories keep no `.hydra` snapshot, so the code point is fixed only by the run logs' date and the resolved config they echo. They are the only archived T63L47 ECHAM year runs on the box.
- The 2M year is its campaign's flagged pre-orientation-fix arm (its own `README.txt`: inverted MACv2-SP shortwave aerosol and ozone), so its *absolute* cover is not a climatology. It is in the table for its offset.
- The 90-day arm is the only archived **post-#707** ECHAM T63L47 output here (verified `git merge-base --is-ancestor 5ba96f7f 0ee92eaa`). It is a spin-up, so its absolute values are not climate; it is here to show the offset survives the #690/#707 cloud-state change.
- `radiation.total_cloud_cover` is **absent** from both year runs (they predate `b772ffec`). The 90-day arm saves it: 0.757 / 0.776 against a max-random 0.535 / 0.527 on the same files — see "not a cross-check" below.

Two readings: the spread across definitions is **~0.27–0.30**, larger than any model change the gate has been asked to judge; and the column max is **0.11–0.15 low**, stably across two microphysics schemes, two code points and the #690/#707 boundary.

### Reconciling with the #638/#782 matrix

Those numbers are **column maxima** from a different campaign, a different (later) code point, at T63 *and* T106, `--last-n 40`. They are not directly comparable with the table above — the column max there reads 0.54–0.56 against 0.60–0.70 in the matrix, and that gap is model change plus resolution, not definition. What transfers is the offset:

| member | column max (#638 → post-#690/#707) | max-random (mapped) |
|---|---|---|
| `echam-1m-t63` | 0.68 → 0.63 | ~0.81 → ~0.74 |
| `echam-1m-t106` | 0.70 → 0.66 | ~0.83 → ~0.77 |
| `echam-2m-t63` | 0.60 → 0.48 | ~0.75 → ~0.59 |
| `echam-2m-t106` | 0.61 → 0.49 | ~0.76 → ~0.60 |
| `echam-jam-t63-l47` | 0.61 → (scrapped) | ~0.76 → — |
| `echam-jam-t63-l95` | 0.61 → (scrapped) | ~0.76 → — |
| `speedy-t31` | 0.57 → 0.58 | **n/a — different quantity** |

using the scheme-matched offset for the #638 column (1M +0.128, 2M +0.148, the JAM members being 2M) and the post-#707 offset (+0.110) for the current one. These are *mapped*, not measured: no post-#707 year run is archived anywhere, so nothing can be measured on this definition yet. The next sweep prints all three covers and should replace the mapping.

## Proposed band: `RANGES["cloud_cover"] = (0.5, 0.9)`, SPEEDY unchanged at (0.4, 0.8)

**The first revision left the band at 0.4–0.8, and that was wrong.** Changing what the gate measures while leaving the numbers loosens the floor and tightens the ceiling by the definitional offset — and the mapping above shows it is not academic: `echam-1m-t106`'s #638 baseline maps to ~0.83 and `echam-1m-t63`'s to ~0.81, so members that passed would now fail on the ceiling for a reason that has nothing to do with the model. That is the opposite of the reassurance the PR text gave.

The ECHAM band is therefore the previous one at the **same width**, shifted by the measured offset rounded to 0.1. Derivation:

- **Offset.** +0.11 to +0.15 measured (table above); rounding to 0.1 rather than 0.13 keeps more room at the floor, which is where the 2M members sit.
- **Observations.** The GEWEX Cloud Assessment database gives a global cloud amount of **0.68 ± 0.03** for optical depth > 0.1, rising to 0.74 for COD > 0.01 (CALIPSO) and falling to 0.56 for COD > 2 (POLDER) — [Stubenrauch et al. 2013, BAMS 94, 1031–1049](https://doi.org/10.1175/BAMS-D-12-00117.1), values as summarised by the [NCAR Climate Data Guide](https://climatedataguide.ucar.edu/climate-data/cloud-dataset-overview). The whole 0.56–0.74 range sits inside 0.5–0.9.
- **Model.** Every **ECHAM** member maps into **0.59–0.83**; the directly measured year runs sit at 0.68–0.70. Tightest margin 0.07, at the ceiling, held by a superseded baseline; the current code point maps to 0.59–0.77 with 0.09 at the floor.

**SPEEDY keeps 0.4–0.8, under its own `RANGES["cloud_cover_speedy"]`.** This is the same argument applied honestly in the other direction, and the first revision got it wrong too: `shortwave_rad.cloudc` is SPEEDY's own RH-based column cover, has no profile to overlap, and nothing here touched it — so no offset applies. Sharing one band would have cut `speedy-t31`'s floor headroom from 0.17 to 0.07 (recorded 0.57 in #638, 0.58 in #782) for a definitional reason that does not apply to it.

The gate keeps its character — loose, spin-up tolerant, "did the model produce a climate" rather than a tuning target. **This is the proposal; the numbers are the maintainer's to set.** If you would rather keep a 0.4 floor (0.4–0.9), or wait for a measured post-#707 year before moving anything, say so and I will change it — but leaving ECHAM at 0.4–0.8 under the new definition is the one option the measurements rule out.

## Two claims from the first revision that were too strong

**"Not comparable on any of the three definitions."** The boundary does apply to all three — they are reductions of a redefined field — but the **−0.066 figure is a column-max measurement** and does not carry across. Nothing has measured #707 on the max-random cover, and `radiation.total_cloud_cover` reduces `effective_cloud_fraction` (`jcm/physics/radiation/mcica.py`), which applies its own independent clear-cell threshold (`cf <= 2*cld_frac_min`) on an overlapping but different set of cells. Now hedged in `health.py`, the README and the design doc.

**`cloud_cover_radiation` as "the cross-check".** It is a *different measurement*, and the two are not expected to agree, for two independent reasons: it reduces a differently-preprocessed field, and it is a time mean of an *instantaneous* cover while `cloud_cover` is an overlap of the **output-averaged** fraction the file holds (`run.output_averages`, set unconditionally by `tools/release_validation/launch.py`). Measured on the 90-day arm: 0.53 against 0.78. That also means the gated number is strictly a lower bound on the time-mean instantaneous cover a climatology reports, on the one arm where both are available — worth knowing before reading a small offset from an anchor as a model bias. Documented rather than "fixed": both are honest numbers, and the file is what it is.

## Changes

- `jcm/analysis.py`: `total_cloud_cover(cloud_fraction, dim="level")`, built from **lazy** xarray slices so a dask-backed window stays lazy; float64 accumulation and the `[0, 1]` clip applied per level, out of place. CF attributes on the result. Orientation independence stated as agreement *to rounding* (the `min(c, zxsec)` guard is applied in loop order).
- `tools/release_validation/health.py`: gates ECHAM on it at 0.5–0.9 and SPEEDY on its own 0.4–0.8; `cloud_cover_fields` returns the *reason* a radiation cover is missing so the NOTE no longer conflates "absent" with "present and identically zero"; the all-zero test and the NaN scan also reduce lazily.
- `tools/release_validation/health_test.py`: helper-level, `main()`-level, and band-placement tests.
- `docs/source/design/cloud_cover_gate.md`: definition, provenance, the measured table, the band derivation, the SPEEDY decision, the #707 discontinuity, the #782 decomposition. In the `design.rst` toctree, cross-linked from `health.py`.
- `docs/source/science/radiation.md`: which total cover jcm reports offline and why, next to the in-model overlap treatments it already describes (CLAUDE.md's living-document rule).
- `docs/source/getting_started.rst`, `tools/release_validation/README.md`: the user-facing halves.

## Verification

- `ruff check .` clean. (Local ruff is 0.15.9; CI pins 0.15.17 — CI is the authority on that one.)
- `jcm/analysis_test.py`: 16 cloud-cover tests — single layer, adjacent equal/unequal, separated-by-clear random combination, bracketing between the bounds, orientation flip (including the `O(zepsec)` near-1 case), clipping, `c = 1` at each end and in the interior, dims/coords preserved, non-default `dim`, missing `dim`, empty `dim`, **dask-backed bit-identical to numpy**, **NaN propagation with the caller's array unmodified**, **float32 input** (production `physics_dtype`, where dropping the upcast silently NaNs every overcast column), CF attributes.
- `tools/release_validation/health_test.py`: 16 tests. The `main()`-level and band-placement ones are the point — verified by reverting: reverting the call site to the column max, the band to (0.4, 0.8), or the float64 upcast each now turns the suite red, and none of them did before.
- Memory, the Codex finding: peak RSS on a synthetic 40-chunk T63L47 dask window (40×47×96×192 float32, one chunk per saved step, single-threaded scheduler) **846 → 460 MiB**, i.e. 508 → 122 MiB above the import baseline.
- Full fast suite (`-m "not slow" -n 4`) run once in this branch and once in a pristine merge-base worktree (`76ffa78e`) on the same machine: **failure sets identical** (157 node IDs — 150 failures plus 7 errors), branch 2525 passed / base 2494. That local checkout is not at requirements parity (older dinosaur without SL transport, no `jax_rrtmgp` / `huggingface_hub` / `mam4_jax`), so set identity was the only meaningful signal; **CI, at dependency parity, is the authority** and is what should be read for this PR.

## One red check is not this PR

`science-register-guard` fails on `docs/source/science/aerosol.md`, a file this branch does not touch. It cites #790 as an open tracked gap, and #790 was closed by #821 (`6a5bef6d`), which landed on `dev` after this branch was cut. The guard runs only on `pull_request`, so `dev`'s own push never checked it and the next PR against `dev` inherits it — this one. Filed as #825, with the two resolutions the guard itself names; deliberately not fixed here, since which one applies is a JAM-optics call and a reviewer would be surprised to find it in a cloud-cover PR.

## Review round

| finding | resolution |
|---|---|
| **P1** band must move with the definition | Fixed in `ca1273a8`. Measured on real output; `RANGES["cloud_cover"]` proposed as (0.5, 0.9), derivation above. |
| **P1** score real output; replace the single-last-chunk anchors | Fixed in `5e8f5055`. Table above and in the design doc, with paths, code points, windows and caveats, plus a post-#707 offset probe. |
| **P2** (Codex, `jcm/analysis.py:254`) `.values` + `np.array(float64)` materialises the window | Fixed in `c814f820`; 846 → 460 MiB peak. Same pattern swept twice more in `health.py` (`d20b3945`). |
| **P2** gate wiring untested | Fixed in `ca1273a8`: `main()` driven over synthetic chunk directories. |
| **P2** uncited anchors in four places | Fixed in `5e8f5055`/`199643bb`. GEWEX/Stubenrauch 2013 for observations, Stevens 2013 for `aclcov`, Kay et al. 2012 for the inter-product spread. The unverifiable "ECHAM6 ~0.62–0.65" is **withdrawn** — see below. |
| **P2** over-broad comparability claim | Fixed in `5e8f5055`/`f6356817`: hedged in `health.py`, the README and the doc. |
| **P3** `FSEL`/`MAX` derivation misdescribed the Fortran | Fixed in `5e8f5055`. |
| **P3** orientation "identical" → "to rounding" | Fixed in `c814f820`, with a test and a 200k-profile check (worst difference 1.1e-16). |
| **P3** #730 "no change in cloud"; ECHAM6 `END DO` line | Fixed in `5e8f5055`: "essentially no change (≤0.005)" with the numbers; line 1383, not 1381. |
| **P3** formula block should show the clip and the loop range | Fixed in `5e8f5055`/`199643bb` (including ICON's `jks+1` vs ECHAM6's `2`). |
| **P3** no attrs on the result | Fixed in `c814f820`: `cloud_area_fraction`, `units` `1`, `long_name`. |
| **P3** NOTE conflates absent with all-zero | Fixed in `ca1273a8`, with tests for both. |
| **P3** dated comment / stale usage line | Fixed in `ca1273a8`; dated context moved to the design doc. |
| **P3** PR numbers leaking into user docs | Fixed in `5e8f5055` (`getting_started.rst`). |
| **P3** orphaned "Rationale," line wrap | Fixed in `5e8f5055`. |
| **2nd round** SPEEDY swept along by the band change | Fixed in `b81f4d31`: its own `RANGES["cloud_cover_speedy"]`, left at 0.4–0.8. |
| **2nd round** "every member maps into 0.59–0.83" false (8 members, 3 listed) | Fixed in `4cfd77c4`: all members tabulated, claim scoped to ECHAM. |
| **2nd round** nothing pinned the band; nothing pinned the float64 upcast | Fixed in `b81f4d31`: 0.85-passes / 0.45-fails runs, a SPEEDY-band run, and a float32 overcast column. Each was verified to fail on a revert. |
| **2nd round** empty vertical axis; `radiation_note` invariant; fixture `len(_PROFILE)` | Fixed in `b81f4d31`. |
| **2nd round** science doc untouched | Fixed in `4cfd77c4`: `science/radiation.md` now records the offline overlap choice. |

**Not changed, deliberately.** An all-NaN `radiation.total_cloud_cover` passes the "is it identically zero" test and would print `nan` in the INFO line. The NaN scan already fails such a run and a printed `nan` is self-evidently not a measurement, so I left it rather than add a second special case — say the word if you would rather it were dropped like the all-zero field.

**One anchor could not be verified.** The earlier text quoted ECHAM6's climatological total cover as ~0.62–0.65 and the satellite products as "ISCCP/MODIS ~0.66–0.67, CALIPSO-GOCCP ~0.70". I could not verify any of those from an accessible source (Stevens et al. 2013 is paywalled), and the MODIS figure is contradicted by the literature: Kay et al. 2012 report 0.49 under the MODIS simulator definition, because it excludes optically thin and broken cloud. Rather than repeat unverifiable numbers, the anchor is the GEWEX value above, which I could verify, and Stevens et al. 2013 is cited for the *formulation* of `aclcov` only. If you have the ECHAM6 number to hand it would be a better third anchor and I will add it.

## What remains of #782

The physics question — retune the trigger and closure against the corrected cloud base, then re-measure low cloud, LWP and SW CRE against CERES (≈ −47 W/m²; the #638 sweep recorded jcm 1M ≈ −98 and 2M ≈ −58) — stays with #682. This PR fixes the measuring stick, not the cloud. `Refs #782` rather than `Closes`, since the validation question is the maintainer's to close.


jcm/analysis.py
```diff
@@ -156,3 +160,111 @@ def column_integral(q: xr.DataArray, dp: xr.DataArray) -> xr.DataArray:
...
+            f"{dim!r} is not a dimension of the cloud fraction "
+            f"(dims: {cloud_fraction.dims})")
+
+    axis = list(cloud_fraction.dims).index(dim)
+    # float64 throughout, in one allocation. ``zxsec`` is not representable in
+    # float32 (it rounds to 1.0, turning an overcast layer's guarded 0/1e-12
+    # into 0/0), so the working copy has to be double precision; and it has to
+    # be a genuine copy, since the in-place clip must not reach back into the
+    # caller's Dataset the way ``np.asarray`` on float64 input would allow.
+    c = np.array(cloud_fraction.values, dtype=np.float64)
```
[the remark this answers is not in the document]
duncanwp: Confirmed and fixed in c814f820. The running clear-sky product is now built from lazy `isel` slices, and the float64 conversion and the [0, 1] clip are applied per level rather than to the whole array, so nothing calls `.values` and a dask-backed window stays a dask graph evaluated chunk by chunk (the numpy path now holds a couple of horizontal slices instead of a full float64 copy of the window).

Measured on a synthetic 40-chunk T63L47 window (40 x 47 x 96 x 192 float32, one chunk per saved step, synchronous scheduler): peak RSS 846 -> 460 MiB, i.e. 508 -> 122 MiB above the import baseline. The dask and numpy paths are asserted bit-identical in `jcm/analysis_test.py`, alongside new NaN-propagation and no-input-mutation tests. The docstring claimed the old behaviour bounded memory and has been corrected.

The same pattern appeared twice more in `health.py` on the same `open_mfdataset` and both were swept in d20b3945: the all-zero test on `radiation.total_cloud_cover` loaded a year of a 2-D field to answer a yes/no, and the NaN scan loaded every saved variable in turn to reduce it. Both now reduce lazily with identical results.

duncanwp:
@codex review

docs/source/getting_started.rst
```diff
@@ -706,6 +707,23 @@ pressure thicknesses and column burdens, all computed on *saved* netCDF output:
    # column_burden already time-broadcasts, so a global-mean burden time series is:
    burden_ts = analysis.global_mean(analysis.column_burden(ds, 'qc'))
 
+   # Total cloud cover under ECHAM's maximum-random overlap (its own
```
duncanwp: None of this belongs in getting started, it's not useful or relevant to the average user who should be looking at the average of the cloud cover the radiation sees 95% of the time

duncanwp: Agreed and reverted in 1e42dc88 — `getting_started.rst` is now byte-identical to `dev` on this branch. Nothing was lost by removing it: `docs/source/api.rst` autodocs `jcm.analysis`, so `total_cloud_cover`'s docstring is already its reference, and the provenance, the measured spread between overlap definitions and the older-output caveat are all in `design/cloud_cover_gate.md`.

On the root cause, since this is the third time it has happened: CLAUDE.md's doc rule said user-facing changes "must be reflected in README.md and/or getting_started.rst" without bounding how much, so "new public function" kept reading as "user-facing change". e970ce62 adds the missing half — the guide orients a new user rather than serving as a reference, a new function is not by itself a guide entry because api.rst autodocs the modules, and rationale/provenance/numbers go to a design doc. That commit is separable if you would rather take it on its own or word it differently.

On the substance of your last point, that the average user wants the cover the radiation sees: `radiation.total_cloud_cover` is printed by the health check but is deliberately not the gated number, for three reasons worth your eye. It only exists in output written after b772ffec (2026-07-31), so it is absent from every archived year including the two scored for this PR. It is identically zero on grey and NN-emulator radiation, which is what the SPEEDY member runs. And on the one run carrying both, it reads 0.757 against max-random 0.535, because saved `cloud_fraction` is an output-interval mean while the McICA cover is sampled per step, so the two are not the same quantity averaged differently.

If you would rather the gate follow the radiation view where it exists and fall back to `aclcov` otherwise, that is a small change to `cloud_cover_fields` and its band, and I am happy to make it — it just needs its own band derived, since the 0.757-vs-0.535 gap means the 0.5-0.9 proposal here would not transfer.

Issue #831, opened 2026-09-16
v3.0 release readiness: physics, API, documentation, and validation


duncanwp:
## Goal

Ship v3.0.0 with internally consistent physics and APIs, accurate documentation, and release gates that fail when the tree is not ready. This is intentionally a major-version release: incompatible API changes are acceptable, but they must be deliberate, documented, and covered by a v2-to-v3 migration path.

This tracker consolidates the release review of `dev` at `c868f7b2ccd8` against `origin/main` (merge base `06fd4771bccf6f8df3a1f032ca0c6e4e7957cf2d`) and the current GitHub backlog.

**Release rule:** every unchecked item below must either be completed or explicitly moved to the “accepted/documented limitation” section with a maintainer decision recorded here. Closing an issue as deferred without documenting the shipped behavior does not satisfy the gate.

## Physics and scientific correctness — resolve before tagging

- [x] #666 — choose and enforce one `PhysicsState.specific_humidity` unit; restore correct moist coupling in the hybrid dycore and keep SPEEDY/pySES conversions consistent.
- [x] #749 — resolve the `ChemistryData` ppmv/ppbv mismatch and audit affected radiation inputs.
- [ ] #782 — validate or correct the Tiedtke-trigger climate shift in the ECHAM 1M/2M configurations.
- [ ] #787 — use the intended HAM/ECHAM soil-moisture quantity for the dust saturation gate.
- [ ] #808 — retune/validate Tegen dust emissions against the wind distribution and document the target budget.
- [ ] #806 — make tendency limiting water-conservative or replace it with a conservative limiter.
- [ ] #807 — apply sea-ice concentration consistently in SPEEDY turbulent surface fluxes.
- [ ] #812 — resolve and validate the convective rain-cover assumption used by sub-cloud evaporation.

Resolved during the audit window:

- [x] #752 — SPEEDY shortwave heating is now cached/replayed between radiation steps.
- [x] #790 — JAM hygroscopic water volume now uses the correct lognormal third moment.

## Public API and internal consistency

- [x] #830 — make `vertical_interp_log_p` honor its public arbitrary-dimensional contract, or deliberately narrow/private the helper and test the exact contract.
- [x] #731 — establish a checkpoint compatibility/migration policy before the release changes more diagnostic pytrees.
- [x] #772 — make `set_constants` effective across physics modules or remove/narrow the public promise.

Already implemented on `dev` and closed as part of this audit:

[...]

PR #832, opened 2026-09-17
Release gates: enforce coverage floors, live constants in JAM/tropopause, general vertical_interp_log_p (#830, #772, #786)

### LINKED ISSUE #772: Import-time constant binding defeats set_constants in ~12 physics modules
The v2 constants contract (`jcm/constants.py`, CLAUDE.md) is that consumers read constants by attribute access (`import jcm.constants as c; c.grav`) so `set_constants(...)` overrides (other planets, sensitivity studies, gradient calibration) reach every consumer. Six non-test modules still capture constants at import time, so after an override they silently compute with the OLD values while dynamics and attribute-access consumers use the new ones — a mixed constant set with no error.

Two capture mechanisms, same effect:
- **Value bind** — `from jcm.constants import grav as _G`: the number is baked in.
- **Stale singleton reference** — `from jcm.constants import physical_constants`: `set_constants` REBINDS the module global (`PhysicalConstants` is a NamedTuple, so it cannot mutate in place), leaving the imported reference pointing at the old object.

Sites:

| file | binds | mechanism |
|---|---|---|
| `physics/aerosol/jam/activation/arg.py` | 11 | values |
| `physics/aerosol/jam/drydep/resistances.py` | 4 | values |
| `physics/aerosol/jam/sedimentation/sedi_term.py` | 3 | values |
| `physics/aerosol/jam/ice_nucleation/ice_term.py` | 2 | values |
| `physics/aerosol/jam/emissions/dust.py` | 1 | values |
| `physics/diagnostics/wmo_tropopause.py` | 1 | singleton reference |

NOT affected (verified): `dycore/dinosaur/dycore.py` imports only the `PhysicalConstants` class and reads the live singleton at construction (dycore.py:211) — its module-level `PHYSICS_SPECS` snapshot is referenced only by tests. Hines, Lott-Miller, `lohmann_2m_params` and `moist_air_state` use the compliant module-alias access.

Fix: convert each to attribute access, or where a value is baked into a precomputed table/closure, recompute it at construction. Add a guard test that greps for `^from jcm.constants import` allowing only `PhysicalConstants` (the class), or monkeypatches `set_constants` and asserts a changed `grav` propagates into each term's output.

Found while writing the science model-description (#641, PR #771); the constants section documents this as a known limitation until fixed.

### LINKED ISSUE #786: CI: the slow-test coverage floor (80%) is not enforced — dev is at 79.68% and the job passes
The PR-only slow-test job runs

```
pytest -v -s -m "slow" --cov=jcm --cov-config=.coveragerc-pr --cov-fail-under=80
```

(`.github/workflows/run_test.yaml`, step "Test with pytest (include slow tests on PR)"). On current `dev` that step prints

```
FAIL Required test coverage of 80% not reached. Total coverage: 79.68%
== 92 passed, 29 skipped, 2409 deselected ... ==
```

and the job is still reported **success** — e.g. run 34404457174 (PR #769, docs-only, 2026-09-09T21:36Z). The same happens locally in a fresh venv built from `requirements.txt` (pytest 9.1.1, pytest-cov 7.1.0, `-n 4`): the FAIL line is printed and pytest exits 0. So the 80 % slow floor documented in `CLAUDE.md` / the `jcm-local-ci` skill is not currently enforced, and `dev` sits 0.3 % under it without anyone being told.

Two separable parts:

1. **Enforcement.** Establish why `--cov-fail-under` no longer sets a non-zero exit status in this pytest/pytest-cov/xdist combination (it did when the floor was introduced — see the `.coveragerc-pr` comment about the aerocom fixes "tipping the slow shard from 80 to 79", which was a red job at the time). A robust alternative that does not depend on plugin behaviour is a separate `coverage report --rcfile=.coveragerc-pr --fail-under=80` step after pytest. Whichever is chosen, the fast job (`--cov-fail-under=90`) should be checked the same way.
2. **The number.** Once enforced, `dev` fails by 0.3 %. Decide whether that is a genuine coverage gap in main-path slow-test code or another entry-point-shim module that belongs in the `.coveragerc-pr` omit list (the file documents that pattern and its rationale).

Found while running the local gate for the dinosaur-1.5.0 pin PR; unrelated to that change (`.coveragerc-pr` omits `dycore.py` from the slow denominator, and the number is identical with and without the PR).

### LINKED ISSUE #830: vertical_interp_log_p is public but only accepts exactly 4-D inputs
## Summary

`jcm.data.bc.interpolate_ozone.vertical_interp_log_p()` is documented as a public, field-generic pressure interpolation helper, but its implementation only accepts exactly four-dimensional `(time, pressure, lat, lon)` input.

Verified on `dev` at `c868f7b2ccd8`.

## Evidence

The docstring at `jcm/data/bc/interpolate_ozone.py:42-56` says the source may have shape `(..., nplev_source, ..., ...)`, calls the helper public because it applies to temperature, humidity, and winds too, and only fixes the pressure axis as axis 1.

The implementation then unpacks exactly four dimensions and indexes four axes:

```python
ntime, _, nlat, nlon = o3_source.shape
...
out[t, :, j, i] = np.interp(..., o3_source[t, :, j, i])
```

A valid three-dimensional `(time, pressure, column)` field or a five-dimensional field with an extra ensemble dimension therefore fails despite satisfying the documented contract.

## Acceptance criteria

Choose and test one explicit contract before the release:

- [ ] **General API:** interpolate axis 1 for arbitrary leading/trailing dimensions (for example by moving/flattening the non-pressure axes), preserving dtype and supporting ascending or descending source pressure; add 3-D, 4-D, and 5-D tests.

or

- [ ] **Ozone-file helper:** make the function private or rename/narrow it, document that only `(time, pressure, lat, lon)` is accepted, and validate the shape with a clear error.

In either case, validate that the source pressure length matches axis 1 and retain a regression test for descending pressure levels.

Found during the v3.0 release review against `origin/main`.

duncanwp:
Three v3.0 release-gate issues from #831. The commits are separable and can be
reviewed independently.

---

## #786 — the coverage floors did not fail below the floor

**Root cause (confirmed against the installed sources, not guessed).**
`coverage.results.should_fail_under` — which pytest-cov *and* `coverage report`
both call — returns `round(total, precision) < fail_under`. At coverage's
default `precision = 0`, 79.68 rounds to 80, so the check passed: **the 80
floor was really a 79.5 floor, and the 90 fast floor really an 89.5 floor.**
The `FAIL Required test coverage...` banner comes from a *different* pytest-cov
code path (`pytest_terminal_summary`) which compares the raw float — hence a
job that printed FAIL and exited 0.

This was never an xdist or output-capture effect: a synthetic package at
79.80% exits 0 identically with and without `-n`, and with and without `-s`.
Nor was the floor broken when introduced — **pytest-cov 6.0.0** changed it:
*"Changed fail under checks to use the precision set in the coverage
configuration. Now it will perform the check just like `coverage report`
would."* That silently disarmed a floor which had been enforced on the raw
value before.

Versions here: **pytest 9.0.2, pytest-cov 7.1.0, pytest-xdist 3.8.0,
coverage 7.13.5.**

**Fix.** `[report] precision = 2` in both `.coveragerc` and `.coveragerc-pr`.
Note the standalone `coverage report --fail-under=80` step the issue suggests
would **not** have fixed this on its own — it shares `should_fail_under` and
also exits 0 at 79.68 unless the precision is set. Both are shipped: the
precision is the fix, and the extra `coverage report` step in each CI job is a
belt whose exit status does not depend on plugin behaviour that has already
changed under us once. It re-reads the `.coverage` file pytest-cov has just
combined across the xdist workers (verified to give the identical total), so it
re-measures nothing and costs about a second.

**This PR proved it on itself.** Its first slow-test run measured 79.89% and
the job went **red** — on `dev` today that same number prints `FAIL` and exits
0. Same data, same floor, precision the only difference:

```
coverage report --rcfile=.coveragerc-pr --fail-under=80                → 79.89%  exit 2  ✅ (this branch)
coverage report --rcfile=.coveragerc-pr --precision=0 --fail-under=80  →    80%  exit 0  ❌ (dev today)
```

**The number: omit, and the enforcement change is what found it.** With the
floor actually enforcing, the PR job failed at **79.89%** — the first honest
reading anyone has had of this number. (The issue's 79.68% was measured before
~40 merges; the agreement is a coincidence.)

My first local measurement said 83.78% and was wrong for the gate, in exactly
the way CLAUDE.md's **CI-dependency-parity** note warns about and PRs #582/#598
were both bitten by: CI installs `requirements.txt` + `pip install -e .` with
**no extras**, my workstation has the extras installed, and the whole 4-point
difference is two families of files:

| omitted | statements | missed in CI | extra |
|---|---|---|---|
| `jcm/dycore/pyses/*` | 639 | 523 | `jcm[pyses]` |
| `jcm/physics/aerosol/jam/microphysics/mam4_jax.py` | 168 | 168 | `jcm[mam4]` |

Every test for these `importorskip`s out in CI, so the code is **unreachable**
there, not under-exercised — and `mam4-jax` is GPL-3.0, kept out of the
Apache-2.0 core on purpose, so CI will *never* install it. Left in the
denominator, the 80% gate measures which extras the runner happens to have and
moves depending on whose machine ran it. That is what `.coveragerc-pr`
documents itself as existing for, and a stronger case than its existing
entries, which are about code the slow suite merely reaches through
entry-point shims.

Recomputed from the failing run's own per-file table: **15804 statements, 2649
missed, 83.24%** — clear of the floor with **no test changes**, because no
integration coverage is lost. Only two fully-uncovered modules remain in CI
(`data/bc/compile.py`, `diagnostics/omega.py`, 67 statements between them);
both are reachable in principle, so they stay in.

I chose **omit over a gap-fix** because there is no gap to fix: a test written
to cover `mam4_jax.py` would skip in CI too. The alternative — installing the
extras in CI — is not available for `mam4-jax` on licensing grounds and would
add substantial runtime for `pyses`.

The base `.coveragerc` is deliberately **left alone**: CI's fast suite reads
**92.46%**, so its 90 floor is met as-is and loosening a gate that is not
failing would be unjustified. The asymmetry matches the file's existing
PR-only entries (`utils.py`, `runners.py`).

`CLAUDE.md`, the `jcm-local-ci` and `jcm-dev-workflow` skills, `local_ci.sh` and
`developer.rst` all gain the second step, with the reason recorded so nobody
later deletes the "redundant" line.

---

## #772 — constants frozen at import defeated `set_constants`

The contract is that a constant is read **when its value is used**, so
`set_constants` reaches every consumer. Modules were breaking it in **three**
ways, all silent, leaving a process computing with a mix of overridden and
Earth values:

1. **Value import** — `from jcm.constants import grav as _G`.
2. **Singleton reference** — `from jcm.constants import physical_constants`;
   equally stale, because `set_constants` rebinds the module global rather than
   mutating it (`PhysicalConstants` is a `NamedTuple`).
3. **Import-time attribute read despite the approved alias** — a derived module
   constant (`_MW_AIR = c.m_air * 1000.0`), or a default argument
   (`def f(..., gravity=c.grav)`, evaluated once when the `def` executes).

The issue listed only the first two. The third came out of the review loop and
is the one worth flagging to you: the TTE-TKE closure reads `c.cpd` **live** and
took gravity from a **frozen** default, in the same expression
(`lapse_rate = gravity / c.cpd`). Nine modules in total:

| module | captures | form |
|---|---|---|
| `aerosol/jam/activation/arg.py` | 11 | value import |
| `aerosol/jam/drydep/resistances.py` | 4 | value import |
| `aerosol/jam/sedimentation/sedi_term.py` | 3 | value import |
| `aerosol/jam/ice_nucleation/ice_term.py` | 2 | value import |
| `diagnostics/wmo_tropopause.py` | 1 | singleton reference |
| `aerosol/jam/chemistry/aqueous.py` | 3 | derived module constants |
| `vertical_diffusion/tte_tke/turbulence_coefficients.py` | 1 | default argument |
| `vertical_diffusion/tte_tke/tke_budget.py` | 1 | default argument |
| `data/emissions/prepare.py` | 1 | singleton ref via alias |

**`emissions/dust.py`, listed in the issue, needed no change** — the Tegen port
in #804 removed its `jcm.constants` use entirely.

Derived constants became small functions called at use (trace time, so free);
default arguments became `None` and resolve to `c.grav` in the body, which keeps
the published signatures working for callers that pass gravity explicitly.
`arg.py`'s eleven short aliases are kept for readability but bound *inside*
`arg_activation`, per call; the formulae are untouched.

**Guard:** `jcm/constants_test.py` parses every non-test module under `jcm/`
with `ast` (not a regex — it handles parenthesised imports and tells a module
body from a function body for free) and rejects all three forms. It inspects
module bodies, class bodies, and a `def`'s or `lambda`'s decorators and
default/keyword-default expressions, while allowing function and lambda
*bodies*, which run at call time — the legitimate pattern `utils.py` and
`state_bridge.py` rely on. Only `PhysicalConstants` may be imported by name; it
is a type and binds no value. Meta-tests pin every accepted and rejected form.
Alongside are behavioural checks that each converted module's output actually
moves when gravity is overridden — **all six of the original tests fail on the
pre-change code**, so the guard is not vacuous. Every one restores the constants
in a `finally`; `set_constants` is process-global and an escaped override would
corrupt later tests in the same xdist worker.

**Caveats, documented rather than hidden:**
- Constants internal to **mam4-jax** belong to that package; `set_constants`
  does not reach them.
- A constant read inside a **jitted** term is fixed when that term is traced, so
  the existing "override before constructing the model" rule still holds. The
  TTE-TKE behavioural test clears the JAX cache to assert exactly that, since
  the contract is about a model *built after* the override.

Documented in `release_notes.rst` (a single self-contained bullet, so the
expected conflict with #824 is mechanical) and in
`docs/source/science/constants.md`, whose "known limitation" paragraph is
rewritten as the behaviour that now holds. **That rewrite is required, not
cosmetic:** `docs/science_pointers_test.py` enforces that every `#NNN` in the
register points at an *open* issue, so leaving `#772` there would turn `dev` red
the moment this merges.

---

## #830 — `vertical_interp_log_p` only accepted exactly 4-D input

**Decision: the general API, not the narrowing.** #767 (commit `33077a48`)
deliberately made this public with an explicit field-generic claim; narrowing it
would retract a public API days after publishing it, so this honours the merged
intent.

Rank generality comes from moving axis 1 to the end and flattening everything
else, so no axis but the pressure one is ever named. The per-column `np.interp`
loop is **kept rather than vectorised**: it is offline, once-only code, and
keeping it makes the result **bit-identical** to the previous implementation on
4-D input — pinned by a test against the old code inline, so the shipped
boundary-condition files cannot move. Also added: validation of the level count
against axis 1, of `plev_source` being 1-D, and of **strict monotonicity** —
that last one because `np.interp` returns silent nonsense for non-monotonic
`xp` rather than failing. Descending source pressure is still flipped
(`np.flip(..., axis=1)`), and the `Returns` docstring no longer claims a fixed
`(time, nplev_target, lat, lon)` shape it never guaranteed.

The first parameter **keeps its published `o3_source` name** — an earlier commit
renamed it to `source` and that was reverted during review, since the rank
generalisation does not need it and a keyword caller would get `TypeError`.
Nothing here is an API break.

Tests cover 2-D, 3-D `(time, p, column)`, 4-D, 5-D `(ensemble, p, time, lat,
lon)`, a descending-pressure regression, exact-node recovery, dtype
preservation, each error path, and the 4-D bit-identity.

---

## Validation

**In CI** (authoritative — this is what the gate measures):

| gate | coverage | floor |
|---|---|---|
| fast (`-m "not slow"`) | **92.46%** | 90 ✅ |
| slow (`-m slow`), before the omit | **79.89%** | 80 ❌ — correctly failed, which is the fix working |
| slow (`-m slow`), after the omit | **83.24%** | 80 ✅ |

**Locally** (`JAX_PLATFORMS=cpu -n 12`, `dinosaur 1.5.0`, extras installed so
the coverage reads high — the suites themselves are the point here):

| gate | result | coverage |
|---|---|---|
| fast (`-m "not slow"`) | **2757 passed, 7 skipped, 0 failed** | 95.79% |
| slow (`-m slow`) | **104 passed, 24 skipped, 0 failed** | 83.78% |

- `ruff check .` — clean.
- `docs/science_pointers_test.py` (the rot guard CI runs on PRs) — 9 tests, OK.
- Three review findings from Codex were all real and all fixed; see the PR
  comments for the per-finding reasoning.

Closes #830
Closes #772
Closes #786
Refs #831


duncanwp:
Both Codex findings were real; both are fixed in `8b8553ef` and `d27e7125`.

**P1 — "Detect all import-time constant captures" (`jcm/constants_test.py:84`). Correct, and the more important of the two.** My guard only rejected `ImportFrom`, so it missed the form where a module uses the approved `import jcm.constants as c` alias and simply reads the attribute too early. The contract is about *when* a constant is read, not which import statement was written.

Rather than fix the three sites named, I enumerated the category — every import-time evaluation of an attribute on the constants module, in any syntactic position — and found **six sites in four modules**:

| site | form |
|---|---|
| `chemistry/aqueous.py:72,74,82` | derived module constants (`_ZRGAS`, `_AVOGADRO`/`_AVO_XTOC`, `_MW_AIR`) |
| `tte_tke/turbulence_coefficients.py:22` | `gravity: float = c.grav` default argument |
| `tte_tke/tke_budget.py:76` | same |
| `data/emissions/prepare.py:42` | `_const.physical_constants.avogadro` — stale singleton via a module alias |

The last one you did not name; it came from enumerating the category rather than grepping the reported spellings. The TTE-TKE pair is the sharpest illustration of why the import form is the wrong thing to check: both functions go on to compute `lapse_rate = gravity / c.cpd`, reading cp **live** and g **frozen**, in the same expression.

Fixes: the derived constants became small functions called at use (trace time, so free); the default arguments became `None` and resolve to `c.grav` in the body, which keeps the published signature working for callers that pass gravity explicitly.

The guard now rejects any `c.<name>` evaluated at import — module and class bodies, plus a `def`s decorators and default/keyword-default expressions — while still allowing function and lambda *bodies*, which run at call time (`utils.py` and `state_bridge.py` rely on that). It also distinguishes a name bound to the module from one bound to the package, so a bare `import jcm.constants` reports `jcm.constants.grav` once rather than also flagging the intermediate module attribute. Meta-tests pin each accepted and rejected form.

Two behavioural tests were added. The TTE-TKE one clears the JAX cache before re-evaluating, which surfaced something worth stating explicitly: the function is jitted, so its trace fixes the constants, and what the contract actually promises is that a model **built after** `set_constants` sees the new value. That caveat is now in `docs/source/science/constants.md` and the release note rather than left implicit — as is the widened list of affected modules.

**P2 — "Preserve the public interpolation keyword" (`interpolate_ozone.py:40`). Agreed, reverted.** The rename was flagged in the PR description as a judgement call; you are right that the rank generalisation does not require it and that a keyword caller would get `TypeError`. `o3_source` stays, with a docstring note saying why an ozone-flavoured name survives on a deliberately field-generic function.

Validation on the updated branch: `ruff check .` clean; the directly affected suites (`constants_test`, `interpolate_ozone_test`, `aqueous_test`, all of `tte_tke/`, `data/emissions/`) are **121 passed, 1 skipped**. Full fast and slow gates are re-running locally and I will post the coverage numbers when they land.

@codex review

duncanwp:
Fixed in `4946e125`.

**P2 — "Inspect lambda defaults before skipping lambda bodies". Correct.** I had skipped the `Lambda` node whole, reasoning that its body runs at call time. That is true of the body and false of the defaults: `lambda g=c.grav: g` evaluates `c.grav` when the lambda object is created, which at module level is import — the same freeze as a `def` default, with the one expression that is *not* lazy left unchecked. Ironic given the commit it was in.

A lambda is now walked exactly like a `def`: decorators (n/a), defaults and keyword defaults are inspected; the body is not. A meta-test pins both halves — `frozen = lambda g=c.grav: g` is flagged, `live = lambda x: x * c.cpd` is not.

Worth noting for the record: **no such site exists in the package today** — the repo-wide guard passes unchanged before and after — so this closes the hole prospectively rather than fixing a live defect. That also means the six captures fixed in `8b8553ef` remain the complete set.

Local: `ruff check .` clean, `jcm/constants_test.py` 13 passed. Full fast and slow gates are re-running; numbers to follow.

duncanwp:
**Local gate results on the final branch** (`4946e125`), both at CI dependency parity (`dinosaur 1.5.0`), `JAX_PLATFORMS=cpu -n 12`:

| gate | result | `--cov-fail-under` exit | `coverage report --fail-under` exit | coverage | floor |
|---|---|---|---|---|---|
| fast (`-m "not slow"`, `.coveragerc`) | **2757 passed, 7 skipped, 0 failed** | 0 | 0 | **95.79%** | 90 |
| slow (`-m slow`, `.coveragerc-pr`) | **104 passed, 24 skipped, 0 failed** | 0 | 0 | **83.78%** | 80 |

`ruff check .` clean. The rot guard (`docs/science_pointers_test.py`) passes — worth stating explicitly because rewriting the `#772` limitation paragraph in `docs/source/science/constants.md` was *required*: that guard fails on a `#NNN` pointing at a closed issue, so leaving the reference there would have turned `dev` red the moment this merged.

Both floors now enforce. Same data, same floor, precision the only difference:

```
coverage report --rcfile=.coveragerc-pr --fail-under=84                → 83.78%  exit 2  ✅
coverage report --rcfile=.coveragerc-pr --precision=0 --fail-under=84  →    84%  exit 0  ❌  (the #786 bug)
```

Two notes on the numbers. The slow suite measured **83.77%** on unmodified `dev` with the identical harness, so this PR moves it +0.01 — the issue’s 79.68% is simply stale, predating ~40 merges, which is why **no gap-fix and no omit were needed**. And I deliberately added nothing to the `.coveragerc-pr` omit list: padding the denominator while the suite is 3.8 points clear of the floor would only hide future regressions.

One caveat stated for accuracy: the fast-gate run above collected `jcm/constants_test.py` as it stood *before* the lambda-defaults fix, since that edit landed mid-run. Test files are omitted from coverage so the 95.79% is unaffected, and `jcm/constants_test.py` at `4946e125` was run separately: **13 passed**.

duncanwp:
**Correction and resolution on the #786 number — the enforcement change did its job on its own PR.**

The PR build failed at **79.89%**, and that failure is the fix working: before this branch, that exact number printed `FAIL` and exited 0. It is also the first honest reading of the slow-suite coverage anyone has had.

It means my earlier comment quoting **83.78%** was measuring the wrong thing, in precisely the way `CLAUDE.md` warns about and PRs #582/#598 were both bitten by: **CI installs `requirements.txt` + `pip install -e .` with no extras**, my workstation has the extras installed, and the entire 4-point gap is two families of files:

| file(s) | stmts | missed in CI | gated behind |
|---|---|---|---|
| `jcm/dycore/pyses/*` | 639 | 523 | `jcm[pyses]` |
| `jcm/physics/aerosol/jam/microphysics/mam4_jax.py` | 168 | 168 | `jcm[mam4]` |

Every test for these `importorskip`s out in CI, so the code is **unreachable** there rather than under-exercised — and `mam4-jax` is GPL-3.0, kept out of the Apache-2.0 core deliberately, so CI will never install it.

**Decision: omit, not a gap-fix** (`dab131f0`), because there is no gap a test could close — a test written for `mam4_jax.py` would skip in CI too. Recomputed from the failing run’s own per-file table: **15804 statements, 2649 missed → 83.24%**, clear of the floor with **no test changes and no integration coverage lost**. Only two fully-uncovered modules remain (`data/bc/compile.py`, `diagnostics/omega.py`, 67 statements total); both are reachable in principle, so they stay in the denominator.

The base `.coveragerc` is deliberately untouched: CI’s fast suite reads **92.46%**, so the 90 floor is met as-is and loosening a gate that is not failing would be unjustified. That PR-only asymmetry matches the file’s existing entries (`utils.py`, `runners.py`).

I have verified the omit patterns actually match (a fresh measurement drops both families from the report) rather than assuming — worth checking because `omit` under `[run]` applies at *measurement* time, so it has no effect on a re-report of an already-collected data file.

Re-running CI now; I expect the slow job green at ~83.2%.

duncanwp:
## CI is green, and #786 is proven end-to-end on this PR

`a89eb732` — all five checks pass.

**The slow job, which failed at 79.89% two runs ago, now reads 83.24% and both enforcement paths agree on the same `.coverage` file:**

```
Test with pytest (include slow tests on PR)
  Required test coverage of 80% reached. Total coverage: 83.24%
  == 98 passed, 30 skipped, 2726 deselected in 3070.76s ==

Enforce slow-test coverage floor (PR)          <- the new belt step
  TOTAL                    15804   2649  83.24%
```

**The fast job likewise:**

```
Test with pytest (no slow tests on push)
  Required test coverage of 90% reached. Total coverage: 92.53%

Enforce fast-test coverage floor (push)        <- the new belt step
  TOTAL                    19877   1485  92.53%
```

Three things worth noting from that output:

1. **The belt steps really are wired in**, not just present in the YAML — each appears as its own step and re-reports the identical total from the data file pytest-cov combined across the xdist workers. Neither re-measures anything.
2. **83.24% is exactly the figure I projected** from the failing run’s per-file table (15804 statements, 2649 missed), so the omit patterns matched precisely what was intended — `dycore/pyses/*` and `mam4_jax.py` appear zero times in the report.
3. **The whole loop is self-demonstrating:** this PR’s own first run is the evidence that the floor previously did not enforce, and its last run is the evidence that it now does.

### Review state

Three Codex findings, all real, all fixed and replied to individually:
- **P1** import-time attribute capture → six sites in four modules (I found three beyond the three reported), guard extended to cover timing rather than import form.
- **P2** `o3_source` keyword → rename reverted; **no API change ships**.
- **P2** lambda defaults → guard now walks them; no such site existed, so it closes the hole prospectively.

Since that last review I also landed the `.coveragerc-pr` omit decision and a docstring/doc note scoping the guard’s one blind spot (a module-level *call* that reads constants internally; `dycore.PHYSICS_SPECS` is the only instance and is safe by construction — built from `PhysicalConstants.default()`, referenced only by tests). Those are substantive enough to deserve a fresh pass before human review.

@codex review

duncanwp:
Fixed in `a8c3c0c5`.

**P2 — "Detect relative constants module aliases". Correct.** `_constants_aliases` matched only the absolute `from jcm import constants`, so `from . import constants as c` bound the same module and the subsequent `c.grav` read at import went unreported. The asymmetry is the real bug: the *value-import* path already resolved relative imports through `_resolve_import`, the *attribute* path did not — and a guard that is green for one spelling and not another is worse than no guard, because it looks like coverage.

Both paths now share `_resolve_import`, so a relative alias counts exactly as the absolute one does at any depth. The new test pins three cases, including the negative one that matters: `from . import constants as c` inside `jcm.physics.thing` resolves to `jcm.physics.constants`, a different module, and is correctly **not** flagged.

No such site exists in the package today — the repo-wide guard is unchanged before and after — so this, like the lambda fix, closes a hole prospectively. The six captures fixed in `8b8553ef` remain the complete set.

`ruff check .` clean; `jcm/constants_test.py` 14 passed. This touches only the test module, so the coverage numbers from the green run above are unaffected.

That is all three review rounds addressed — thanks, all four findings across them were real.

Issue #820, opened 2026-09-15
check_gradients' absolute finite-difference step is blind to large-magnitude input leaves


duncanwp:
## What is missing

`jcm.testing.check_gradients(reference="difference")` walks a ladder of
**absolute** steps (`DEFAULT_STEPS = 1e-3 * 0.5**k`, `jcm/testing.py`) applied
to a tangent from `random_direction`. For an input leaf whose values are large,
that step is below a float32 ulp and the perturbation rounds away entirely, so
the secant carries no information about that leaf at all.

Concretely, in SPEEDY's vertical diffusion the dry static energy is
`se ~ 3.1e5`, whose float32 ulp is ~0.031 — 30x the *largest* step on the
ladder. Wrapping `physics_data.convection.se` (or `state.geopotential`) in
`jax.lax.stop_gradient` inside `get_vertical_diffusion_tend` leaves the
`reference="difference"` check passing untouched. It compounds with a second
effect: the projection is a single inner product, so even a leaf that *is*
resolvable can be swamped — at seed 0 the sigma-grid metrics contribute -53.8
of a -62.5 total while `se` contributes -1.6e-4, about 3e-6 of it.

The output side of this problem is already solved: `_cotangent` scales each
output leaf's random cotangent by the inverse RMS of that leaf's primal, "so a
gradient zeroed on any small field would not move the inner product". The
input side has no equivalent — `random_direction` seeds per leaf-name but does
not scale by magnitude.

## Where the hooks already are

- `jcm/testing.py` — `DEFAULT_STEPS`, `random_direction`, `_finite_difference`,
  and the `_cotangent` docstring that already argues the case for the output
  side.
- `jcm/physics/vertical_diffusion/speedy_vdiff_test.py::test_moisture_branch_input_sensitivities`
  — works around it for one scheme by asserting a per-input-leaf liveness
  table directly, and points here.

## What is needed

Most likely a **relative** perturbation: scale each leaf's tangent by that
leaf's RMS (the mirror of what `_cotangent` does for outputs), so the step is
meaningful whatever the leaf's magnitude. That changes the effective step, and
so the convergence behaviour and tolerance, of **every** `check_gradients`
caller in the repo, which is why it is not being folded into #814's PR: it
needs its own sweep over the existing gradient checks to confirm none of them
was passing only because of the current scaling.

A per-input-leaf liveness guard (the mirror of the existing per-output-leaf
[...]

PR #845, opened 2026-09-19
SPEEDY vertical diffusion: cover the moisture branches against the reference Fortran (#814)

### LINKED ISSUE #814: SPEEDY vertical diffusion: the moisture branch is never exercised by any test
## What is missing

`get_vertical_diffusion_tend` returns a temperature tendency and a specific
humidity tendency. The moisture branch is gated on

```python
drh  = rh[kx - 1] - rh[nl1 - 1]
drh0 = parameters.vertical_diffusion.rhgrad * (fsg[kx - 1] - fsg[nl1 - 1])
```

opening only when `drh > drh0`. Every test that touches the scheme builds its
state with `PhysicsData.ones(...)` / `PhysicsState.ones(...)`, which makes `rh`
uniform, so `drh` is exactly `0` and the gate never opens. The moisture half of
the scheme therefore has **no** test coverage at all — not the gradient check,
and not the value tests either.

This surfaced while converting the gradient checks: the new per-output-leaf
guard in `jcm.testing.check_gradients(reference="adjoint")` reported the
specific-humidity tendency as carrying an identically zero gradient, which is
correct and previously invisible because the whole tendency struct was
contracted into one projection.

## Where the hooks already are

- `jcm/physics/vertical_diffusion/speedy_vdiff.py` — the gate, and the
  `rh_gate_smoothing` crossfade that already exists to make the branch
  differentiable (so somebody intended it to be exercised).
- `jcm/physics/vertical_diffusion/speedy_vdiff_test.py::test_get_vertical_diffusion_gradient_check`
  — asserts the dead components explicitly and points here; adding a realistic
  profile is the fix.

## What is needed

A state with a genuine vertical relative-humidity gradient — a moist boundary
layer under a drier free troposphere, i.e. `rh[surface] - rh[nl1] > drh0` — and
then both a value test against the Fortran and a gradient check. I did not
invent one in the converting PR: `drh0` depends on `rhgrad` and the sigma
spacing, so the profile needs to be chosen against `vdifsc.f90` rather than
picked to open the gate.

## Reference formulation

SPEEDY `vdifsc.f90`, the `drh > drh0` shallow-convection/moisture-diffusion
branch.

duncanwp:
Closes #814.

## What was missing

`get_vertical_diffusion_tend` returns a temperature *and* a specific-humidity
tendency, but nothing had ever made the moisture half non-zero under a gradient
check: every gradient test builds its state with `PhysicsData.ones()`, which
makes `rh` uniform, so `drh = rh[surface] - rh[nl1]` is exactly `0` and no
moisture gate opens. #816's per-output-leaf guard is what made that visible, as
an identically zero specific-humidity gradient.

The gap turned out to be wider than the gradient check. `vdifsc.f90` splits the
lowest interface between two complementary branches:

| branch | condition | what it drives |
| --- | --- | --- |
| shallow convection | `dmse >= 0` | dry-static-energy flux, plus a moisture flux if `drh >= 0`; damped by `redshc` under deep convection |
| stable moisture diffusion | `dmse < 0` and `drh > drh0` | moisture flux only |

Every existing test — including the value test, whose `rh = linspace(0.1, 0.9)`
*does* open `drh > drh0` — sits in the **stable** branch. So the whole
`dmse >= 0` side had no coverage of any kind: not its heat flux, not its
moisture flux, and not `redshc`, which is reachable only through it.

## What this adds

Two soundings, one per branch, chosen against the scheme's own gate arithmetic
rather than nudged until a branch opened:

* **Trade cumulus** — a moist boundary layer (RH 0.88) under a drier free
  troposphere with a near-moist-adiabatic lapse rate. Worth noting for the
  issue's framing: built realistically, that profile comes out *conditionally
  unstable* (`dmse = +6.3 kJ/kg`), so "moist BL under dry free troposphere" is
  the **shallow-convection** case, not the diffusion case. Run with and without
  deep convection to reach `redshc`.
* **Stratocumulus** — a cool moist boundary layer (RH 0.90) capped by a
  subsidence inversion (the surface layer is *colder* than the one above) under
  very dry air. That is the stable branch: `dmse = -32 kJ/kg`,
  `drh = 0.30 > 0.0575`. Its deep dry layer also makes step 4 fire.

A sounding sitting far on the open side of a gate pins the flux but not the
**threshold** — deleting the gate outright would not change its answer. So each
gate additionally gets a pair of variants straddling it, made by moving one
layer's RH and nothing else (none of the thresholds depends on `rh`, and the
overrides never touch `qa[-1]` or `qsat[-2]`, so `dmse` and with it the branch
are unchanged):

| gate | threshold | closed / open variants |
| --- | --- | --- |
| shallow moisture | `drh > 0` | `-0.02` / `+0.02` |
| stable diffusion | `drh > 0.0575` | `0.050` / `0.065` |
| step 3, interface 4 | `drh > 0.0875` | `0.085` / `0.10` (unmodified) |
| step 3, `hsg > 0.5` | interface 3 excluded | RH gate thrown open at interface 3 (`drh = 0.15` vs `drh0 = 0.085`), which only `hsg[4] = 0.42` keeps shut |

That last row exists because both soundings leave interface 3 at `drh = 0.05` —
already shut on RH grounds — so the sigma condition did no work and deleting it
changed nothing.

Four tests: values against the reference Fortran for all ten cases; a guard
asserting each case really does sit on the side of its gate it is named for; the
gradient check the issue asked for; and a per-input-leaf sensitivity table
(below).

## Decisions a reviewer might second-guess

**The reference numbers come from the Fortran, and the recipe to redo them is
in the file.** They were produced by compiling SPEEDY's
`source/vertical_diffusion.f90` — pinned to commit `f0a358e9`, digest recorded
and checked — **unmodified** against four stub modules (`types`, `params`,
`physical_constants`, `geometry`) cut down to the names it imports, supplying
`cp = 1004.64` / `alhc = 2501.0` — jcm's values, not SPEEDY's `cp = 1004.0` — so
the comparison isolates the *formulation* rather than the choice of constants.

This matters because the file notes that a `cpd` revision means regenerating (it
has moved once already, 1004.0 → 1004.64). Regenerating from the JAX port under
test would silently turn an independent cross-implementation check into a
self-consistency one, so the comment above `FORTRAN_REFERENCE` carries the full
procedure: the pinned commit and sha256, the stub modules and the names each has
to supply, how `sigh`/`fsg`/`dhs` are filled, and what drives each case.

I originally shipped that as `tools/regenerate_speedy_vdiff_reference.py`;
**removed at the maintainer's request** — this PR is the test harness, not the
harness that produced its numbers. The last commit does the removal rather than
rewriting it out of history, because three commits here are *about* building it
and stripping them would leave their messages describing work that is not in the
diff. Nothing licensed is committed either way: speedy.f90 is
non-commercial-only and requires its notice on substantial portions, neither of
which fits an Apache-2.0 tree — which is also why `fortran_harness/` was deleted
from the tip after serving the ECHAM ports.

Writing it did earn one finding worth keeping: upstream declares
`segrad = 0.1` as a *single-precision* literal widening to `0.10000000149`,
which my first hand-transcribed driver had silently "fixed" to `0.1_p`. Worth
1e-7 in the step-4 tendencies — far inside tolerance, but the committed values
are the faithful ones, and the comment now says to compile rather than
transcribe for exactly that reason.

**The port matches to 1.98e-5 relative**, which is float32 cancellation, not a
formulation difference. What sets the tolerance is step 4: `se0 - se[k]` is
330 J/kg out of dry static energies of 3.1e5, so it loses ~3 decimal digits.
(The `dmse` path is 5x milder at 3.9e-6 and the moisture tendencies 20x at
8.6e-7.) The assertion uses `rtol = 2e-4`, a 10x margin, with `atol = 1e-12` so
a shut gate has to give exactly zero.

**The gradient check compares against a real central difference**
(`rtol = 5e-3`), not the `reference="adjoint"` fallback the uniform-rh check has
to use. Within a branch the fluxes are smooth in every input and the base
soundings sit well away from the gates (nearest margin `drh - drh0 = 0.012`), so
the difference converges: seeds 0-3 agree with AD to 5e-4 on both soundings, and
5e-3 is a 10x margin on the worst of them. The `adjoint` pass still runs first,
for its per-output-leaf liveness guard — a difference contracts the outputs onto
one projection, where a dead specific-humidity gradient (the exact defect this
test exists for) would be invisible.

**The gradient check is paired with a per-input-leaf sensitivity table**, and
this is the part I'd most want a second opinion on. A central difference here
cannot see a destroyed gradient with respect to `se` or `phi`: `check_gradients`
steps by an *absolute* 1e-3, which is below a float32 ulp of `se ~ 3e5`, so the
perturbation rounds away entirely — and the projection is dominated by the
sigma-grid metrics (−53.8 of a −62.5 total at seed 0, against −1.6e-4 for `se`).
`stop_gradient` on `convection.se` or `state.geopotential` passed *both* the
adjoint pass and the `rtol=5e-3` difference. Since the shallow branch this PR
exists to cover is keyed on `dmse = se[-1] - se[-2] + alhc*(qa[-1] - qsat[-2])`,
that is exactly the sensitivity a reader would assume was being checked.

`test_moisture_branch_input_sensitivities` asserts it directly — the mirror of
`check_gradients`' own per-output-leaf guard. All four `stop_gradient` mutations
(`se`, `phi`, `qa`, `rh`) now fail. The **dead** entries carry as much
information as the live ones, and each is explained where declared:

| input | trade cumulus | stratocumulus | why dead |
| --- | --- | --- | --- |
| `se`, `rh`, `qsat` | live | live | — |
| `qa` | live | **dead** | read only through `dmse`; in the stable branch `g_mse` is a step and the SE hinge is flat |
| `phi` | **dead** | live | read only by step 4, which never fires on the stable trade-cumulus column |
| `fsg` | **dead** | **dead** | reaches only the gate thresholds, which have no derivative at zero smoothing width — the differentiability `rh_gate_smoothing` exists to restore |

The underlying blind spot belongs to the step ladder, not to this scheme. Fixing
it properly (scale each leaf's tangent by its RMS, as `_cotangent` already does
for outputs) would move the effective step of **every** gradient check in the
repo and needs its own sweep over them, so it is **#820** rather than folded in
here, cross-linked from the table.

**The gate guard compares strictly (`>`), where `vdifsc.f90` writes `>=`.** The
scheme's width-0 gates are `smooth_gate(x, thr, 0) == (x > thr)`, so the two
differ only for a column landing exactly on a threshold. That is unreachable in
practice, but the guard has to describe what the *scheme* does if it is to
predict it; noted in the test where it is asserted.

**Soundings are frozen literals, not rebuilt in the test.** Each was derived
offline from a temperature profile (geopotential integrated hydrostatically from
`phi_surface = 0`, `qsat` from `speedy_humidity.get_qsat` at 1013 hPa), and that
derivation is documented in the file. Freezing the result keeps the test pinning
`vdifsc` alone, and stops it drifting away from the Fortran reference if the
saturation formula changes. The *expected tendencies*, unlike the inputs, do
scale with `c.cpd` and `alhc` — flagged in the file, so the next constants
revision regenerates them rather than hunting a port bug.

**One case is bit-identical to another, deliberately.**
`stratocumulus_deep_convection` matches `stratocumulus` exactly, because `fcnv`
multiplies only quantities that are zero when `dmse < 0`. That is what it pins:
`vdifsc` sets `fcnv` inside the `dmse >= 0` block only, and applying it to the
stable moisture flux too fails this case and no other new one.

**The port's one deliberate divergence from `vdifsc` is asserted, not covered.**
The Fortran hard-codes step 3's range as `do k = 3, kx-2`; the port replaces the
lower bound with the sigma<0.2 stratosphere mask so it scales with `nlev`. At
`kx = 8` the two coincide — the mask excludes nothing `hsg > 0.5` doesn't
already exclude — so no case here can see the substitution, and deleting it
leaves the file green. The straddle guard asserts that coincidence instead, so
an `nlev` or sigma table where they part company surfaces as the coverage gap it
is rather than as a silent change of behaviour.

**`iptop`, not `icnv`.** The scheme branches on `icnv = kx - iptop`, but the
field a real column carries is `iptop`, which SPEEDY's convection initialises to
the `kx + 1` sentinel and lowers only where it triggers. The helper takes
`deep_convection` and picks the `iptop` the model would actually present.

## Not changed

No production code changes — the port already matches the Fortran on every path
these soundings touch; this closes a test-coverage hole.

Verified by mutation, across three self-review rounds. The base soundings
already caught the structural mutations (index swaps, `rsig`/`rsig1`
off-by-ones, `redshc` dropped or always applied, `dmse` reading `qa` where it
should read `qsat`, branch swap, step-4 sign and condition, `qsat[k]→qsat[k+1]`
in step 3). Six that escaped them are what motivated the straddling variants and
now fail: deleting `g_rh1`, `g_rh2` or `g_rh3` outright, swapping the shallow and
stable thresholds, computing step 2's `drh0` from the wrong `fsg` pair, and
deleting the `hsg > 0.5` condition. Four `stop_gradient` mutations (`se`, `phi`,
`qa`, `rh`) are caught by the sensitivity table; `se` and `phi` passed everything
before it existed.

The third round also independently re-implemented `vdifsc` in double-precision
NumPy and reproduced all 20 reference arrays at **0.00e+00** relative error.

The `rh_gate_smoothing` crossfade is deliberately not re-tested here — it
already has coverage in
`jcm/physics/speedy/speedy_smooth_gradients_test.py::TestVdiffGateSmoothing`.

## Verification

* `jcm/physics/vertical_diffusion` + `jcm/physics/speedy` + `jcm/testing_test.py`
  + `tools`: green (`JAX_PLATFORMS=cpu`, serial — this host has 1 core).
* `ruff check` clean.
* Rebased onto current `dev` after it moved 8 commits (including the
  `physics_interface.py` units standardization in #824); re-verified green.
* The reference was verified byte-for-byte reproducible from the pinned upstream
  commit before the regeneration tooling was removed.

Seven local `/code-review high` rounds went into this before it was pushed. The
last three found no correctness defects — the findings were test-design and
comment-accuracy items, several of which I'd got wrong by reasoning rather than
measuring (the smoothing-width liveness note and the tolerance rationale were
both corrected against measurements).


duncanwp:
**CI green** — `lint`, and both `build (3.11)` jobs (fast at 90% and slow at 80%).

**Codex did not review this.** Its only comment on the PR is that the account has
reached its code-review usage limit, so the automatic pass never ran. In its
place this went through **seven** local `/code-review high` rounds before the
push; the last three found no correctness defects, and the review record is
worth knowing since it changed the PR substantially:

| round | what it found |
| --- | --- |
| 1 | both soundings sat on the open side of every RH gate — deleting `g_rh1`/`g_rh2`/`g_rh3` outright left all cases green. Motivated the straddling variants. |
| 2 | the finite difference could not see a destroyed gradient w.r.t. `se` or `phi` (absolute 1e-3 step, below a float32 ulp of `se ~ 3e5`); `stop_gradient` passed. Motivated the sensitivity table, and issue #820. |
| 3 | the `hsg > 0.5` condition was pinned by nothing; deleting it changed no case. |
| 4 | the frozen reference had no generator, so a constants revision could only be answered from the port under test. Motivated `tools/regenerate_speedy_vdiff_reference.py`. |
| 5 | the generator tracked upstream `master`, and the port's one deliberate divergence from `vdifsc` was unasserted. |
| 6 | the generator hardcoded `kx`, fed the Fortran a different `icnv` than the port, and resolved `jcm` from the installed package rather than the checkout. |
| 7 | six comments that said something untrue about the code beside them. |

Two of round 7's corrections were claims I had reasoned to rather than measured
— the smoothing-width liveness note and the tolerance rationale — and both were
wrong; they are now stated against measurements.

If you would rather wait for a Codex pass before reviewing, it will need the
usage limit lifted first.

duncanwp:
Dropped the reference-regeneration tooling in b370ee36 — this PR is the test harness only. The procedure to reproduce the reference from the pinned upstream Fortran moved into the comment above `FORTRAN_REFERENCE` (commit, sha256, the four stub modules and what each supplies, and what drives each case), since the expected tendencies scale with `cpd`/`alhc` and must not be regenerated from the port under test.

Net diff against dev is now one file: `jcm/physics/vertical_diffusion/speedy_vdiff_test.py`.

@codex review

Issue #846, opened 2026-09-19
Lohmann 2M cirrus nucleation (nic_cirrus=1): ICNC gradient is NaN for ice radius below ~3e-7 m, and the forward candidate is unphysical there


duncanwp:
## What is wrong

`jcm/physics/clouds/lohmann_2m/assembly.py:449`, the default `nic_cirrus=1` branch:

```python
icnc_candidate = 0.75 / (pi * params.rhoice) * air_density * cloud_ice_in_cloud / jnp.maximum(ice_radius_mean**3, params.d_epsilon)
```

Both AD modes weight the denominator by `den**-2`, which float32 evaluates as `1/(den*den)` and returns `inf` once `ice_radius_mean**3` is below ~2**-63, i.e. `ice_radius_mean` below ~3e-7 m. On a cell where `ll2_ic` holds (cold, cloudy, ICNC at or below `icemin` — exactly the nucleating-cirrus cell this branch exists for) the ICNC output's jvp is `nan` **even for a zero tangent**, off a finite forward pass. Verified at `r` in {0, 1e-7, 3e-7} m; clean at 1e-6 m and above. Found by the #820 sweep (the same float32 `den**-2` mechanism fixed in `echam_1m.py::_saturation_adjustment_layer` and `grey_two_stream/planck.py` in that PR).

## Why it is not a guard fix

Raising the floor to the float32-differentiable bound (`_MIN_DIFFERENTIABLE_DENOMINATOR = 1e-18`, see `echam_1m.py`) would make the derivative finite, but the forward value on that branch is already unphysical: `icnc_candidate` is 1.2e11 m^-3 at `r = 1e-7` m and 1.2e20 at `r = 0`, against realistic cirrus ICNC of 1e3 to 1e6 m^-3. A physically motivated minimum ice radius (or a cap on the candidate) is a science choice against the ECHAM-HAM reference (`mo_cloud_micro_2m.f90`, the `nic_cirrus` branch) and needs a `docs/source/science/` update, so it is not folded into the numerics-only #820 fixes.

## Where the hooks are

- `assembly.py:440-449` (`ll2_ic`, the `nic_cirrus == 1` candidate); `params.d_epsilon`, `params.icemin`.
- `jcm/physics/clouds/lohmann_2m_test.py` — the scheme-level `check_gradients` fences added in #820 to extend with a nucleating-cirrus column once the floor is decided.
[...]

Issue #843, opened 2026-09-19
Gradient smoothing: the kinks, jumps and zero-gradient staircases the #820 sweep pinned across SPEEDY and ECHAM physics


duncanwp:
## What is missing

The #820 sweep ran `jcm.testing.check_gradients` (AD against a relative-step central difference) across every SPEEDY and ECHAM scheme. The NaN/infinite-gradient defects it found were fixed in that PR. What remains is a catalogue of places where the physics is **piecewise** — a kink (C0, not C1), a jump, or a staircase whose derivative is identically zero — so that no central difference exists and, more importantly, an optimiser or assimilation system sees either a discontinuous or a zero sensitivity there. Each is pinned by a strict `xfail` or a "gradient is identically zero" fence in the tests, so a fix flips a test visibly. None is a guard-level fix: each needs a differentiable reformulation checked against the reference (ECHAM/ICON/SPEEDY Fortran) and validated in a run.

### Zero-gradient staircases (`argmax`/threshold-selected levels that feed the physics)

- `jcm/physics/vertical_diffusion/tte_tke/turbulence_coefficients.py:392` — `compute_boundary_layer_height` picks the PBL top with `jnp.argmax` over the exchange-coefficient threshold; `pbl_height` sets the mixing length at `:108-111`, so the diffusion's dependence on PBL depth is invisible to a gradient.
- `jcm/physics/convection/speedy_convection.py:98-107` — `get_cloud_top` is an `argmax` over a boolean instability mask; 2844/4608 columns flip at eps=5e-4. Needs a differentiable cloud-top formulation.
- `jcm/physics/convection/tiedtke_nordeng/tiedtke_nordeng.py` — cloud base/top and `ktype` are `argmax`-derived (`:283,429,578,587,678,949`) with integer casts (`:275,679,951,1277`); the tendency ledger is checked with `reference="adjoint"` only.
- `jcm/physics/radiation/mcica.py::generate_subcolumns` — hard Bernoulli draw; gradient to cloud fraction is identically zero (pinned).
- `jcm/physics/gravity_waves/spectral/frontal.py::gw_cm_src` — `frontgf > frontgfc` hard threshold zeroes the gradient to frontogenesis (pinned).
- `jcm/physics/diagnostics/wmo_tropopause.py` — level scan; `d/dT ≡ 0` (pinned, `reference="adjoint"`).

### Kinks and jumps with no central difference

[...]

Issue #842, opened 2026-09-19
Simple GWD deposits no momentum: the launch stress is antiparallel to the wind, so the critical-level filter absorbs it at every level


duncanwp:
## What is wrong

`jcm/physics/gravity_waves/simple/simple_gwd.py::simple_gwd` produces **identically zero tendencies for any column**. The critical-level test `u·τ < 0` is satisfied by construction because the launch stress τ is defined antiparallel to the low-level wind, so every filtered level absorbs the entire flux before it can deposit momentum anywhere. Found while adding gradient fences in the #820 sweep (`simple_gwd_test.py::test_gradients_are_finite_on_a_calm_column` and neighbours); the existing smoke test passes vacuously because it only asserts finiteness.

## Evidence

`simple_gwd(u, 0, T, p, z, rho, h_std=300, dt=1800, SimpleGwdParameters.default())` on a
30-level column with a 6.5 K/km troposphere and a 210 K stratosphere, for four wind
profiles:

| profile | max abs tendency | surface `tau_x` | top `tau_x` |
|---|---|---|---|
| uniform 10 m/s | 0 | -415 | 0 |
| sheared 5 -> 40 m/s | 0 | -402 | 0 |
| 40 m/s jet at 12 km | 0 | -415 | 0 |
| reversal at 10 km | 0 | -415 | 0 |

The tendency is exactly zero in every case. `tau_x` is negative for a positive `u`, so
`u . tau < 0` at every level above launch (`simple_gwd.py:368-373`) and the whole flux is
"absorbed" immediately. Separately, a launch stress of ~4e2 Pa for `h_std = 300 m` and
`u = 10 m/s` is three to four orders above typical orographic GWD stresses (0.01 to 0.1 Pa),
so the launch-stress formula or its units want checking at the same time.

## Where the hooks are

- `simple_gwd.py` lines ~164-250 (launch stress and the critical-level filter) and ~399-433 (deposition).
- The scheme is a non-default fallback (`gravity_waves/simple/`), so no production configuration is affected, but anything selecting it gets no drag at all.

## What is needed

Compare against the intended reference formulation (the module docstring names it) for the sign convention of τ relative to u at the launch level and the critical-level criterion; add a test that a sheared column produces non-zero drag at the expected level.

Issue #841, opened 2026-09-19
WMO tropopause: level-orientation mismatch at the AeroCom call site, and a test fixture that never finds a tropopause


duncanwp:
## What is wrong

Two independent problems in the WMO tropopause diagnostic, found while adding gradient fences in the #820 sweep.

1. **Vertical orientation mismatch at the AeroCom call site.** `jcm/physics/diagnostics/wmo_tropopause.py::compute_geopotential_height` prepends `ps` and `cumsum`s upward, and the level scan in `find_tropopause_level` takes the first (lowest) qualifying level — i.e. it expects a **surface-first** column. `jcm/physics/diagnostics/aerocom.py:745` calls `find_tropopause_level(temperature.T, …)` with a **top-first** level axis (its `searchsorted(ref_p, 4000.)` window requires ascending pressure). On a top-first column the scan returns the *highest* qualifying level: 29 hPa on a real L47 sounding versus 209 hPa surface-first. Needs a check against ICON `mo_tropopause.f90` and a single stated convention (see `docs/source/design/output_vertical_conventions.md` for the three coexisting ones).

2. **The existing test fixture is vacuous.** `wmo_tropopause_test.py::create_test_atmosphere` clamps `maximum(T, 220)`, leaving an isothermal stratosphere, so the finder always returns `P_DEFAULT = 20000 Pa`, which satisfies the `10000 < p < 40000` assertions without a tropopause ever being found. The new gradient-fence fixture's docstring records this; the forward tests need a real lapse-rate profile.

## Where the hooks are

`wmo_tropopause.py:234::wmo_tropopause`, `find_tropopause_level`, `compute_geopotential_height`; `aerocom.py:745`; `wmo_tropopause_test.py`.

PR #847, opened 2026-09-19
check_gradients: a relative finite-difference step, and the NaN gradients it found across the ECHAM physics (#820)

### LINKED ISSUE #820: check_gradients' absolute finite-difference step is blind to large-magnitude input leaves

duncanwp:
Closes #820.

## What changed in the harness

`jcm.testing.check_gradients` perturbed every input leaf along a unit-scale tangent with an absolute step ladder (1e-3 down to ~1e-6), so a leaf at O(1e5) — SPEEDY's dry static energy, geopotential, pressure — never moved in float32 and the finite-difference reference carried no information about it. The tangent is now scaled per leaf by that leaf's primal RMS (`_tangent`, the mirror of the existing output-side `_cotangent`), the projection is contracted in float64, an empty leaf falls back to scale 1, and two opt-in knobs were added: `live_inputs=` asserts named input leaves carry a finite non-zero reverse gradient per leaf, and `fixed_inputs=` zeroes the tangent on structural leaves (grid metrics) whose derivative is meaningless. The step ladder and the jump/kink walk are unchanged; the `DEFAULT_STEPS` comment is re-derived for a relative step. On the vdiff column from the issue, `se` went from moving on 0 of 11 rungs to 11 of 11 and from 0.0008 % of the projection to 70 %.

All 28 existing SPEEDY callers were re-swept. None was hiding a wrong gradient; six fixtures sat exactly on a branch (a level mask boundary, `rh == rhcl1`, an isothermal surface) and were moved off it, two discrete outputs moved to `reference="adjoint"` + `live_inputs`, and the surface-flux `rtol` went to 2e-2 with the derivation in the test. Every file now passes at six independent seeds.

## What the sweep found in ECHAM

`check_gradients` had never been applied to the ECHAM package. It is now run at term level over every term of `echam_physics(radiation_scheme="grey")` at two operating points (`jcm/physics/echam/term_gradients_test.py`, 49 cases + a per-state-field finiteness sweep of the whole package tendency) and at scheme level beside each scheme's tests. Twelve gradient defects with a correct forward value were fixed, all by the double-`where` idiom. The forward value is bit-identical on the physical branch for every fix except the Planck constant fold and the two-stream refactors, where it moves by float32 round-off only (Planck: ≤1.6e-7 relative; all 124 existing radiation tests pass unmodified):

| where | mechanism |
|---|---|
| `grey_two_stream/planck.py` | `(hc)/(kT)`: the quotient's VJP squares `kT`, `(1.38e-23·250)² ~ 1e-41` underflows float32, `dB/dT = inf` at **every** temperature — the default ECHAM longwave gradient was NaN everywhere |
| `grey_two_stream/two_stream.py` ×4 | `exp(λτ)` on the discarded `where` branch (`0·inf`); layer albedo as a quotient of raw exponentials (forward NaN for λτ ≳ 44, the whole optically thick longwave column); `sqrt(γ1²−γ2²)` cancellation at ssa→1; `γ2/γ1` 0/0 at ssa=g=1 |
| `clouds/echam_1m.py` | liquid/ice partition `qc/(qc+qi)` floored at a float64-derived 1e-30; `den⁻²` overflows float32 below ~1e-19, so one exactly-zero liquid level under a vanishing ice tail poisoned every partial in the column |
| `convection/tiedtke_nordeng.py` | the 5 K/hr cap: `clip(M/max(|dtedt|,1e-30),0,1)` gives `0·inf` on every inactive level; the term's jvp w.r.t. incoming diagnostics was NaN |
| `tte_tke/tke_budget.py` | `sqrt(maximum(tke, 0))` at TKE = 0: the `maximum` tie splits 0.5/0.5 against `sqrt'(0) = inf` |
| `diagnostics/aerocom.py` | same tie-at-zero shape on the cloud-base updraft |
| `gravity_waves/simple/simple_gwd.py` ×4 | bare `sqrt(u²+v²)` at calm wind (cone tip) |

`JAX_gotchas.md` gains a section naming these four shapes and the idiom; the scheme-writing guide and `CLAUDE.md` now recommend `check_gradients`.

## What is tracked rather than fixed

- #843 — the kinks, jumps, `argmax` staircases and float32-conditioning items the sweep pinned (SPEEDY cloud top, TTE-TKE PBL height, SSO calm-column and orography-floor NaNs, the grey `maximum(condensate, 0)` clip, ECHAM 1M's 8 % float32 jvp/vjp gap near saturation, …). Each is a strict `xfail` or a pinned-zero fence naming `file:line` and the issue, so a fix flips a test.
- #846 — Lohmann 2M `nic_cirrus=1` ICNC gradient NaN below r ≈ 3e-7 m; the forward candidate is unphysical there too, so the floor is a science decision.
- #842 — Simple GWD deposits no momentum at all (launch stress antiparallel to the wind); #841 — WMO tropopause level-orientation mismatch at the AeroCom call site. Both are forward-physics findings outside gradient scope.

## Verification

`ruff check .` clean. All 37 test files the branch touches plus `jcm/testing_test.py` were run locally with `JAX_PLATFORMS=cpu`: 738 passed, 11 xfailed (all strict, all with reasons), 8 failed — the 8 reproduce identically on a clean `origin/dev` worktree and come from the laptop env (editable dinosaur 1.3.5 without semi-Lagrangian transport, no `jax_rrtmgp`/`jcosp`/`huggingface_hub`). The full fast/slow coverage gates run in CI (fast 92.7 %, slow 82.7 % on the first run). The one CI failure outside this branch's scope, `rrtmgp_test::test_skipped_steps_hold_the_effect_not_the_raw_flux`, is dependency drift: CI resolved the freshly released `jax-rrtmgp 0.4.0` and the test's held-fraction assertion is one float32 ulp wide (#849, fixed by #850; the same one-line commit is mirrored here so this PR's checks can go green). `term_gradients_test.py` takes ~230 s single-process; its cases are ~4 s each so none is marked slow.


duncanwp (commented):
## Adversarial pre-review — correctness only

Everything below was run against the branch worktree (`PYTHONPATH` pinned to it, `JAX_PLATFORMS=cpu`). Ranked most severe first.

---

### 1. `jcm/testing.py:222-227` — the relative step silently reverts to an absolute one at RMS ≲ 1e-19, not at the documented 1e-30

`_scaled_direction` computes the scale as

```python
values = jnp.asarray(leaf, jnp.float32)
rms = float(jnp.sqrt(jnp.mean(values**2))) if values.size else 0.0
scale = scale_from_rms(rms) if rms > 1e-30 else 1.0
```

`values**2` is a float32 square, so it underflows to zero for `|v| ≲ 1.08e-19` — the *same* `2**-63` bound this PR documents at `jcm/physics/clouds/echam_1m.py:37` for `_MIN_DIFFERENTIABLE_DENOMINATOR`. The `rms > 1e-30` guard is therefore unreachable: everything from ~1e-19 downwards lands on `scale = 1.0`. Measured (`_tangent` on a uniform leaf, top rung `1e-3`):

```
|x|=1.0e-18  tangent RMS=1.226e-18  -> top rung displaces by 1.23e-03 x the field
|x|=1.0e-19  tangent RMS=1.227e+00  -> top rung displaces by 1.23e+16 x the field
|x|=1.0e-22  tangent RMS=1.227e+00  -> top rung displaces by 1.23e+19 x the field
|x|=1.0e-30  tangent RMS=1.227e+00  -> top rung displaces by 1.23e+27 x the field
```

Failing scenario: a caller whose input leaf is a trace field that is small but not zero — a tracer at 1e-20 kg/kg, an aerosol number, or exactly the "spectral-ringing condensate tail — 4e-30 kg/kg of ice … an ordinary state of a running column" that `echam_1m.py:678-682` is written for — is displaced to O(1e-3) on every rung. The check then reports on a state the model never reaches, and it does so with no diagnostic: the docstring's "there is no better scale knowable than 1.0" reasoning applies to a leaf that is *identically zero*, but the code cannot distinguish that leaf from one at 1e-20.

Same expression, mirrored at the top end: an RMS at or above ~1e19 makes `values**2` overflow to `inf`, so `_tangent` returns a non-finite direction and `_cotangent` returns `1/inf = 0`, dropping that output leaf from the projection silently. Measured:

```
|x|=1.0e+18 tangent finite=True   cotangent max=2.299e-18
|x|=1.0e+19 tangent finite=False  cotangent max=0.000e+00
```

The module docstring's claim that rung `eps` "moves every leaf by a fraction `eps` of its own magnitude … whatever that magnitude is" holds only on roughly `[1e-18, 1e18]`.

---

### 2. `jcm/testing.py:428-470` — `fixed_inputs` turns a caught wrong gradient into a pass, and three SPEEDY callers apply it to a whole subtree

`_freeze` zeroes the tangent, and the reverse projection contracts against that same tangent, so a frozen leaf leaves both sides of the comparison. A `stop_gradient` on a frozen leaf is therefore invisible. Demonstrated:

```python
def bad(b, s): return {"a": jnp.sin(jax.lax.stop_gradient(b)/3e5), "s": s**2}
b = jnp.full((8,), 3.1e5); s = jnp.full((8,), 2.0)
check_gradients(bad, (b, s), rtol=1e-2)                          # AssertionError
check_gradients(bad, (b, s), rtol=1e-2, fixed_inputs=["[0]"])    # PASSED
```

The only guard is the docstring's "never that the check converges better without it". The three callers that use it name an interior node, and `_match_leaves` expands that to the whole subtree — `speedy_coords` is 10 leaves:

```
hsg, fsg, dhs, sigl, grdsig, grdscp, wvi, radang, sia, coa
```

`_freeze`'s own criterion is a leaf "whose value selects a code path rather than scaling a result", and the justifications given at `speedy_longwave_test.py:243-248`, `:291-296` and `speedy_shortwave_test.py:645-651` are entirely about `fsg` and `stratosphere_mask`. `dhs`, `grdsig`, `grdscp`, `sia`, `coa` and `radang` scale results rather than select branches, so six of the ten leaves are removed from both sides of those three checks without a stated reason.

---

### 3. Four `file:line` citations in xfail reasons point at `origin/dev` line numbers, not this branch's

`jcm/physics/clouds/echam_1m_test.py:1297` names `echam_1m.py:1216/1220` (`cloud_fraction > config.epsilon`), `:426` (`qc_in_cloud > 0.0`) and `:518` (`qi_in_cloud > 0.0`). Those are the pre-PR positions; the PR's own `+24` lines near the top of `echam_1m.py` moved them:

```
$ grep -n "qc_in_cloud > 0\|qi_in_cloud > 0\|cloud_fraction > config.epsilon" <dev>   -> 426, 518, 1217, 1221
$ grep -n  (same)                                                            <HEAD> -> 444, 536, 1241, 1245
```

On HEAD, `echam_1m.py:518` is a `cloud_fraction > config.epsilon`, so the reason's third citation names a different switch than the one it describes. `jcm/physics/echam/term_gradients_test.py:295` cites `echam_1m.py:1383` for the `wbf` structural zero; `"wbf": zero_col` is at 1418 on HEAD (1394 on dev). The remaining citations check out: `turbulence_coefficients.py:392` and `:108-111`, `lott_miller.py:111/114/377/442/466/537/554/606`, `hines.py:253-258`, `radiation_scheme.py:266-267`, `cloud_data.py:311`, `macv2_sp.py:105-111`, `cloud_optics.py:617/629`, `turbulent_fluxes.py:139/141/220/311/318/322`, `surface_physics.py:154`, `deposition_freezing.py:210`, `land.py:241/402`, `scheme.py:302`.

---

### 4. Every strict `xfail` omits `raises=`, so an unrelated exception also counts as the expected failure

`_match_leaves` raises `ValueError` when a name matches nothing (`jcm/testing.py:437`), and `term_gradients_test.py:276` carries both `live_inputs=("[0]/u_wind",)` and `xfail_reference=_PBL_HEIGHT_DEFECT` on the same cell. A rename of that leaf converts a loud `ValueError` into a green xfail, and the strictness that is supposed to flip on a fix never fires.

The current failures are honest — verified with `--runxfail`:

```
grey_two_stream_radiation-{stable,convecting}:
  AssertionError: no finite-difference step is usable in this direction ... one_sided=0.94..1.62 at every rung
tte_tke_vertical_diffusion-{stable,convecting}:
  AssertionError: ['vertical_diffusion']/pbl_height: gradient is identically zero
```

Both match their reason strings, and the `_CONDENSATE_CLIP_KINK` numbers (`D-=-1605.72  D+=-100.988` at the stable point) reproduce.

---

### 5. The PR body's "forward bit-identical" claim does not hold for `planck.py`

`grey_two_stream/planck.py:47-62` folds `hc/k` into one constant. Over 180–320 K in float32, 98 of 141 temperatures change, max relative 1.57e-7 (1–2 ulp):

```
planck hc_kt: bit-identical? False  max rel diff 1.57e-07  ndiff 98 / 141
d(hc_kt)/dT  old finite frac 0.0    new finite frac 1.0    new vs analytic max rel 1.15e-07
```

The defect claim is confirmed exactly — the old form's derivative was non-finite at *every* temperature. The forward shift is harmless in magnitude, but the PR body groups `planck.py` with the bit-identical fixes ("round-off only for the two-stream refactors"), and the code comment does not mention that the forward value moves at all; any bit-exact radiation regression reference at these temperatures shifts by ~1 ulp.

---

### 6. `two_stream.py:150` — the `lambda_tau >= 88` cut-off is a step of up to 7.8x in the layer albedo

Pre-existing, and the refactor preserves it exactly, but it sits in the lines this PR rewrites for differentiability: the normal branch tends to `gamma2` as `exp_minus_sq -> 0`, while `R_dif_asymptotic` is `clip(gamma2/gamma1_safe, 0, 1)`. The two do not meet. Measured across the cut-off:

```
ssa=0.9  g=0.7 : lam*tau = 87.9 -> 88.1   R_dif 0.17750 -> 0.47020   (2.6x)
ssa=0.99 g=0.85: lam*tau = 87.9 -> 88.1   R_dif 0.10887 -> 0.84481   (7.8x)
ssa=0.5  g=0.6 : lam*tau = 87.9 -> 88.1   R_dif 0.02500 -> 0.02439
```

At `ssa ~ 0` both branches are 0, so the absorbing grey longwave layers are unaffected; it is reachable for thick scattering layers. It is neither fixed nor filed, while adjacent jumps of the same class went to #843.

---

### 7. `term_gradients_test.py:300-302` — `adjoint_rtol=1.2e-1` under `reference="adjoint"` leaves that cell unable to detect a 10 % gradient error

With no difference reference, the echam_1m convecting cell asserts only (a) jvp against vjp to 12 % and (b) per-leaf liveness of `thermo_run` T and q. A sign-preserving factor-1.1 error in the scheme passes. The evidence offered (float64 identity to 6.8e-11, no `custom_jvp`/`custom_vjp`/`stop_gradient`) is the *same* evidence that `vertical_diffusion_test.py:1462` uses to record a strict xfail against #843 rather than to relax a tolerance. Two float32-conditioning findings of one class, two different dispositions.

---

### 8. `speedy_surface_flux_test.py:351` — rtol 5e-3 → 2e-2 buys a 4608-column grid the sweep elsewhere argues against

`IX, IL = 96, 48`, so the comment's "an O(eps) fraction of columns sits within any step of the stability kink" is a property of running 4608 columns at once. `term_gradients_test.py:23-27` states the alternative in this same PR ("**One column, not a grid** … the failure would be a property of the grid, not of the term"). The derivation itself is sound and the residual is real; the tolerance is what absorbs a fixture choice rather than a scheme defect.

Fixture hygiene on the same family is clean and the reasons are factually right: the old `qa = 0.5 * 1000. * [...]` against the stated `qsat` does put rh at exactly 0.30 = `rhcl1` on two of eight levels (levels 3 and 4), which is what `speedy_shortwave_test.py:691-700` claims; the replacement `rh_profile = linspace(0.35, 0.85, 8)` is a physically ordinary monotonic column that still produces cloud. The surface-flux fixture change (294 K SST under a 50 K lapse, 292 K land) is a more physical operating point than the isothermal default it replaces, not a weaker one.

---

### Categories that came out clean

**The 12 physics fixes' forward values and both-mode NaN-freedom.** Verified directly against the branch:

- `simple_gwd._safe_hypot` and `tiedtke_nordeng._tendency_cap_scale` are finite in both jvp and vjp where the forms they replace are not, and `_tendency_cap_scale` is bit-equal to the old `clip(_DTDT_MAX/maximum(|dtedt|,1e-30),0,1)` (`old == new forward: True`, `[1/3, 1, 1]` on a mostly-zero column).
- The two-stream algebra is exact, not approximate: with the Eddington coefficients at `two_stream.py:38-39`, `(gamma1-gamma2)(gamma1+gamma2) = 2(1-ssa) * 1.5(1-ssa*g) = 3(1-ssa)(1-ssa*g)`, and dividing numerator and denominator by `exp(+lambda_tau)` reproduces `1 - r^2 e^{-2λτ}` and `gamma2(1-e^{-2λτ})` identically. Across `tau in [1e-4, 200]` the forward change is ≤ 3.2e-6 absolute in R and ≤ 2.3e-5 in T for every `ssa ≤ 0.9999`; the one larger case (`ssa = 1-1e-6`, 7.3e-3 in R) is the cancellation regime, and a float64 reference confirms the new value is the accurate one. `ssa = 1` and `ssa = g = 1` are finite forward and in both AD modes where the old form was NaN; `ssa*g > 1` gives `lambda_sq < 0 -> lambda = 0` instead of the old `sqrt` of a negative.
- `echam_1m`'s threshold move from `d_epsilon` (1e-30) to `1e-18` changes the forward result only on `1e-30 < total_cloud <= 1e-18`, where `qc_frac`/`qi_frac` are reached only through `cond_total * qc_frac` with `|cond_total| <= total_cloud`, so the increment moves by ≤ 1e-18 kg/kg and `dT` by `(alhs-alhc)*1e-18/cpd ≈ 3e-16` K — the comment's "~1e-15 K" is the right order. The `2**-63` derivation behind the constant is correct.
- `aerocom.py` and `tke_budget.py` are `sqrt(max(x,0)) == where(x>0, sqrt(where(x>0,x,1)), 0)` for all x including negatives, so forward-identical.

**Science docs.** No `docs/source/science/` page is required by this diff: every fix is a numerics guard whose forward value is unchanged to round-off, the two pinned-to-zero derivatives (`sqrt` at its endpoint, `lambda` at `ssa = 1`) are the only finite choices available, and the one materially changed forward number (near-conservative-scattering albedo) is an accuracy correction rather than a different scheme.

**Repo hygiene.** No scratch artefacts, personal paths, GPU indices or bespoke drivers in the diff — `grep` for `/Users/`, `/glade/`, `/home/`, `TODO`, `FIXME`, `breakpoint(`, stray `print(` over the added lines returns nothing, only two files are added and both are `*_test.py` co-located with their module, and the worktree is clean.

`jcm/testing_test.py`, `planck_test.py` and `two_stream_test.py` pass here (88 passed), and the `grey_two_stream`/`tte_tke` slice of `term_gradients_test.py` runs 5 passed + 4 xfailed as described.

duncanwp:
Response to the adversarial pre-review, item by item.

1. **RMS underflow in `_scaled_direction`** — confirmed and fixed in 9cfa5e9f. The RMS is now reduced on the host in float64 (the projection already was). New tests place a leaf at 1e-25 and one at 1e25 and measure a relative displacement of ~1e-3 at the top rung; the issue's `stop_gradient` scenario still fails at both extremes. The docstring states the float32 window in which the scaled tangent itself is representable.
2. **`fixed_inputs` on the whole `speedy_coords` subtree** — confirmed for the two longwave callers and the shortwave cloud caller, which now freeze `speedy_coords/fsg` only (0282cf7a); `_freeze` raises if it would freeze every differentiable leaf, and both docstrings state that a frozen leaf's gradient is not checked at all. Refuted for `get_zonal_average_fields`: it does not read the sigma grid. It freezes `sia`/`coa`, which are sin and cos of the same latitudes, so an independent perturbation of the pair leaves `sia**2 + coa**2 = 1`. Narrowing the freeze exposed a downward-longwave fixture straddling `radset`'s temperature clip (seeds 3 and 5); the gradient check builds its own 210–310 K ramp and the Fortran regression fixture is untouched.
3. **Stale `file:line` citations** — confirmed and fixed in 8741d127, together with three further drifted citations (`cloud_optics.py`, `matrix_solver.py`, `macv2_sp.py`) and a `sundqvist_test.py` fixture comment that described a hinge and an `argmax` no longer present in the scheme.
4. **Strict xfails without `raises=`** — confirmed; every one now carries `raises=AssertionError` (5f20a078), and `--runxfail` shows each still fails with an `AssertionError` matching its reason.
5. **Planck fold not bit-identical** — confirmed; the PR body now lists it with the two-stream refactors as round-off only (≤1.6e-7 relative).
6. **Albedo step at the `lambda_tau` cut-off** — confirmed, and the finding goes further: the normal branch is also wrong, using `gamma2` and `gamma2/gamma1` where the homogeneous-layer solution has `Gamma = gamma2/(gamma1 + lambda)`, so a conservative layer reflects nothing and R at ssa=0.99, g=0.85, τ=10 is 0.099 against the exact 0.446. That changes the forward flux of every scattering layer and needs the grey references regenerated, so it is tracked in #848 with the adding-identity and conservative-limit proofs, and cross-linked from `two_stream.py`.
7. **`adjoint_rtol=1.2e-1`** — confirmed; the echam_1m convecting cell is now a strict xfail naming the float32 conditioning and #843, with its finiteness check live and the default `adjoint_rtol` restored (220a1ca7). Measured over seeds 0–5 the other relaxations were sized to their spread: hines 5e-2 → 1e-3 (≤1.2e-5), echam_1m stable 2e-2 → 2e-3 (≤3.6e-4), tte_tke keeps 1e-3 (≤3.0e-5).
8. **Surface-flux `rtol` paid for a T30 grid** — confirmed; the check now runs on an eight-column block placed at −3 K, the midpoint of `_stability_factor`'s stable branch, with `rtol=5e-3` restored (49cb95b8). It holds at seeds 0–9 except seed 7, which reports no usable rung rather than a disagreement.

duncanwp:
@codex review

duncanwp:
@codex review

